Issue #149, opened 2023-11-29
git lfs issue on cloning


aabrown100-git:
### Description

When cloning svFSIplus, get the following message:

`Encountered 76 files that should have been pointers, but weren't:
	tests/cases/cep/niederer_benchmark_ECGs_quadrature/mesh/h0.1/domain_info.dat
	tests/cases/cep/niederer_benchmark_ECGs_quadrature/mesh/x_coords.dat
	tests/cases/cep/niederer_benchmark_ECGs_quadrature/mesh/y_coords.dat
	tests/cases/cep/niederer_benchmark_ECGs_quadrature/mesh/z_coords.dat
	tests/cases/heats/diffusion_line_source/bc_left.dat
	tests/cases/stokes/manufactured_solution/P1P1/bforce/N004/bforce.dat
	tests/cases/stokes/manufactured_solution/P1P1/bforce/N004/bottom_vbc.dat
	tests/cases/stokes/manufactured_solution/P1P1/bforce/N004/left_vbc.dat
	tests/cases/stokes/manufactured_solution/P1P1/bforce/N004/right_vbc.dat
	tests/cases/stokes/manufactured_solution/P1P1/bforce/N004/top_vbc.dat
	tests/cases/stokes/manufactured_solution/P1P1/bforce/N008/bforce.dat
	tests/cases/stokes/manufactured_solution/P1P1/bforce/N008/bottom_vbc.dat
	tests/cases/stokes/manufactured_solution/P1P1/bforce/N008/left_vbc.dat
	tests/cases/stokes/manufactured_solution/P1P1/bforce/N008/right_vbc.dat
	tests/cases/stokes/manufactured_solution/P1P1/bforce/N008/top_vbc.dat
	tests/cases/stokes/manufactured_solution/P1P1/bforce/N016/bforce.dat
	tests/cases/stokes/manufactured_solution/P1P1/bforce/N016/bottom_vbc.dat
	tests/cases/stokes/manufactured_solution/P1P1/bforce/N016/left_vbc.dat
	tests/cases/stokes/manufactured_solution/P1P1/bforce/N016/right_vbc.dat
	tests/cases/stokes/manufactured_solution/P1P1/bforce/N016/top_vbc.dat
	tests/cases/stokes/manufactured_solution/P1P1/bforce/N032/bforce.dat
	tests/cases/stokes/manufactured_solution/P1P1/bforce/N032/bottom_vbc.dat
[...]

PR #170, opened 2024-01-24
Add GitHub Action to check for large files without git lfs

### LINKED ISSUE #26: Add warning when adding git lfs-tracked files without git lfs
We use `git lfs` to store large files (`.vtp`, `.vtu`) which is set in `.gitattributes`. This is tracked automatically, so files added with `git add` get automatically tracked.

Unfortunately, developers must first install `git lfs` before using this feature. If a developer doesn't have `git lfs` installed and adds files with `git add`, it adds the actual (potentially large) files to the repository.

I'll add a tool to give a warning when a pull request gets opened that contains actual files that should be tracked with `git lfs` instead of their pointers. This should avoid adding large files to the repository

mrp089:
## Current situation
Closes #26.

## Release Notes 
Will fail the pipeline if someone tries to merge a large file (currently set to `1MB`) if it's not tracked by `git lfs`. This hopefully will prevent the chaos behind #149. I will add the action as a branch protection rule for the master.

## Documentation
None.

## Testing
I'm opening this as a draft to see if the action is executed as intended.


mrp089:
Giving up on this. I can't get past the error `Resource not accessible by integration`, which indicates that the Action doesn't have sufficient permissions. I tried:
- setting `permissions` to `read-all` and `write-all` and combinations thereof
- adding a secret token
- checking only the `main` branch
Couldn't get it to work, giving up for now.

PR #174, opened 2024-01-25
Add shell test cases plate and valve

mrp089:
## Current situation
Shells were added in #75, this PR adds the tests for them. Crossing off one more test in #95.

## Release Notes 
Meshes and setup from https://github.com/vvedula22/svFSI-Tests/tree/master/09-shells, `xml` input files provided by @ktbolt. I also fixed a bug in the error message reporting the number of nodes that are above the tolerance.

## Documentation
None

## Testing
- Verified that `C++` and `FORTRAN` results match by running both solvers and comparing the outputs in ParaView
- Checked that the tests fail by outputting different time steps

## Code of Conduct & Contributing Guidelines 
- [x] I agree to follow the [Code of Conduct](https://github.com/SimVascular/.github/blob/main/CODE_OF_CONDUCT.md) and [Contributing Guidelines](https://github.com/SimVascular/.github/blob/main/CONTRIBUTING.md).

PR #175, opened 2024-01-25
Fix test fields

### LINKED ISSUE #173: Results are only compared for first field in each test
### Description

When I added the ALE test case in 9146764, I modified the test routine `run_with_reference` to incorporate testing several fields (e.g., pressure and velocity). This is a good idea in theory.

However, for reasons no longer obvious to me, I included this `return` statement if the results pass for a given field:
https://github.com/SimVascular/svFSIplus/blob/25e9186f7ab8d3102ec375d9a58224f565d6b7de/tests/conftest.py#L108-L109

This means that the testing will stop after checking the first field. If there is an error in any other field, it will not be detected.

### Reproduction

Add a bogus field to any of the `fields` arrays in any physics, and the tests will still pass.

### Expected behavior

We should loop over all defined fields for a test case and compare them to the reference. If the field doesn't exist (in simulation or reference), an error should be thrown.

### Additional context

The fix itself is easy: Remove the `return` statement.

However, there are now two problems:
1. Some reference solutions never existed
2. Some fields could have a difference higher than the default tolerance of `1.0e-12`

Problem 1 is somewhat easy to fix. I just need to recompute all the reference solutions that have fields missing and update them. Problem 2 could be spicy since there might be bugs already in the code that we haven't caught (but falsely thought we did). I'll report back.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct and Contributing Guidelines

mrp089:
## Current situation
Closes #173.

## Release Notes 
- fix test loop to run ALL fields (and report errors for all individually)
- reran ALL test cases
- tightened tolerances as much as possible (had to change some linear solvers from `NS` to `GMRES`)
- limited max number of nonlinear steps to the current converged one
- adapted tolerances for ALL fields we test to the tightest ones with which all tests still pass

## Documentation
Added testing guide #195

## Testing
- tests still pass
- and importantly: didn't pass before adapting the tolerances


mrp089:
@MatteoSalvador, I would merge this one first, and then I'll fix the conflicts in #174 and #188

MatteoSalvador (changes requested):
Thank you very much @mrp089, really nice job (as always)!

Before merging, I would only make sure that the `.inp` files perfectly match the `.xml` for all the test cases, for reproducibility reasons. I saw that sometimes tolerances and other settings have not been updated accordingly.

mrp089:
You mean `.inp` from FORTRAN `svFSI`? I tried deleting the remaining ones I found so there are no duplicates.

MatteoSalvador:
Thanks @mrp089! Following up on our conversation this afternoon, it is probably better to remove the `.inp` files from all the test cases and stick with `.xml` files only.

MatteoSalvador (approved):

PR #180, opened 2024-02-05
Add git ignore file and remove METISLib/temp directory.

ktbolt:
When building `svFSIplus/Code/ThirdParty/metis_svfsi` a `temp` directory is created containing some include files.

I've added a .gitignore file and removed the `temp` directory from the repository so when building this directory does not show up as being changed in git.


mrp089 (approved):

mrp089:
Thanks!

PR #182, opened 2024-02-06
Add CMake commands to use PETSc

ktbolt:
I've added an SV_PETSC_DIR option to set the PETSc install directory and CMake commands to define PETSc include and library flags (see https://github.com/SimVascular/svFSIplus/issues/181).


ktbolt:
This is really part of https://github.com/SimVascular/svFSIplus/issues/30.

MatteoSalvador (approved):

Issue #150, opened 2023-11-29
Add first unit test


mrp089:
### Problem

When implementing new features, having unit tests would be very useful. This is especially true for solid material models since they have a clearly defined interface:
- in: deformation gradient $\mathbf F$
- out: 2PK stress $\mathbf S$, tangent matrix $\mathbb{C}$

It's also straight-forward to come up with analytical solutions for materials to test against.

### Solution

We will add a single unit test, starting with the Neo-Hookean material under different loading conditions. Once we have a framework integrated into GitHub actions, we can expand it to other currently implemented hyperelastic materials (since we need to cover them anyway). Finally, we can use that framework to add new material models (#2, #33, #49).

### Additional context

@yuecheng-yu will add a `.cpp` and a `cmake` file. @mrp089 will integrate the executable into GitHub actions. @MatteoSalvador, input welcome!

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct and Contributing Guidelines

mrp089:
Reopening this after talking with @ktbolt, so we have a place to discuss and exchange ideas.

My PhD lab is using GoogleTest (or gtest) in their (closed-source) multi-physics FEM code. Here is their 30 min [video](https://www.youtube.com/watch?v=H9lWRVma2EE) guide how to include GoogleTest ([slides](https://zenodo.org/records/8228421)). This was part of a [Summer of Testing](https://www.youtube.com/playlist?list=PLzCxBiTw83uil4uoIhi3wlR3qiiulmKJ0) series. [Here](https://google.github.io/googletest/quickstart-cmake.html) is a quick start on including it using CMake.

A challenge in the current `svFSI+` will be creating mock objects, especially for `ComMod`.

mrp089:
@yuecheng-yu has prototyped a first unit test using GoogleTest and calling `get_pk2cc` on a solid material model. He's currently cleaning it up after running into issues with `cmake` and relative path definitions. The first application could be testing for a stress-free reference configuration in a couple of material models.

yuecheng-yu:
[...]

PR #198, opened 2024-02-22
Unit test 150

yuecheng-yu:
## Current situation
Refer to #150 

## Release Notes 
A unit testing program using GoogleTest to test `get_pk2cc` is implemented and built. It allows user defined deformation gradient as input to check the correctness of `get_pk2cc` (material model) by comparing the output stress and tangent tensors. 

## Documentation
None

## Testing
To be tested


mrp089:
@ktbolt, I'll look over this PR and after @yuecheng-yu and I have iterated over it, I'll assign you as a reviewer

tests/unitTests/unitTest.xml
```diff
@@ -0,0 +1,114 @@
...
+
+  <Add_face name="X0">
+      <Face_file_path>  ../../../../tests/unitTests/mesh/unitCube/mesh_surfaces/X0.vtp </Face_file_path>
+  </Add_face>
+
+  <Add_face name="X1">
+      <Face_file_path>  ../../../../tests/unitTests/mesh/unitCube/mesh_surfaces/X1.vtp </Face_file_path>
+  </Add_face>
+
+  <Mesh_scale_factor> 0.001 </Mesh_scale_factor>
```
mrp089, line 55: Would remove since it's not necessary

tests/unitTests/unitTest.xml
```diff
@@ -0,0 +1,114 @@
...
+      <Face_file_path>  ../../../../tests/unitTests/mesh/unitCube/mesh_surfaces/Y1.vtp </Face_file_path>
+  </Add_face>
+
+  <Add_face name="X0">
+      <Face_file_path>  ../../../../tests/unitTests/mesh/unitCube/mesh_surfaces/X0.vtp </Face_file_path>
+  </Add_face>
+
+  <Add_face name="X1">
+      <Face_file_path>  ../../../../tests/unitTests/mesh/unitCube/mesh_surfaces/X1.vtp </Face_file_path>
+  </Add_face>
```
mrp089, line 53: Can we remove all surfaces?

tests/unitTests/unitTest.xml
```diff
@@ -0,0 +1,114 @@
...
+      <Type> Dir </Type> 
+      <Value> 0.0 </Value> 
+      <Effective_direction> (0, 1, 0) </Effective_direction> 
+   </Add_BC> 
+
+   <Add_BC name="X0" > 
+      <Type> Dir </Type> 
+      <Value> 0.0 </Value> 
+      <Effective_direction> (1, 0, 0) </Effective_direction> 
+   </Add_BC>
```
mrp089, line 108: Would also remove BCs

tests/unitTests/unitTest.xml
```diff
@@ -0,0 +1,114 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
mrp089, line 1: In general, make this as minimal as possible and throw out any lines that are not necessary for the unit test

tests/unitTests/test.cpp
```diff
@@ -0,0 +1,186 @@
...
+    // Dm.print("Output stiffness matrix:");
+
+    // the following is where each argument of get_pk2cc from: 
+    // mat_models_carray::get_pk2cc<3>(com_mod, cep_mod, dmn, F, nFn, fN, ya_g, S, Dm);
+    // from struct_3d(...): com_mod, cep_mod, dmn <- eq <- com_mod, nFn, fN 
+    // from construct_dsolid(...): com_mod, cep_mod, nFn <- lM, fN <- lM
+    // from global_eq_assem(...): com_mod, cep_mod, lM
+    // from iterate_simulation(simulation): com_mod <- simulation, cep_mod <- simulation, 
+    // lM <- com_mod.msh[iM] (iM < com_mod.nMsh)
+    // iterate_simulation(simulation) <- run_simulation(simulation) <- main(...)
```
mrp089, line 180: Remove debug statements

tests/unitTests/test.cpp
```diff
@@ -0,0 +1,186 @@
...
+    auto iM = com_mod.nMsh;   // nMsh = 1 from unit mesh 
+    auto& lM = com_mod.msh[iM-1]; 
+    auto e = lM.gnEl;   // global number of elements
+    int eNoN = lM.eNoN;
+    int nFn = lM.nFn;
+    if (nFn == 0) {nFn = 1;}
+    Array<double> fN(nsd,nFn);
+    fN  = 0.0;
+
+    if (lM.fN.size() != 0) {
```
mrp089, line 148: There's a codecov warning that this if statement is not executed. Probably because no fibers are defined in the input file. If that's the case, just remove the if statement and leave `fN` as zero. If someone wants fibers in the future, they can add this code. I'd prefer this over having some code parts that we don't execute.

tests/unitTests/test.cpp
```diff
@@ -0,0 +1,186 @@
...
+#include "distribute.h"
+#include "initialize.h"
+#include "read_files.h"
+
+
+// Using global variables: 
+// for the output: stress tensor and tangent tensor
+
+Array<double> S_0(3,3), Dm_0(6,6);      // for identity test
+Array<double> S_1(3,3), Dm_1(6,6);      // for non-identity test
```
mrp089, line 56: Is there a way of structuring the tests without global variables? If these are reference solutions, can we put them inside the individual tests?

tests/unitTests/test.cpp
```diff
@@ -0,0 +1,186 @@
...
+}
+
+TEST(GetPK2ccTest, Identity) {
+  // input: 
+  // 1 0 0
+  // 0 1 0
+  // 0 0 1
+  for (short i = 0; i < 3; i++){
+      for (short j = 0; j < 3; j++){
+        EXPECT_EQ(S_0(i,j), 0);
```
mrp089, line 72: `0` -> `0.0` wherever it's a double (and not an int)

tests/unitTests/test.cpp
```diff
@@ -0,0 +1,186 @@
...
+Array<double> S_1(3,3), Dm_1(6,6);      // for non-identity test
+
+
+// Google Test test cases
+
+TEST(ExampleTest, OneEqualsToOne) {
+  EXPECT_EQ(1, 1);
+}
+
+TEST(GetPK2ccTest, Identity) {
```
mrp089, line 65: Give this a descriptive name like `zero_reference_stress`

tests/unitTests/test.cpp
```diff
@@ -0,0 +1,186 @@
...
+// Using global variables: 
+// for the output: stress tensor and tangent tensor
+
+Array<double> S_0(3,3), Dm_0(6,6);      // for identity test
+Array<double> S_1(3,3), Dm_1(6,6);      // for non-identity test
+
+
+// Google Test test cases
+
+TEST(ExampleTest, OneEqualsToOne) {
```
mrp089, line 61: Remove this test

tests/unitTests/test.cpp
```diff
@@ -0,0 +1,186 @@
...
+    auto cep_mod = simulation->get_cep_mod();
+    const int nsd  = com_mod.nsd;
+    auto iM = com_mod.nMsh;   // nMsh = 1 from unit mesh 
+    auto& lM = com_mod.msh[iM-1]; 
+    auto e = lM.gnEl;   // global number of elements
+    int eNoN = lM.eNoN;
+    int nFn = lM.nFn;
+    if (nFn == 0) {nFn = 1;}
+    Array<double> fN(nsd,nFn);
+    fN  = 0.0;
```
mrp089, line 146: Can these things be moved to a setup function?

tests/unitTests/hello_test.cpp
```diff
@@ -0,0 +1,53 @@
+/* Copyright (c) Stanford University, The Regents of the University of California, and others.
```
mrp089, line 1: Remove this file

tests/unitTests/test.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+
+// Google Test test cases
+
+TEST(ExampleTest, OneEqualsToOne) {
+  EXPECT_EQ(1, 1);
+}
+
+TEST(GetPK2ccTest, Identity) {
+  // input:
```
mrp089, line 66: Can you make three test cases with a different material each? Then it would already be a bit more modular for when we want to extend it

mrp089 (changes requested):
Thank you for making the unit tests work, @yuecheng-yu! I added a couple of comments on how to clean up the code and make it more modular. Please have a look and let me know if you have any questions.

Code/Source/svFSI/CMakeLists.txt
```diff
@@ -260,4 +260,72 @@ if(ENABLE_COVERAGE)
...
+  )
+
+  # link Google Test
+  target_link_libraries(
+    runUnitTest
+    gtest
+    GTest::gtest_main 
+    pthread   # link pthread on ubuntu20
+  )
+  target_compile_definitions(runUnitTest PRIVATE GTEST_HAS_PTHREAD=1)
```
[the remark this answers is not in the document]
yuecheng-yu: Yes, I love this idea. I think it could relate to "rearrange" the code to fit more test cases. Let's discuss it later!

tests/unitTests/test.cpp
```diff
@@ -0,0 +1,186 @@
...
+    //
+    int mpi_rank, mpi_size;
+    MPI_Init(&argc, &argv);
+    MPI_Comm_rank(MPI_COMM_WORLD, &mpi_rank);
+    MPI_Comm_size(MPI_COMM_WORLD, &mpi_size);
+
+    auto simulation = new Simulation();
+
+    std::string file_name = "../../../../tests/unitTests/unitTest.xml";
+    // std::string file_name = "";
```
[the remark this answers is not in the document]
yuecheng-yu: No problem!

tests/unitTests/test.cpp
```diff
@@ -0,0 +1,186 @@
...
+    // Initialize MPI.
+    //
+    int mpi_rank, mpi_size;
+    MPI_Init(&argc, &argv);
+    MPI_Comm_rank(MPI_COMM_WORLD, &mpi_rank);
+    MPI_Comm_size(MPI_COMM_WORLD, &mpi_size);
+
+    auto simulation = new Simulation();
+
+    std::string file_name = "../../../../tests/unitTests/unitTest.xml";
```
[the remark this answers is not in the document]
yuecheng-yu: Sure, it's a good idea!

tests/unitTests/unitTest.xml
yuecheng-yu: I'll try to simplify the input file and remain only the necessary lines.

tests/unitTests/test.cpp
yuecheng-yu: It's about why we need to read a file at this stage. And hopefully, this information could be useful for other people who will restructure the material models. Let's discuss how to rearrange the comments today!

tests/unitTests/test.cpp
yuecheng-yu: I struggled on this a little bit. Let's discuss it today.

tests/unitTests/test.cpp
yuecheng-yu: While GTest can implicitly convert int to double, it's a good practice to safe the type!

tests/unitTests/test.cpp
yuecheng-yu: Good idea! I'll try it. Let's discuss it later!

mrp089:
Very nice idea to use a fixture! Do you think we could use a [Mock object](https://chromium.googlesource.com/external/github.com/google/googletest/+/refs/tags/release-1.8.0/googlemock/docs/ForDummies.md#how-to-define-it) for things like `com_mod` and `dmn`?

yuecheng-yu:
Yes, I think it's possible. When I compared mock object with test fixture, I felt that while mock object is more ideal for unit testing, it may be harder to use for our current structures of material model. Because each type of model has completely different parameters/classes and also the parameters used in get_pk2cc such as C10/C01 are different from the input file (Elastic Modules, etc.), I think it could be a better timing to use Mock method for comprehensive unit testing after we have a new material model interface. Meanwhile, I personally really like the fixture thing which can rerun the setup code, which I think is the main reason I picked fixture to try first, please forget my previous blah blah. I'll try to write a GMock frame for nHK first to see how it works.

mrp089:
True, but unfortunately, creating a (draft) PR is necessary to get the code review function in GitHub (which nicely lets multiple people interact on the code). And it's better to have a discussion in the PR than not having a discussion. The discussion is still easily accessible since the merged commit is linked to the PR (and indirectly to the issue).

yuecheng-yu:
@mrp089 I created an Isotropic material class for unit test, wrapped the mock class inside the test class. So the users only need to define material parameters, input deformation gradient, and reference solutions at the testing interface (test.cpp).

Code/Source/svFSI/mat_models.cpp
mrp089, line 1: Checkout file from main

tests/unitTests/mesh/unit_cube_volume.vtu
mrp089, line 1: The mesh folder can be deleted now, right?

tests/unitTests/test.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+    // Step 5: run unit test
+    MR.runUnitTest(F, S_ref, Dm_ref);
+      
+}
+
+// int main(int argc, char **argv) {
+//    ::testing::InitGoogleTest(&argc, argv);
+//    return RUN_ALL_TESTS();
+// }
```
mrp089, line 94: Delete comments

tests/unitTests/test.h
```diff
@@ -0,0 +1,129 @@
...
+};
+
+
+class UnitTestIso {
+public:
+    MockComMod com_mod;
+    MockCepMod cep_mod;
+    
+    // UnitTestIso(consts::ConstitutiveModelType matType, double E, double nu, 
+    //             consts::ConstitutiveModelType penType, double pen)
```
mrp089, line 91: Delete comments

tests/unitTests/test.h
```diff
@@ -0,0 +1,129 @@
...
+    void runUnitTest(double F[3][3], double S_ref[3][3], double Dm_ref[6][6]) {
+        int nsd = com_mod.nsd;
+        auto &dmn = com_mod.mockEq.mockDmn;
+        // hard code for nHK
+        int nFn = 1; 
+        Array<double> fN(nsd, nFn);
+        double ya_g = 0.0;   
+        double S[3][3], Dm[6][6];
+        // memset(S, 0, sizeof(S));
+        // memset(Dm, 0, sizeof(Dm));
```
mrp089, line 114: Delete comments

tests/unitTests/test.h
```diff
@@ -0,0 +1,129 @@
...
+        // initialize if needed 
+        nsd = 3;
+    }
+    MockeqType mockEq;
+    MockmshType mockMsh;
+    // Mock methods if needed
+};
+
+
+class UnitTestIso {
```
mrp089, line 85: Add brief comment what this class does (unit test an isotropic material)

tests/unitTests/test.cpp
```diff
@@ -0,0 +1,94 @@
...
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "test.h"
+
+using namespace mat_fun;
+using namespace std;
+
+
+TEST(UnitTest, nHK) {
```
mrp089, line 37: Very nice!!!

tests/unitTests/test.cpp
```diff
@@ -0,0 +1,94 @@
...
+    double S_ref[3][3], Dm_ref[6][6];
+    memset(S_ref, 0, sizeof(S_ref));
+    memset(Dm_ref, 0, sizeof(Dm_ref));
+
+    // Step 5: run unit test
+    nHK.runUnitTest(F, S_ref, Dm_ref);
+      
+}
+
+TEST(UnitTest, MR) {
```
mrp089, line 64: Do these need different names? I only see this output:
```
Test project /Users/runner/work/svFSIplus/svFSIplus/build/svFSI-build/Source/svFSI
    Start 1: UnitTest
1/1 Test #1: UnitTest .........................   Passed    0.39 sec

100% tests passed, 0 tests failed out of 1

Total Test time (real) =   0.40 sec
```
Shouldn't there be two tests?

tests/unitTests/unitTest.xml
mrp089, line 1: This file can be deleted now?

mrp089 (changes requested):
Nicely done, @yuecheng-yu! This is looking good, just some minor cleanups.

tests/unitTests/test.cpp
yuecheng-yu: No, I think the 'Test #1' here means all test cases inside the 'test.cpp' (we can rename it also). It's a little bit confusing, but based on my understanding, no matter how many test cases we put in the 'test.cpp' file, there is only one "test" for `ctest`, because in the CMakeLists: `add_test(NAME all_unit_tests COMMAND run_all_unit_tests)` defines the "test". However, if you type `ctest -VV` or `ctest -verbose`, it will print out the result for each test case.

mrp089: Ah, I see! It's the same for `pytest`. Can you add verbosity to the `ctest` call in `test.yml`? Then I think we're good to go!

tests/unitTests/test.cpp
```diff
@@ -0,0 +1,94 @@
...
+    double nu = 0.5;   // Poisson_ratio
+    double pen = 4e9;   // Penalty_parameter
+    double C01;   // additional parameter to C10 (optional)
+
+    // Step 2: construct test object
+    UnitTestIso nHK(matType, E, nu, volType, pen, C01);
+
+    // Step 3: define the input 
+    double F[3][3]; 
+    memset(F, 0, sizeof(F));   // initialize
```
ktbolt, line 51: You can initialize arrays to 0 using braces, for example `F[3][3] = {}`;

yuecheng-yu: Thank you!! I didn't know this way for static array.

Issue #205, opened 2024-03-05
High order element face mesh cannot be loaded


wgyang:
### Description

When a high order mesh is used, the face mesh cannot be represented by vtp file. It should be written in an unstructured grid (vtu) file. svFSIplus uses load_vtp to read face files leading to an error when a high-order element mesh is used, for example, a curved quadratic triangle mesh. The P2 surface meshes used in the block compression test in svFSI-Tests are incorrect as they are written in vtp files. The figure below shows the face vtp file.
![x1_vtp](https://github.com/SimVascular/svFSIplus/assets/16581783/50ac5a2e-220b-4482-a0c7-5a9f4f990b66)

To support true high-order element meshes,  function read_vtp needs to accept mshtype as input in addition to face type when it reads face files.

### Reproduction

A tet10 mesh, svFSI.inp, and svFSI.xml are attached below for testing. 
[tet10_test.zip](https://github.com/SimVascular/svFSIplus/files/14505435/tet10_test.zip)

### Expected behavior

svFSIplus can read and output the high-order meshes it supports.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct and Contributing Guidelines

ktbolt:
@wgyang Please upload all of the files needed to run a simulation so I can test.

wgyang:
> @wgyang Please upload all of the files needed to run a simulation so I can test.

I modified my original post to include the mesh and input files.

ktbolt:
The problem was that the `read_vtp()` function was expecting to read a VTP file for some reason. 

I've added a check for a VTU file in`read_vtp()` and added a `load_vtu()` function that works with faces. I also needed to add `vtk_xml_parser` functions to store `vtkUnstructuredGrid` data into a face.

The simulation now runs but svFSIplus results don't match the Fortran results.

ktbolt:
I've fixed several bugs 
- indexing
- array out of bounds
- tet10 2nd derivatives
- C++ code not updated with Fortran bug fix 

A bug fest for sure!

So after fixing the .inp and .xml solver files to have the same parameters the C++ and Fortran code produce the same results.

wgyang:
@ktbolt  Thank you Dave!! I can add this example to the test cases once you merge your fix into the main branch.

ktbolt:
Thank Q @wgyang, allowed me to find several important bugs.

Issue #202, opened 2024-03-02
Using pre-computed velocity solutions in one-way coupled physics


zasexton:
### Problem

The general problem: We wish to use the (potentially time-varying) velocity solutions obtained from a previous simulation within a subsequent physics (namely, scalar advection-diffusion). 

### Solution

This has been done previously in `svFSI` on side-branches where residence time calculations have been desired or where thrombosis modeling has been done. See [Gutierrez _et al_ 2021](https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1009331).

Although it is not clear if this exact framework would work well for a more object-oriented `svFSIplus` the following scheme is generally implemented in `svFSI`:

1. `True`/`False` flag is selected for `Use pre-computed velocity`
2. If `True` the following subroutine is called in `svFSI` to read in the velocity solution vector fields across some set of time points. This subroutine can be found in the `VTKXML.f` file of the attached `svFSI` source code. *Reminder* this will not be found on the main branches of `svFSI` as this functionality was never formally included. 
```fortran
!**********************************************************************

      SUBROUTINE READVELOCITYVTU(fName)
      
      USE COMMOD
      USE LISTMOD
      USE ALLFUN
      USE vtkXMLMod
      
      IMPLICIT NONE

      CHARACTER(LEN=STDL) :: fName, sName
      
      TYPE(vtkXMLType) :: vtu
      
      INTEGER :: iStat, iEq, iOut, iM, l, s, e, a, b, Ac, nNo, oGrp
      INTEGER :: i, j, iTS, iVel, zp
      CHARACTER(LEN=stdL) :: varName, velName
      REAL(KIND=8), ALLOCATABLE :: tmpS(:,:), tmpGS(:,:)
      i = (lStep - fStep)/iStep + 1
      ALLOCATE (allU(nsd,gtnNo,i))
      iStat = 0
      CALL loadVTK(vtu, fName, iStat)
      IF (iStat .LT. 0) err = "VTU file read error (init)"
      
      CALL getVTK_numPoints(vtu, nNo, iStat)
      IF (iStat .LT. 0) err = "VTU file read error (num points)"
      IF (nNo .NE. SUM(msh(:)%gnNo)) 
     2   err = "Incompatible mesh and "//TRIM(fName)
      j=0
      ALLOCATE(tmpGS(nsd,gtnNo))

      iStat = -1
      iVel = 0
      zp = 0

      IF (fStep .GE. 1000) THEN
[...]

PR #209, opened 2024-03-26
Testing if the current build works without modifications

mrp089:
Two PRs, #202 and #205, seem to fail randomly in VTK under Ubuntu 22. Testing if they fail without modifications.


mrp089:
It doesn't

PR #229, opened 2024-07-28
Ustruct and volumetric material tests

aabrown100-git:
## Current situation
Material unit tests do not test ustruct material model implementations, nor volumetric penalty model implementations. https://github.com/SimVascular/svFSIplus/issues/150

## Release Notes 
- Add/update comments in material testing code and in material model implementation code
- Add instructions for running material tests
- Add tests for ustruct material model implementations
- Add tests for volumetric penalty model implementations (for both struct and ustruct)
- Fix bug in ustruct HO model where PK2 stress was not being updated

## Testing
 
This code itself is testing.


MatteoSalvador (approved):

Code/Source/svFSI/mat_models.cpp
```diff
@@ -765,7 +790,7 @@ void get_pk2cc_dev(const ComMod& com_mod, const CepMod& cep_mod, const dmnType&
       }
 
       double r1 = J2d*mat_ddot(C, Sb, nsd) / nd;
-      auto S = J2d*Sb - r1*Ci;
```
mrp089, line 768: Kudos for finding that! Did you observe any differences in the [rate of convergence](https://github.com/SimVascular/svFSIplus/issues/197#issuecomment-1962182146) using ustruct?

tests/unitTests/test.h
```diff
@@ -951,7 +1105,9 @@ class TestHolzapfelOgden : public TestMaterialModel {
             Psi += a_f / (2.0 * b_f) * chi(I4_f, k) * (exp(b_f * pow(I4_f - 1.0, 2)) - 1.0);   // Fiber term
             Psi += a_s / (2.0 * b_s) * chi(I4_s, k) * (exp(b_s * pow(I4_s - 1.0, 2)) - 1.0);   // Sheet term
             Psi += a_fs / (2.0 * b_fs) * (exp(b_fs * pow(I8_fs, 2)) - 1.0);                   // Cross-fiber term
-            Psi += kappa / 4.0 * (pow(smTerms.J, 2) - 1.0 - 2.0 * log(smTerms.J));      // Simo-Taylor 91 volumetric penalty term
+            if (!ustruct) {
+                Psi += kappa / 4.0 * (pow(smTerms.J, 2) - 1.0 - 2.0 * log(smTerms.J));      // Add Simo-Taylor 91 volumetric penalty term
+            }
```
mrp089, line 1110: It looks like the `!ustruct` part of the code is not executed. Maybe either add a test or delete?

tests/unitTests/test.h
```diff
@@ -774,6 +882,47 @@ class TestMaterialModel {
...
+
+        // Compute rho, beta, drhodp, dbetadp from g_vol_pen()
+        g_vol_pen(p, rho0, rho, beta, drhodp, dbetadp, Ja);
+
+        // Compare rho, beta, drho, dbeta with reference solutions
+        EXPECT_NEAR(rho, rho_ref, fmax(abs_tol, rel_tol * fabs(rho_ref)));
+        EXPECT_NEAR(beta, beta_ref, fmax(abs_tol, rel_tol * fabs(beta_ref)));
+        EXPECT_NEAR(drhodp, drhodp_ref, fmax(abs_tol, rel_tol * fabs(drhodp_ref)));
+        EXPECT_NEAR(dbetadp, dbetadp_ref, fmax(abs_tol, rel_tol * fabs(dbetadp_ref)));
+
```
mrp089, line 911: Optional: Could we write our own "near" function a la [numpy isclose](https://numpy.org/doc/stable/reference/generated/numpy.isclose.html)?
```
absolute(a - b) <= (atol + rtol * absolute(b))
```

tests/unitTests/test.h
```diff
@@ -851,6 +1000,11 @@ class TestMooneyRivlin : public TestMaterialModel {
 
 // ----------------------------------------------------------------------------
 // Class for test of Holzapfel-Ogden material model
+// Implements two versions of the Holzapfel-Ogden model:
+// - Full anisotropic invariants (I1_bar, I4_f, I4_s, I8_fs) used by cardiac mechanics benchmark paper (Arostica et al., 2024)
+// - Modified anisotropic invariants (I1_bar, I4_bar_f, I4_bar_s, I8_bar_fs) used by svFSIplus
+// Also, the strain energy includes the Simo-Taylor 91 volumetric penalty term,
+// with parameter HolzapfelOgdenParams.kappa
 class TestHolzapfelOgden : public TestMaterialModel {
```
mrp089, line 1008: In a perfect world, we would make `computeStrainEnergy` part of the (object-based) material models. Just to keep in mind for the future.

mrp089 (approved):
@aabrown100-git, I'm very happy you tackled testing the solid material models!

As someone who has [messed up a test-PR before](https://github.com/SimVascular/svFSIplus/pull/175), I have one piece of wisdom: Do your tests fail if you change parts of the code that the test should be sensitive to ("test the test")? You can verify this locally.

Also, thank you for adding comments to your (and existing code)! You could slightly change their format and they would show up in [Doxygen](https://simvascular.github.io/svFSIplus/class_test_material_model.html). There's a quick guide [here](https://opensource.com/sites/default/files/2022-05/OSDC_cheatsheet-doxygen-2022.5.30.pdf) (it's mostly converting to `///` or `@brief`). There's also a [plugin](https://marketplace.visualstudio.com/items?itemName=cschlosser.doxdocgen) for VS Code that makes it very easy (probably for other editors as well).

MatteoSalvador:
Thanks @mrp089 for your comments!
@aabrown100-git, once all the points raised by @mrp089 are addressed I will merge this PR!

tests/unitTests/test.h
aabrown100-git: Yes, is the class hierarchy I've done here what we had in mind for the object-oriented material model redesign?

Code/Source/svFSI/mat_models.cpp
aabrown100-git: No I haven't. This bug caused S for the ustruct HO model to always be zero, which obviously led to big issues when using the ustruct HO model. (we actually don't have a ustruct HO test yet). I think the ustruct rate of convergence issue must be due to another bug.

tests/unitTests/test.h
aabrown100-git: Yes thanks for pointing this out! I actually decided to remove the volumetric component from the HO model here, to be consistent with the other material model classes above. In doing so, I ended up removing this part of the code

aabrown100-git:
@mrp089 Thanks for the comments!

- I was able to make all the tests fail, so I think the tests are tested :)
- I just modified many of the comments to cooperate with Doxygen.
- I responded to your other comments individually.

tests/unitTests/test.h
mrp089: Yup!

tests/unitTests/test.h
mrp089: Makes sense!

Issue #233, opened 2024-08-01
No documentation on how to use Docker containers on an HPC system


dcodoni:
### Problem

There is not any documentation that describes how to use Docker containers on an HPC system. The present README.md file in the Docker/ folder describes how to execute a Docker container on a local machine through the Docker software. On an HPC system such as Sherlock, Singularity software is used to deal with containers, and a brief guide on how to use it would be very helpful for users.

### Solution

I will update the README.md file in the Docker/ folder to include a new section on how to use a container on HPC systems.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct and Contributing Guidelines

PR #234, opened 2024-08-01
Update Docker README.md

dcodoni:
Docker containers documentation updated to include the use of containers on an HPC system ( addresses #233 )

## Current situation
The present README.md file in Docker/ folder does not contain any information on how to use containers on HPC systems. It only describes how to use Docker software to build and execute Docker containers on a local machine.

## Documentation 
I added a new section on the README.md file describing how to pull a container from a public DockerHub repository, and how to execute the container using Singularity software (loaded by default on Sherlock).

### DIFF
--- a/Code/Source/svFSI/mat_models.cpp
+++ b/Code/Source/svFSI/mat_models.cpp
@@ -137,11 +137,23 @@ void get_fib_stress(const ComMod& com_mod, const CepMod& cep_mod, const fibStrsT
   }
 }
 
-/// @brief Compute 2nd Piola-Kirchhoff stress and material stiffness tensors
-/// including both dilational and isochoric components.
-///
-/// Reproduces the Fortran 'GETPK2CC' subroutine.
-//
+/**
+ * @brief Compute 2nd Piola-Kirchhoff stress and material stiffness tensors
+ * including both dilational and isochoric components.
+ *
+ * Reproduces the Fortran 'GETPK2CC' subroutine.
+ *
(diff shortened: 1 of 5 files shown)


ktbolt:
@dcodoni You should not be working in your **main** branch.

ktbolt:
@dcodoni The `How to use containers on an HPC system:` should be in the https://github.com/SimVascular/svFSIplus/blob/main/README.md file.

Also don't talk about Sherlock since only Stanford people know what that is. 

Most users will not want to build svFSIplus in a container. They will just want to download a container with a run-time image  to their HPC cluster. 

We will probably want to create a simvascular account on Docker Hub for our images.

MatteoSalvador:
Thanks @dcodoni for all these important instructions on HPC systems!

However, I suggest following the two comments from @ktbolt before merging this PR.

dcodoni:
Thanks @ktbolt for your feedback! 
I created this doc file quickly because many people were asking me about how to use containers on Sherlock, but I can close this pull request and make modifications on a branch. 
I will make these new modifications in https://github.com/SimVascular/svFSIplus/blob/main/README.md file. Should I add a new section in "getting started" (https://simvascular.github.io/svFSIplus/index.html)?

I created the containers just for automatic testing purposing. If there is a need to create a container with the executable only, it can be done. The problem with this is that an image must be created everytime the solver is updated. I may figure out something to automize this.

 I agree, we should create a DockerHub account for the group.

PR #240, opened 2024-08-20
Replace fsi_ls_upd() call that was mistakenly removed in https://gith…

aabrown100-git:
…ub.com/SimVascular/svFSIplus/pull/216. Also, call it if mvMsh is true, which I think is necessary for FSI sims. This improves nonlinear convergence in 0D-coupled struct/ustruct test cases.

## Current situation
https://github.com/SimVascular/svFSIplus/issues/239

## Release Notes 
- Replaces fsi_ls_upd() definition and subsequent call, which was mistakenly removed in a previous commit

## Testing
- No new tests, but existing struct/ustruct 0D-coupled tests (struct/LV_NeoHookean_passive_genBC, struct/LV_NeoHookean_passive_sv0D, ustruct/LV_NeoHookean_passive_genBC, ustruct/LV_NeoHookean_passive_sv0D) display improved nonlinear convergence

### DIFF
--- a/Code/Source/svFSI/eq_assem.cpp
+++ b/Code/Source/svFSI/eq_assem.cpp
@@ -102,7 +102,7 @@ void b_assem_neu_bc(ComMod& com_mod, const faceType& lFa, const Vector<double>& 
 
     for (int g = 0; g < lFa.nG; g++) {
       Vector<double> nV(nsd);
-      auto Nx = lFa.Nx.slice(g);
+      auto Nx = lFa.Nx.rslice(g);
       nn::gnnb(com_mod, lFa, e, g, nsd, nsd-1, eNoN, Nx, nV);
       double Jac = sqrt(utils::norm(nV));
       nV = nV / Jac;
@@ -271,7 +271,7 @@ void b_neu_folw_p(ComMod& com_mod, const bcType& lBc, const faceType& lFa, const
 
       // Get surface normal vector
       Vector<double> nV(nsd);
-      auto Nx_g = lFa.Nx.slice(g);
+      auto Nx_g = lFa.Nx.rslice(g);
       nn::gnnb(com_mod, lFa, e, g, nsd, nsd-1, eNoNb, Nx_g, nV);
       Jac = sqrt(utils::norm(nV));
       nV = nV / Jac;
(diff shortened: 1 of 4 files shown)


Code/Source/svFSI/eq_assem.cpp
```diff
@@ -302,6 +302,74 @@ void b_neu_folw_p(ComMod& com_mod, const bcType& lBc, const faceType& lFa, const
...
+  // Updating the value of the surface integral of the normal vector
+  // using the deformed configuration ('n' = new = timestep n+1)
+  sV = 0.0;
+  for (int e = 0; e < lFa.nEl; e++) {
+    if (lFa.eType == ElementType::NRB) {
+      // CALL NRBNNXB(msh(iM),lFa,e)
+    }
+    for (int g = 0; g < lFa.nG; g++) {
+      Vector<double> n(nsd);
+      auto Nx = lFa.Nx.slice(g);
```
ktbolt, line 348: It is more efficient to use the **Array3::rslice()** method here, does not copy data.

Code/Source/svFSI/eq_assem.cpp
```diff
@@ -302,6 +302,74 @@ void b_neu_folw_p(ComMod& com_mod, const bcType& lBc, const faceType& lFa, const
...
+/// contribution to the stiffness matrix to reflect deformed geometry, if using
+/// a follower pressure load.
+/// The value of this integral is stored in lhs%face%val.
+/// This integral is sV = int_Gammat (Na * n_i) (See Brown et al. 2024, Eq. 56)
+/// where Na is the shape function and n_i is the normal vector.
+///
+/// This function updates the variable lhs%face%val with the new value, which
+/// is eventually used in ADDBCMUL() in the linear solver to add the contribution
+/// from the resistance BC to the matrix-vector product of the tangent matrix and
+/// an arbitrary vector.
```
ktbolt, line 315: Nice comments!

dcodoni:
Good job @aabrown100-git, I don't have any other comments!

Code/Source/svFSI/eq_assem.cpp
aabrown100-git: Thanks!

Code/Source/svFSI/eq_assem.cpp
aabrown100-git: Done, thanks

aabrown100-git: I also changed all other instances of slice() in this file to rslice(). Each instance uses slice() in the same way: get shape function Nx for the current Gauss point g, and pass it to gnnb(). Since Nx is treated as a constant in gnnb(), and Nx isn't used anywhere else, I think it's fine to switch all instances to rslice()

aabrown100-git:
Addressed @ktbolt comments!

ktbolt (approved):
Looks good!

PR #250, opened 2024-08-26
Traction values file path 235

ktbolt:
I've added some checks for reading vtp and vtu files.


ktbolt:
I screwed this up!

Contrary to the function names (`read_vtu_pdata`) and the documentation (`(pressure/traction) is applied by reading a vtp file`) seems that lots of boundary condition data can take either a vtp or vtu file.

So I've modified the `read_trac_bcff()` function to read either a vtp or vtu file like everybody else.

PR #252, opened 2024-08-27
Fix reading in traction values file path vtu file 235

ktbolt:
I've modified the `read_trac_bcff()` function to read vtp and vtu files for the `Traction_values_file_path`.

Note that there are no tests for the `Traction_values_file_path` keyword.


aabrown100-git (approved):
Looks good to me!

Issue #244, opened 2024-08-23
There are duplicate trilinos fsi and fluid test cases 


ktbolt:
### Description

The fluid test cases have duplicate trilinos directories
```
pipe_RCR_3d_bj_trilinos/
pipe_RCR_3d_trilinos_bj/

pipe_RCR_3d_ilut_trilinos/
pipe_RCR_3d_trilinos_ilut/
```
The fsi test cases also have duplicate trilinos directories
```
pipe_3d_bj_trilinos/
pipe_3d_trilinos_bj/

pipe_3d_ml_trilinos/
pipe_3d_trilinos_ml/
```

### Reproduction

see above

### Expected behavior

Should have only one of the directories. 

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct and Contributing Guidelines

PR #253, opened 2024-08-27
Removed the duplicate directories (#244)

dcodoni:
## Current situation
In folder test/cases there are some duplicate folders that are not used to run the tests.
The directories used to run test using trilinos are: 
- in cases/fluid: pipe_RCR_3d_bj_trilinos/ pipe_RCR_3d_ilut_trilinos/
- in cases/fsi: pipe_3d_bj_trilinos/ pipe_3d_ml_trilinos/
- 
## Release Notes 
I removed the following folders:
- in cases/fluid: pipe_RCR_3d_trilinos_bj/ pipe_RCR_3d_trilinos_ilut/
- in cases/fyi: pipe_3d_trilinos_bj/ pipe_3d_trilinos_ml/


ktbolt:
@dcodoni Sorry to be so annoying but it would be better to name directories with the package name before the preconditioner (e.g. `pipe_RCR_3d_trilinos_bj` because all of the Trilinos directories would be listed together like this
```
pipe_RCR_3d_trilinos_bj/
pipe_RCR_3d_trilinos_ilut/
```

dcodoni:
@ktbolt that is not annoying at all. I agree with you and the solution you mentioned was my first choice. The reason why I changed it is that it is easier to distinguish which case must be run using PETSc and which using Trilinos in the conftest.py file (simply using "if folder.endswith("petsc")"). If this is a problem, I will find another way.

dcodoni:
@ktbolt I changed the directories' names

ktbolt:
@dcodoni Very good!

ktbolt (approved):

PR #256, opened 2024-08-27
Add updated test README.

ktbolt:
I've updated the tests/README a bit.


tests/README.md
```diff
@@ -1,57 +1,37 @@
-@page testing Testing Guide
+# Testing Guide
 
-[TOC]
+[Integration testing](https://en.wikipedia.org/wiki/Integration_testing) is an essential part of software development. It is performed when integrating code changes into the main development branch to verify that the code works as expected. The following sections describe how to run and add integration tests used to the svFSIplus program.
 
-[Integration testing](https://en.wikipedia.org/wiki/Integration_testing) is an essential part of software development. It is performed when integrating code changes into the main development branch to verify that the code works as expected. Below is a quick guide on how to run and add integration tests for `svFSI`.
+Running a test case requires 
+- Build svFSIplus
+- Install Git LFS used to download test data
```
dcodoni, line 7: Do you think it makes sense to invert the order, putting "install gti lfs" as first item? 
I don't know exactly how git lfs works but I noticed that it would be better to install it before cloning or at least before building to avoid any problems.

ktbolt:
I wonder why all the solver tests are run when updating a README file ?

tests/README.md
```diff
@@ -1,57 +1,37 @@
...
+- Build svZeroDSolver (only required from certain tests)
 
+# Build svFSIplus
+svFSIplus can be built follow these [instructions](../README.md).
```
dcodoni, line 11: by following these steps

tests/README.md
```diff
@@ -71,10 +51,10 @@ You can run an individual test by navigating to the `./tests/cases/<physics>/<te
 For more options, simply call `pytest -h`.
 
 ## Code coverage
-We expect that new code is fully covered with at least one integration test. We also strive to increase our coverage of existing code. You can have a look at our current code coverage [with Codecov](https://codecov.io/github/SimVascular/svFSIplus). It analyzes every pull request and checks the change of coverage (ideally increasing) and if any non-covered lines have been modified. We avoid modifying untested lines of codeas there is no guarante that the code will still do the same thing as before.
+We expect that new code is fully covered with at least one integration test. We also strive to increase our coverage of existing code. You can have a look at our current code coverage [with Codecov](https://codecov.io/github/SimVascular/svFSIplus). It analyzes every pull request and checks the change of coverage (ideally increasing) and if any non-covered lines have been modified. We avoid modifying untested lines of codeas there is no guarantee that the code will still do the same thing as before.
```
dcodoni, line 54: typo: We avoid modifying untested lines of _code as_

dcodoni (commented):
Very good, just very few comments/suggestions

dcodoni:
> I wonder why all the solver tests are run when updating a README file ?

@ktbolt I think because the .yml file for testing is set to run for any push/commit independent on what was changed

tests/README.md
ktbolt: The svFSIplus build process and `git lfs` are not connected, just need to run `git lfs` before running the tests.

tests/README.md
ktbolt: I changed it to `svFSIplus can be built following these [instructions](../README.md).`, the build process includes installing software and then the actual build steps.

ktbolt:
Let's run the solver tests again!

dcodoni (approved):

PR #261, opened 2024-08-30
Holzapfel Ogden Modified Anisotropy (HO-ma) Constitutive Model

divyaadil23:
## Current situation
https://github.com/SimVascular/svFSIplus/issues/249

## Release Notes 
- Added the Holzapfel Ogden Modified Anisotropy (HO-ma) Constitutive Model to material models.
- Updated Holzapfel Ogden Model.
- Both HO and HO-ma models are taken from https://github.com/vvedula22/svFSI/blob/master/Code/Source/svFSI/MATMODELS.f  and modified. Added additional terms containing the second derivative of smoothed heaviside funciton to the elasticity tensor to make it exact. Changed the approximate derivatives of heaviside function to exact. Removed fds (dot product of fiber directions in 8th invariant).
- Tested it with the new framework of unit tests - the unit tests now test the order of convergence for finite difference estimates. 
- Added some integrated tests for struct and ustruct for both models. 

## Documentation
 
Where can I document how to use the new material model? Also, should I add the derivation?

## Testing
 
Added 4 integration tests for HO and HO-ma models. Added new unit tests for both.


ktbolt:
It is impossible to review some of these commits because of all of the merging that went on, diffs are out of date.

divyaadil23:
What do you recommend? We can copy all changes to a single commit if that would help.

ktbolt:
It is good to try to squash commits, especially when there are 46 of them ...

And this PR seems to include changes for a lot of things, more than one material model and also some sort of testing.

PR #276, opened 2024-09-25
Terminate simulation for max iterations 192

ktbolt:
Improve the warning message a bit if the maximum number of nonlinear iterations is exceeded. 

Also added a more noticeable warning when a linear system solution does not converge.


Code/Source/svFSI/output.cpp
```diff
@@ -146,31 +146,40 @@ void output_result(Simulation* simulation,  std::array<double,3>& timeP, const i
...
     logger << sOut << std::endl;
   } else {
     logger << sOut << std::endl;
   }
 
-  // if max number of iterations is reached, throw a warning
-  if (eq.itr == eq.maxItr) {
-    logger << "MAX NUMBER OF NONLINEAR ITERATIONS REACHED. CHECK CONVERGENCE." << std::endl;
+  // Print a warning message if the maximum number of nonlinear iterations has been exceeded.
+  if (eq.itr > eq.maxItr) {
```
dcodoni, line 174: can eq.itr be greater than eq.maxItr? I think the loop stops at eq.itr=eq.maxItr, I am not sure the message will be printed.

dcodoni (commented):
I just have one comment.

Code/Source/svFSI/output.cpp
ktbolt: @dcodoni `eq.itr` can be larger than `eq.maxItr`, my thinking was that if it was equal than that is ok. But what really matters is the size of the residual. 

Note that there is code in the `picc()` that also checks `eq.itr` and other convergence measures
```
  bool l1 = (eq.itr >= eq.maxItr);
  bool l2 = (r1 <= eq.tol);
  bool l3 = (r1 <= eq.tol*eq.pNorm);
  bool l4 = (eq.itr >= eq.minItr);

  if (l1 || ((l2 || l3) && l4)) {
    eq.ok = true;
   }
```
If `l1` if false then `eq.ok` is set to false and later in the code
```
    if (eq.ok) {
      cEq = cEq + 1;
    }
```
so the current equation is not incremented?

dcodoni: I see. If l1 is false then eq.ok is set to false only if l2 and l3 are false. At this point the current equation is not incremented. But if l1 is true then the current equation is always incremented because eq.ok is always true.
I think that as soon as eq.itr becomes equal to eq.maxItr the current equation is incremented. For this reason I thought eq.itr would never go above eq.maxItr.
But maybe I am wrong

ktbolt: @dcodoni The tests I ran seemed to do the right thing but I can use `if (eq.itr == eq.maxItr) {` if you think it is better.

ktbolt: I've done more testing and thins seems to do the right thing. Let's merge it.

dcodoni: Sounds good @ktbolt. And thanks for doing this!

PR #283, opened 2024-10-04
Docker solver

dcodoni:
TEST: I want to test if SimVascular GitHub can push container on simvascular DockerHub account

## Current situation
We currently have docker container that includes scientific libraries. They set an environment to build the solver.

## Release Notes 
I added a dockerfile that uses dockerhub containers with the scientific libraries installed, and it builds the svFSIplus solver. The container is built through a workflow file which I created and it pushes the image created at every pull request on the dockerhub page of simvascular

## Testing
No further tests needed

### DIFF
--- a/.github/workflows/docker_solver_build.yml
+++ b/.github/workflows/docker_solver_build.yml
@@ -1,6 +1,6 @@
 name: Build and Push Docker Image
 
-on: [push]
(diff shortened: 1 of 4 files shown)


mrp089:
Super cool stuff!

PR #287, opened 2024-10-08
Updating docker_solver_build.yml

dcodoni:
Removing the old cache directory before updating it.

## Current situation
The workflow file correctly build and push the container with precompiled svFSIplus.
Currently it worked only for the first push on main. Subsequent pushes on main have problems updating the cache since the cache folder is not empty

## Release Notes 
I added a line that removes the old cache folder before updating it.


aabrown100-git (approved):
Looks good to me!

PR #288, opened 2024-10-10
Updating dockerfile for building the solver.

dcodoni:
## Current situation
Currently the container produced by the workflow cannot be run by any devices since there is a "permission problem" when the container is run.

## Release Notes 
The dockerfile now just build the solver without inserting an entrypoint with the command mpirun

### DIFF
--- a/.github/workflows/docker_solver_build.yml
+++ b/.github/workflows/docker_solver_build.yml
@@ -20,14 +20,6 @@ jobs:
       with:
         install: true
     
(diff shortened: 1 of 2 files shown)


dcodoni:
Unfortunately this particular modification cannot be tested before merging it to the main. The previous container had a unexpected behavior "no permission for mpirun". Probably because it was built through a workflow, since when I built the same container locally it worked as it should have.

Issue #292, opened 2024-10-22
Add test case for fluid equation with multiple domains


JonathanPham:
### Problem

It would be useful to have a test case that illustrates the use of multiple domains (with different parameter values, e.g., density, viscosity, etc.) for the fluid equation.

### Solution

We would create a new test case in the `fluid/` folder that illustrates the use of multiple domains.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct and Contributing Guidelines

PR #293, opened 2024-10-23
Updated fluid equations to add option for Brinkman porous media

### LINKED ISSUE #236: Merge Navier-Stokes-Brinkman implementation from Fortran to C++
### Problem

The [Navier-Stokes-Brinkman equations](https://github.com/fgerosa/svFSI/tree/Contact) were originally implemented by @fgerosa in the Fortran version of svFSI for modeling porous media in her [paper](https://www.sciencedirect.com/science/article/pii/S0045782524001981). We want to merge this implementation into svFSIplus.

### Solution

The Navier-Stokes equations implemented in `fluid.[h/cpp]` should be updated to be Navier-Stokes-Brinkman equations, where the Brinkman component can be turned off or on, depending on the desired physics. If this component is turned off, then we should recover exactly the original Navier-Stokes implementation. If it is turned on, then the simulations should exhibit permeability effects (from the Brinkman term).

There is a permeability parameter used in the Brinkman component. We can set this parameter to have a default value of 0, so that the default physics is Navier-Stokes. 

### Additional context

_No response_

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct and Contributing Guidelines

JonathanPham:
## Current situation
Closes #236 and #292

## Release Notes 
- Updated the Navier-Stokes equations in `fluid.cpp` to be the Navier-Stokes-Brinkman equations. This involves the addition of a single Darcy permeability term. This permeability is `Inverse_darcy_permeability` in the `.xml` input file. A value of 0.0 means zero inverse permeability, which effectively reduces the Navier-Stokes-Brinkman equations to the traditional Navier-Stokes equations. A non-zero value uses the Navier-Stokes-Brinkman formulation.

## Documentation
No further documentation needed.

## Testing
- Added an example case, `driven_cavity_2d_porous`, to illustrate the use of the permeability term and Navier-Stokes-Brinkman equations. This example case is similar to the classic 2D lid-driven cavity, but with the bottom half of the domain being porous media and the top half being fluid. This example case uses the fluid equation with multiple domains, where each domain has different properties (viscosity and permeability) and the domains are specified by a `.dat` text file.


Code/Source/svFSI/ComMod.h
```diff
@@ -1457,7 +1458,7 @@ class ComMod {
 
     /// @brief Total number of nodes (number of nodes on current proc across
     /// all meshes)
-    int tnNo = 0;
+    int tnNo = 0; // total number of nodes across all meshes, but only on current processor
```
ktbolt, line 1461: In-line comments clutter the  code, use better variable names or document these in the function comments.

ktbolt (changes requested):
Move all of these in-line comments elsewhere, maybe in the code block that uses them or in the function documentation.

ktbolt:
Looks good!

ktbolt (approved):

Issue #305, opened 2024-11-20
Rename svFSIplus to svMultiPhyisics 


ktbolt:
### Use Case

It is time to rename svFSIplus to svMultiPhyisics.

### Problem

none

### Solution

I will first make changes to the CMake files. 

### Alternatives considered

none

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct and Contributing Guidelines

ktbolt:
I've renamed svFSIplus in the CMake files. Building now creates
```
CMakeCache.txt		Externals-build/	Makefile		svMultiPhysics-build/
CMakeFiles/		Externals-prefix/	cmake_install.cmake	svMultiPhysics-prefix/
```
and `svMultiPhysics-build/bin/svmultiphysics`

ktbolt:
I've replaced the `svfsi`  Code/ThirdParty directories names 
```
README		metis_svfsi/	parmetis_svfsi/	tetgen/
eigen/		msmpi/		share/		tinyxml/
```
with `internal`
```
README.md		gklib_internal/		msmpi/			share/			tinyxml/
eigen/			metis_internal/		parmetis_internal/	tetgen/
```

It took me a `very very very long time to understand` how to do this:  variable names related to these directory names are created automatically in some CMake macros using a specific format. The directory names are also used in different forms, upper case, lower case, and different positions in a variable
```
THIRDPARTY_METIS_INTERNAL
metis_internal
METIS_INTERNAL_LIBRARY_NAME
SV_LIB_THIRDPARTY_METIS_INTERNAL_NAME
SV_USE_INTERNAL_METIS
USE_INTERNAL_METIS
```

In addition to these variables and the directory names themselves are used all over the place, ended up needing to change five files, a bunch of directory names and a bunch variable names. 

I will document this in the CMake code.

mrp089:
Happy the renaming is happening! I had a brief look at your branch, @ktbolt. The paths in `tests/conftest.py` still contain the `svfsiplus` executable.

The tests can't run anyways because they fail to activate the `svmultiphysics` conda environment because the Docker images still have the `svfsiplus` environment. However, the Docker containers will only be updated once the branch is merged into main (which it can't because the tests are failing). Seems like a 🐥 vs. 🥚 sort of problem. 

@dcodoni, is this correct, and is there a good way to resolve this?

dcodoni:
[...]

PR #308, opened 2024-12-04
Rename svFSIplus to svMultiPhysics

ktbolt:
These are the changes needed to rename the svFSIplus solver to svMultiPhysics (see https://github.com/SimVascular/svMultiPhysics/issues/305).

Lots of CMake code has been changed.

I also changed some of the Thirdparty directory names and the Code/Source files svFSI to solver and svFSILS to linear_solver.


mrp089:
There are a couple more files that come up with `grep -irl svfsiplus --exclude-dir={build,.git}`.

Those are mostly comments that can be changed with `find path-to-folder -type f -exec sed -i '' 's/svFSIplus/svMultiPhysics/g' {} +`.

The files `tests/test_*struct.py` refer to the two folders `tests/cases/*struct/LV_NeoHookean_passive_genBC/genBC_svFSIplus` that have `svFSIplus` in the name.

Docker/solver/dockerfile
```diff
@@ -1,24 +1,24 @@
...
 FROM simvascular/libraries:ubuntu22 AS builder
 
-# Get latest svFSIplus solver from GitHub repository
-RUN git clone https://github.com/SimVascular/svFSIplus.git
+# Get latest svMultiPhysics solver from GitHub repository
+RUN git clone https://github.com/SimVascular/svMultiPhysics.git
 
 # Trilinos build
-WORKDIR /svFSIplus/build-trilinos
+WORKDIR /svMultiPhysicsbuild-trilinos
```
dcodoni, line 8: there is a missing / (svMultiPhysics/build-trilinos)

dcodoni (commented):
I could only find a missing / in a dockerfile, I have no other comments.

dcodoni:
@ktbolt I saw you modified the dockerfiles for ubuntu20 and ubuntu22. I will make those changes locally and I will build the docker container and upload it to dockerhub.

Docker/solver/dockerfile
ktbolt: @dcodoni Good eye, fixed!

dcodoni (approved):

dcodoni:
@ktbolt I was checking why the tests failed. I was expecting test Ubuntu to fail since obviously the conda environment svmultiphysics does not exist at this point (I updated the simvascular/libraries:ubuntu22 on dockerhub but still the simvascular/solver:latest uses the old version of it and it will be updated after merging). 
Test MacOs should have worked but I observed this error: sh: genBC_svMultiPhysicsplus/genBC.exe: No such file or directory
Also i think you should change some tolerances of the tests. I remember I increased the tolerance after using the latest MacOs.

ktbolt:
@dcodoni As @mrp089 mentioned `struct` and `ustruct` tests used genbc directories using the svfsiplus solver name (?). I changed these names.

ktbolt:
There were more files named using the svFSIplus solver name, renamed those.

mrp089:
@dcodoni Renaming the solver is hopefully something we only do every couple of years (ideally never again). However, would it be useful to have an automated workflow for triggering a new Docker build and upload to Docker Hub from the Dockerfiles stored in this repository?

A use case would be updating, adding, or removing externals. This would also ensure that these are valid Dockerfiles that work with the current version of svMultiPhysics.

dcodoni:
@ktbolt @mrp089 do you think we should just have one container for ubuntu with the latest Ubuntu? Currently the workflow file is using container in my personal dockerhub account and it is looking for both Ubuntu20 and Ubuntu22.
I propose the following steps:
- create one docker container with Ubuntu latest
- upload the container in simvascular/libraries
- change the workflow file to just use simvascular/libraries:ubuntu-latest

ktbolt:
@dcodoni If we are going to use Actions to create installers then we need multiple versions of Ubuntu.

Note that builds and tests on the latest Ubuntu versions could fail. For example, VTK currently does not work on Ubuntu 22 and 24.

mrp089:
We could do an "officially supported" hard-coded version and a `-latest` version. This way, we see immediately if updates break things. Also, we already have the latest version running when support is dropped for a previous version.

`4C` has an action to [check if a Docker rebuild is required](https://github.com/4C-multiphysics/4C/blob/main/.github/actions/check_docker_build_required/action.yml). I think they do that by [comparing the hashes of their dependencies](https://github.com/4C-multiphysics/4C/blob/main/.github/actions/compute-and-check-dependencies-hash/action.yml).

ktbolt:
@mrp089 Sounds like a good strategy.

dcodoni:
@mrp089 this is a good idea! It requires some thinking: we currently build at each PR a container that as a BASE imports simvascular:libraries-ubuntu22. If we want an ubuntu-latest base library, this cant be built at each PR it would require an enormous amount of time, so we need to think who will build it and when to build it.

Anyway for now I uploaded:
- simvascular/libraries:ubuntu20
- simvascular/libraries:ubuntu22
@ktbolt try to update the test_ubuntu.yml in the container section replacing _container: dcodoni/lib:${{ matrix.image }}_ with _container: simvascular/libraries:${{ matrix.image }}_

Hopefully this will allow us to run the tests succesfully!

Issue #309, opened 2024-12-04
Results of the boundary integral (B_INT) are not written for each time step


ktbolt:
### Description

The results of the boundary integral (B_INT) are only written for the first time step and not for the remaining time steps.

### Reproduction

For a fluids simulation add the following lines to the solver XML file
```
   <Output type="B_INT" >
     <Pressure> true </Pressure>
   </Output>
```

The `B_NS_Pressure_average.txt` file contains only a single set of values
```
 lumen_inlet  lumen_outlet  lumen_wall
 1.249e+01  1.249e+01  3.766e+02 
```

### Expected behavior

Values should be written for each time step. 

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct and Contributing Guidelines

ktbolt:
The values in the B_NS_Pressure_average.txt file were actually face areas, values would follow later.

The original code used a `std::vector<bool> wtn(3)` to implicitly indicate output for spatial[0], boundary[1] or volume[2]. When in the code you are wondering what `wtn` might mean and see its obscure use like
```
if (!output.wtn[0]) {

for (int i = 0; i < 2; i++) {
  if (!wtn[i]) {
    continue;
  }

  if (i == 0) {
```

I've replaced `std::vector<bool> wtn(3)` with an `OutputOptions` class encapsulating Booleans for spatial, boundary and volume output types so you can do `lEq.output[k].options.boundary_integral = true` rather than `lEq.output[k].wtn[1] = true`.

The code originally computed and wrote both boundary and volume quantities in a single function. I've separated that code into two functions. I've also added a time step and time to the values written. For example: `B_NS_Pressure_average.txt` file 
```
# svMultiPhysics boundary integral results file. 
# Quantities represent averaged scalar or flux values over each mesh face.
#        
# Format 
# ------ 
# face areas: [list of mesh face areas]  
# step time [list of mesh face names] 
# [time step] [time] [list of computed values for each mesh face]
face areas:  2.957e+00  2.973e+00  4.644e+01 
step   time   lumen_inlet  lumen_outlet  lumen_wall 
 1   5.0000000000e-03  4.6318356113e+02  1.3745997101e+02  3.0236471504e+02 
 2   1.0000000000e-02  -6.7384998121e+01  1.7091915172e+02  4.8689056707e+01 
```

PR #313, opened 2024-12-11
Boundary integral file time step 309

ktbolt:
I've fixed writing integral quantities (see https://github.com/SimVascular/svMultiPhysics/issues/309).

I've also updated `svMultiPhysics/tests/cases/fluid/pipe_RCR_3d` to write integral quantities since this is not tested anywhere.


ktbolt:
The errors are caused by exceptions thrown by the integration functions (e.g. `all_fun::integ()`).  This was caused by initializing the output types to true like the original code did, changed that.

Code/Source/solver/read_files.cpp
```diff
@@ -2243,35 +2243,37 @@ void read_outputs(Simulation* simulation, EquationParameters* eq_params, eqType&
   // 
   for (int j = 0; j < 3; j++) {
     for (int k = 0; k < nDOP[j+1]; k++) {
-      lEq.output[k].wtn[j] = true;
+      lEq.output[k].options.boundary_integral = false;
```
dcodoni, line 2246: @ktbolt is the loop on j still necessary here?

dcodoni (commented):
Very good! I only have one comment/question.

Code/Source/solver/read_files.cpp
ktbolt: @dcodoni It was not clear to me from the code why all of the `wtn[j]` were set to true so I left the loop. But the `boundary_integral` are all initialized to false anyway so I will remove the loop.

dcodoni (approved):

Issue #297, opened 2024-11-04
Restructure material models with Eigen library


aabrown100-git:
### Problem
`mat_models.cpp` and `mat_models_carray.h` are basically redundant. They both implement material models, except `mat_models.cpp` uses custom `Array`s and `mat_models_carray.h` uses native carrays. Currently, the `get_pk2cc_dev()` function from `mat_models.cpp` is used for `ustruct`, while the `get_pk2cc()` function from `mat_models_carray.h` is used for `struct`. Also, there is a lot of redundancy between `get_pk2cc()` and `get_pk2cc_dev()`. 

In addition, both `Array`s and carrays have drawbacks. `Array`s allow to clean syntax, but [lead to slower code](https://github.com/SimVascular/svFSIplus/issues/85). carrays are faster, but require clunky syntax for basic operations and are memory unsafe.

All of this makes it confusing and leads to extra work when implementing or restructuring materials models.

### Solution

- Consolidate `mat_models.cpp` and `mat_models_carray.h` and use only `Array`s
- Consolidate `get_pk2cc()` and `get_pk2cc_dev()`
- Replace all `Array`s with [Eigen](https://eigen.tuxfamily.org/index.php?title=Main_Page) arrays.

### Additional context

This is a useful first step for subsequent restructuring and optimizations of the material models.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct and Contributing Guidelines

aabrown100-git:
I consolidated get_pk2cc() and get_pk2cc_dev(), and also consolidated mat_models.cpp and mat_models_carray.h (I found it more challenging to do these steps sequentially, so I just did them all together). You can find the latest branch [here](https://github.com/aabrown100-git/svFSIplus/tree/reorg_mat_models). I decided to remove the carray implementations (in mat_models_carray.h) and make everything use Array templates (in mat_models.cpp). The material unit tests and the integration tests all pass.

Next, I think we should use this branch to test out how to use Eigen arrays. After some preliminary thinking, I think a way to organize this would be to decide to use Eigen arrays "below" some level in the code. For example, we could convert all arrays in get_pk2cc() to Eigen arrays, but in struct3d() still use Array templates, and have some code to convert appropriately. Also, I think we should still use templated functions to deal with 2 or 3 spatial dimensions, so that the code is suitably performant. chatGPT suggested a nice way to implement this using a helper function. I can share at the DoD meeting on Friday.

[...]

PR #317, opened 2024-12-18
Location eigen not set 304

ktbolt:
Fix the CMake commands to build using the ThirdParty Eigen install rather than the system install.


aabrown100-git (approved):
Looks good. I confirmed that this change allows me to compile svMultiPhysics with Eigen on Sherlock (https://github.com/SimVascular/svMultiPhysics/issues/297)

Issue #307, opened 2024-12-03
Add Eigen as an optional backend for Vector and Array classes


zasexton:
### Use Case

For the `Vector.h` and `Array.h` files  we currently implement custom Vector and Array objects. Eigen is a high-level [C++](https://en.wikipedia.org/wiki/C%2B%2B) [library](https://en.wikipedia.org/wiki/Library_(computing)) of [template headers](https://en.wikipedia.org/wiki/Template_(C%2B%2B)) for [linear algebra](https://en.wikipedia.org/wiki/Linear_algebra), [matrix](https://en.wikipedia.org/wiki/Matrix_(mathematics)) and [vector](https://en.wikipedia.org/wiki/Vector_(mathematics_and_physics)) operations, geometrical transformations, [numerical solvers](https://en.wikipedia.org/wiki/Numerical_analysis) and related algorithms. Often Eigen or another such robust linear algebra library is used within large scientific computation or industry projects for clean, clear, and optimized data abstractions. Ultimately, leveraging Eigen can allow for many routines in svMultiPhysics to be expressed algebraically while accelerating computation with library vectorization.  

### Problem

Custom `Vector.h` and `Array.h` classes offer straightforward but non-optimized mathematical operations and lack of advanced linear algebra functionalities. These implementations typically require manual management of memory and computations, leading to increased code complexity and a higher risk of bugs and numerical inaccuracies. Additionally these custom classes may struggle with scalability and efficiency, making it difficult to handle large datasets or perform complex operations efficiently. This may result in code redundancies, performance inefficiencies, and limited extensibility of the current code base. 

### Solution

- Add flexible configuration flag `USE_EIGEN` to `CMAKE` to allow for `Vector.h` and `Array.h` to be built using `Eigen::Map` objects. 
- Add Eigen linear algebra algorithms to `mat_fun.h`/`mat_fun.cpp` files 

### Alternatives considered

Eigen seems the most convenient to add as a backend; however other libraries such as [Boost](https://en.wikipedia.org/wiki/Boost_(C%2B%2B_libraries)) might be considered with this same interfacing scheme. 

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct and Contributing Guidelines

zasexton:
The branch for add the optional Eigen backend is given here: https://github.com/zasexton/svFSIplus_public/tree/update-array-with-eigen-v2 . I will be testing the Array and Vector performances for the Eigen and original custom Array and Vector implementations. 

[...]

PR #318, opened 2024-12-18
Reorganize material models with Eigen

aabrown100-git:
## Current situation

Resolves https://github.com/SimVascular/svMultiPhysics/issues/297

## Release Notes 

- Removes struct implementation using C-arrays (`struct_3d_carray()` and `mat_models_carray.h`)
- Consolidates `get_pk2cc()` (for struct) and `get_pk2cc_dev()` (ustruct)
- Consolidates some common tensor operations in `get_pk2cc()` into functions
- Implements `get_pk2cc()` with fixed-size Eigen matrices and tensors

## Testing

- No new tests, but all integration and unit tests pass
- Updated [tests/cases/struct/LV_HolzapfelOgdenModifiedAnisotropy_passive/result_001.vtu](https://github.com/SimVascular/svMultiPhysics/compare/main...aabrown100-git:svMultiPhysics:reorg_mat_models?expand=1#diff-895937b164642e9d9416396a319d47f467e2935092c7b27df717fca9dd339c69), since original `get_pk2cc()` implementation was missing pressure component; see commit https://github.com/SimVascular/svMultiPhysics/commit/d1cf9251bd2d9c9ae4b99b34eba3deb9d1545847


mrp089:
Very nice work, @aabrown100-git!

I saw that there is a [`mat_models_Dave_array.cpp`](https://github.com/SimVascular/svMultiPhysics/pull/318/files#diff-398704688bac4a12e7d4f4e5e6b1672df8103d3efeb6fef3dd8631e7ccf1fcb6), which is marked as a new file. Is that just incorrectly tracked by git, or is that really a new file? Do we still want to redundant implementations, `Eigen` vs. `Dave array`?

Code/Source/solver/mat_fun.h
```diff
@@ -43,12 +47,62 @@
...
+    template <int nsd>
+    Eigen::Matrix<double, nsd, 1> cross_eigen(const Eigen::Matrix<double, nsd, 1>& u, const Eigen::Matrix<double, nsd, 1>& v) {
+        if constexpr (nsd == 2) {
+            return Eigen::Matrix<double, 2, 1>(v(1), - v(0));
+        }
+        else if constexpr (nsd == 3) {
+            return u.cross(v);
+        }
+        else {
+            throw std::runtime_error("Invalid number of spatial dimensions");
```
ktbolt, line 85: Add more information for the `throw` message
```
throw std::runtime_error("[cross_eigen] Invalid number of spatial dimensions '" + std::to_string(nsd) + "'.  Valid dimensions are 2 or 3.");
```

Code/Source/solver/mat_fun.h
```diff
@@ -43,12 +47,62 @@
...
+        }
+    }
+
     double mat_ddot(const Array<double>& A, const Array<double>& B, const int nd);
+    
+    template <int nsd>
+    double mat_ddot_eigen(const Eigen::Matrix<double, nsd, nsd>& A, const Eigen::Matrix<double, nsd, nsd>& B) {
```
ktbolt, line 92: Rename this function to something meaningful without `mat` and `eigen`, like `double_dot_product`.

Code/Source/solver/mat_fun.h
```diff
@@ -61,20 +115,307 @@ namespace mat_fun {
...
+     * @brief Contracts two 4th order tensors A and B, C_ijkl = A_ijmn * B_klmn over the last two dimensions
+     * 
+     * @tparam nsd, the number of spatial dimensions
+     * @param A, the first 4th order tensor
+     * @param B, the second 4th order tensor
+     * @return Eigen::TensorFixedSize<double, Eigen::Sizes<nsd, nsd, nsd, nsd>> 
+     */
+    template <int nsd>
+    Eigen::TensorFixedSize<double, Eigen::Sizes<nsd, nsd, nsd, nsd>>
+    ten_ddot_eigen(const Eigen::TensorFixedSize<double, Eigen::Sizes<nsd, nsd, nsd, nsd>>& A,
```
ktbolt, line 141: We don't need to prefix function names with `ten` and with an `eigen` suffix. 

It would be good to create a `Tensor` alias for `Eigen::TensorFixedSize<double, Eigen::Sizes<nsd, nsd, nsd, nsd>>`

Code/Source/solver/mat_fun.h
```diff
@@ -61,20 +115,307 @@ namespace mat_fun {
...
+        //                     for (int n = 0; n < nsd; ++n) {
+        //                         C(i,j,k,l) += A(i,j,m,n) * B(k,l,m,n);
+        //                     }
+        //                 }
+        //             }
+        //         }
+        //     }
+        // }
+
+        // return C;
```
ktbolt, line 174: Remove all of this commented-out code.

Code/Source/solver/mat_fun.h
```diff
@@ -61,20 +115,307 @@ namespace mat_fun {
...
+        //                     for (int n = 0; n < nsd; ++n) {
+        //                         C(i,j,k,l) += A(i,m,j,n) * B(m,n,k,l);
+        //                     }
+        //                 }
+        //             }
+        //         }
+        //     }
+        // }
+
+        // return C;
```
ktbolt, line 218: Remove commented-out code.

Code/Source/solver/mat_fun.h
```diff
@@ -61,20 +115,307 @@ namespace mat_fun {
...
+        //                     for (int n = 0; n < nsd; ++n) {
+        //                         C(i,j,k,l) += A(i,j,m,n) * B(k,m,l,n);
+        //                     }
+        //                 }
+        //             }
+        //         }
+        //     }
+        // }
+
+        // return C;
```
ktbolt, line 262: Remove commented-out code.

Code/Source/solver/mat_fun.h
```diff
@@ -61,20 +115,307 @@ namespace mat_fun {
...
+     * @brief Compute the dyadic product of two 2nd order tensors A and B, C_ijkl = A_ij * B_kl
+     * 
+     * @tparam nsd, the number of spatial dimensions
+     * @param A, the first 2nd order tensor
+     * @param B, the second 2nd order tensor
+     * @return Eigen::TensorFixedSize<double, Eigen::Sizes<nsd, nsd, nsd, nsd>> 
+     */
+    template <int nsd>
+    Eigen::TensorFixedSize<double, Eigen::Sizes<nsd, nsd, nsd, nsd>> 
+    ten_dyad_prod_eigen(const Eigen::Matrix<double, nsd, nsd>& A, const Eigen::Matrix<double, nsd, nsd>& B) {
```
ktbolt, line 277: Rename the function.

Code/Source/solver/mat_fun.h
```diff
@@ -61,20 +115,307 @@ namespace mat_fun {
...
+    /**
+     * @brief Create a 4th order identity tensor:
+     * I_ijkl = 0.5 * (δ_ik * δ_jl + δ_il * δ_jk)
+     * 
+     * @tparam nsd, the number of spatial dimensions
+     * @return Eigen::TensorFixedSize<double, Eigen::Sizes<nsd, nsd, nsd, nsd>> 
+     */
+    template <int nsd>
+    Eigen::TensorFixedSize<double, Eigen::Sizes<nsd, nsd, nsd, nsd>> 
+    ten_ids_eigen() {
```
ktbolt, line 320: Rename the function and perhaps create this data at compile time.

Code/Source/solver/mat_fun.h
```diff
@@ -61,20 +115,307 @@ namespace mat_fun {
...
+        // // Compute the product term1 = A_ik * B_jl and term2 = A_il * B_jk by shuffling the indices
+        // Eigen::TensorFixedSize<double, Eigen::Sizes<nsd, nsd, nsd, nsd>> term1, term2;
+        // term1.shuffle(Eigen::array<int, 4>{0, 2, 1, 3}) = AB; // A_{ik} * B_{jl}
+        // term2.shuffle(Eigen::array<int, 4>{0, 3, 1, 2}) = AB; // A_{il} * B_{jk}
+
+        // // Compute the symmetric product: C_{ijkl} = 0.5 * (A_{ik} * B_{jl} + A_{il} * B_{jk})
+        // Eigen::TensorFixedSize<double, Eigen::Sizes<nsd, nsd, nsd, nsd>> C = 0.5 * (term1 + term2);
+
+        // For some reason, the Eigen::Tensor contract function above is slower than the manual implementation below
+        // Initialize the result tensor
```
ktbolt, line 369: Remove commented-out code.

Code/Source/solver/mat_models.cpp
```diff
@@ -138,6 +250,57 @@ void get_fib_stress(const ComMod& com_mod, const CepMod& cep_mod, const fibStrsT
...
+  auto S_iso = J2d*S_bar - r1*Ci;
+
+  // Compute isochoric material elasticity tensor
+  EigenTensor<nsd> PP = ten_ids_eigen<nsd>() - (1.0/nsd) * ten_dyad_prod_eigen<nsd>(Ci, C); // Important: using auto here causes tests to fail
+  auto CC_iso = ten_ddot_eigen<nsd>(CC_bar, PP);
+  CC_iso = ten_transpose_eigen<nsd>(CC_iso);
+  CC_iso = ten_ddot_eigen<nsd>(PP, CC_iso);
+  CC_iso += (-2.0/nsd) * (ten_dyad_prod_eigen<nsd>(Ci, S_iso)  + ten_dyad_prod_eigen<nsd>(S_iso, Ci));
+  CC_iso += 2.0 * r1 * ten_symm_prod_eigen<nsd>(Ci, Ci)  +  (- 2.0*r1/nsd) * ten_dyad_prod_eigen<nsd>(Ci, Ci);
+
```
ktbolt, line 299: Would it be possible to replace some of the function calls by expressions ? 
```
auto CC_iso = CC_bar.contract(PP, contractionDims);
```
Would Eigen then be able to do a better job of optimizing them?

Code/Source/solver/mat_models.cpp
```diff
@@ -138,6 +250,57 @@ void get_fib_stress(const ComMod& com_mod, const CepMod& cep_mod, const fibStrsT
...
+  return std::make_pair(S_iso, CC_iso);
```
ktbolt, line 300: It's nice to return the result as the function value!

Note that `std::make_pair` makes copies of its arguments, not sure if that matters though, better to have a clear sense of what's returned.

Code/Source/solver/mat_models.cpp
```diff
@@ -153,10 +316,12 @@ void get_fib_stress(const ComMod& com_mod, const CepMod& cep_mod, const fibStrsT
...
  * @param[out] S 2nd Piola-Kirchhoff stress tensor (modified in place).
  * @param[out] Dm Material stiffness tensor (modified in place).
- * @return None, but modifies S and Dm in place.
+ * @param[out] Ja Jacobian for active strain
+ * @return None, but modifies S, Dm, and Ja in place.
  */
-void get_pk2cc(const ComMod& com_mod, const CepMod& cep_mod, const dmnType& lDmn, const Array<double>& F, const int nfd,
-    const Array<double>& fl, const double ya, Array<double>& S, Array<double>& Dm)
+template<size_t nsd>
+void _get_pk2cc(const ComMod& com_mod, const CepMod& cep_mod, const dmnType& lDmn, const EigenMatrix<nsd>& F, const int nfd,
```
ktbolt, line 323: What't the `_` for ?

Code/Source/solver/mat_models.cpp
```diff
@@ -188,116 +360,120 @@ void get_pk2cc(const ComMod& com_mod, const CepMod& cep_mod, const dmnType& lDmn
...
-  double J = mat_det(Fe, nsd);
+  EigenMatrix<nsd> Fe  = F;
+  EigenMatrix<nsd> Fa = EigenMatrix<nsd>::Identity();
+  EigenMatrix<nsd> Fai = Fa;
+
+  // if (cep_mod.cem.aStrain) {
+  //   actv_strain(com_mod, cep_mod, ya, nfd, fl, Fa);
+  //   Fai = Fa.inverse();
+  //   Fe = F * Fai;
+  // }
```
ktbolt, line 371: Remove commented-out code.

Code/Source/solver/mat_models.cpp
```diff
@@ -188,116 +360,120 @@ void get_pk2cc(const ComMod& com_mod, const CepMod& cep_mod, const dmnType& lDmn
...
-  auto Idm = mat_id(nsd);
-  auto C = mat_mul(transpose(Fe), Fe);
-  Array<double> E(nsd,nsd);
-  for (int i = 0; i < nsd; i++) {
-    for (int j = 0; j < nsd; j++) {
-      E(i,j) = 0.50 * (C(i,j) - Idm(i,j));
+  EigenMatrix<nsd> Idm = EigenMatrix<nsd>::Identity();
+  EigenMatrix<nsd> C = Fe.transpose() * Fe;
+  EigenMatrix<nsd> E = 0.50 * (C - Idm);
+
```
ktbolt, line 381: It's not clear to me what an `EigenMatrix<nsd>` is. Is this a vector or an nsd x nsd array ?

aabrown100-git:
> Very nice work, @aabrown100-git!
> 
> I saw that there is a [`mat_models_Dave_array.cpp`](https://github.com/SimVascular/svMultiPhysics/pull/318/files#diff-398704688bac4a12e7d4f4e5e6b1672df8103d3efeb6fef3dd8631e7ccf1fcb6), which is marked as a new file. Is that just incorrectly tracked by git, or is that really a new file? Do we still want to redundant implementations, `Eigen` vs. `Dave array`?

Oops sorry! `mat_models_Dave_array.cpp` was just a temporary file I created while converting to Eigen. I just removed it!

Code/Source/solver/mat_models.cpp
```diff
@@ -188,116 +360,120 @@ void get_pk2cc(const ComMod& com_mod, const CepMod& cep_mod, const dmnType& lDmn
...
-  }
-
-  // Now, compute isochoric and total stress, elasticity tensors
-  //
-  Tensor4<double> CC(nsd,nsd,nsd,nsd);
-
+  // Now, add isochoric and total stress, elasticity tensors
   switch (stM.isoType) {
     case ConstitutiveModelType::stIso_lin: {
+      if (ustruct) {
```
ktbolt, line 406: This check should be made much sooner than after the simulation has been partitioned and material properties are starting to be computed!

And also it is not clear what the meaning of the `ustruct` variable is unless you read the error message.

Code/Source/solver/mat_models.cpp
```diff
@@ -188,116 +360,120 @@ void get_pk2cc(const ComMod& com_mod, const CepMod& cep_mod, const dmnType& lDmn
...
     // NeoHookean model
     case ConstitutiveModelType::stIso_nHook: {
-      double g1 = 2.0 * stM.C10;
-      auto Sb = g1*Idm;
 
-      // Fiber reinforcement/active stress
-      Sb += Tfa * mat_dyad_prod(fl.col(0), fl.col(0), nsd);
+      // Compute fictious stress and elasticity tensor
+      EigenMatrix<nsd> S_bar = 2.0 * stM.C10 * Idm;
+      EigenTensor<nsd> CC_bar; CC_bar.setZero();
```
ktbolt, line 447: Put `CC_bar.setZero();` on a new line.

Code/Source/solver/mat_models.cpp
```diff
@@ -188,116 +360,120 @@ void get_pk2cc(const ComMod& com_mod, const CepMod& cep_mod, const dmnType& lDmn
...
 
-      // Fiber reinforcement/active stress
-      Sb = Sb + Tfa*mat_dyad_prod(fl.col(0), fl.col(0), nsd);
+      // Compute fictious stress and elasticity tensor
+      EigenMatrix<nsd> S_bar = 2.0 * (stM.C10 + Inv1 * stM.C01) * Idm 
+                              -2.0 * stM.C01 * J2d * C;
 
-      g1  = 4.0*J4d* stM.C01;
-      auto CCb = g1 * (ten_dyad_prod(Idm, Idm, nsd) - ten_ids(nsd));
+      EigenTensor<nsd> CC_bar = 4.0 * J4d * stM.C01 * (ten_dyad_prod_eigen<nsd>(Idm, Idm) - ten_ids_eigen<nsd>());
```
ktbolt, line 466: Note that you don't need to use `<nsd>` for template functions that use `nsd` in its parameters
```
EigenTensor<nsd> CC_bar = 4.0 * J4d * stM.C01 * (ten_dyad_prod_eigen(Idm, Idm) - ten_ids_eigen<nsd>());
```

Code/Source/solver/mat_models.cpp
```diff
@@ -542,523 +701,129 @@ void get_pk2cc(const ComMod& com_mod, const CepMod& cep_mod, const dmnType& lDmn
...
 
-      // Fiber-fiber interaction stress + additional reinforcement (Tfa)
-      double rexp = exp(stM.bff * Eff * Eff);
-      g1 = c4f*Eff*rexp;
-      g1 = g1 + (0.5*dc4f/stM.bff)*(rexp - 1.0);
-      g1 = (2.0*stM.aff*g1) + Tfa;
-      auto Hff = mat_dyad_prod(fl.col(0), fl.col(0), nsd);
-      S  = S + g1*Hff;
+        // Call templated function
+        _get_pk2cc<2>(com_mod, cep_mod, lDmn, F_2D, nfd, fl_2D, ya, S_2D, Dm_2D, Ja);
```
ktbolt, line 796: Note that a function with a prefixed  `_` has a special meaning in the C++ implementation (e.g. global namespaces).

ktbolt (commented):
@aabrown100-git Lots of good work here!

mrp089 (changes requested):

aabrown100-git:
Looks like `-march=native` causes an issue on GitHub Ubuntu machines; the integration tests timed out when I included `-march=native` in CMakeLists.txt, but passed in the most recent commit when I removed `-march=native`. From the [timing tests I performed](https://github.com/SimVascular/svMultiPhysics/issues/297#issuecomment-2552060797), there is some improvement with `-march=native`, but it's not huge. Also, here's a relevant [discussion](https://stackoverflow.com/questions/52653025/why-is-march-native-not-enabled-by-default-by-compilers-ides) why `-march=native` is not enabled by default. For these reasons, I'll remove `-march=native` from CMakeLists.txt. Perhaps we can revisit this once @zasexton [completes the Eigen backend for Arrays](https://github.com/SimVascular/svMultiPhysics/issues/307) and learns more about Eigen performance.

Code/Source/solver/mat_fun.h
```diff
@@ -61,20 +115,307 @@ namespace mat_fun {
...
+        // Compute the dyadic product: C_ijkl = A_ij * B_kl
+        for (int i = 0; i < nsd; ++i) {
+            for (int j = 0; j < nsd; ++j) {
+                for (int k = 0; k < nsd; ++k) {
+                    for (int l = 0; l < nsd; ++l) {
+                        C(i,j,k,l) = A(i,j) * B(k,l);
+                    }
+                }
+            }
+        }
```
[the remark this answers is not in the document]
aabrown100-git: Yes, the for loop is faster in this case. This may be because the tensors we are dealing with are small (2x2 or 3x3), so the contract method is overkill.

Code/Source/solver/mat_models.cpp
```diff
@@ -138,6 +250,57 @@ void get_fib_stress(const ComMod& com_mod, const CepMod& cep_mod, const fibStrsT
...
+  double r1 = J2d * mat_ddot_eigen<nsd>(C, S_bar) / nsd;
+
+  // Compute isochoric 2nd Piola-Kirchhoff stress
+  auto S_iso = J2d*S_bar - r1*Ci;
+
+  // Compute isochoric material elasticity tensor
+  EigenTensor<nsd> PP = ten_ids_eigen<nsd>() - (1.0/nsd) * ten_dyad_prod_eigen<nsd>(Ci, C); // Important: using auto here causes tests to fail
+  auto CC_iso = ten_ddot_eigen<nsd>(CC_bar, PP);
+  CC_iso = ten_transpose_eigen<nsd>(CC_iso);
+  CC_iso = ten_ddot_eigen<nsd>(PP, CC_iso);
```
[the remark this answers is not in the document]
aabrown100-git: Yes, CC_iso should be overwritten twice. The full operation is CC_iso = PP : (CC_bar : PP)^T, but it's broken up into three operations.

Code/Source/solver/mat_models.cpp
```diff
@@ -138,6 +250,57 @@ void get_fib_stress(const ComMod& com_mod, const CepMod& cep_mod, const fibStrsT
...
+  CC_iso += (-2.0/nsd) * (ten_dyad_prod_eigen<nsd>(Ci, S_iso)  + ten_dyad_prod_eigen<nsd>(S_iso, Ci));
+  CC_iso += 2.0 * r1 * ten_symm_prod_eigen<nsd>(Ci, Ci)  +  (- 2.0*r1/nsd) * ten_dyad_prod_eigen<nsd>(Ci, Ci);
```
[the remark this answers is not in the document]
aabrown100-git: I think I'll keep it as is, since it's probably a bit more efficient

Code/Source/solver/mat_models.cpp
aabrown100-git: Yes, in fact, currently `double_dot_product_2323()` is just a wrapper on the `contract()` method. I think it's just a question of readability. Here is the code contained inside `double_dot_product_2323()`

        // Define the contraction dimensions
        Eigen::array<Eigen::IndexPair<int>, 2> contractionDims = {
            Eigen::IndexPair<int>(2, 2), // Contract A's 2nd dimension with B's 2nd
            Eigen::IndexPair<int>(3, 3)  // Contract A's 3rd dimension with B's 3rd
        };

        // Return the double dot product
        return A.contract(B, contractionDims);

I feel that by wrapping this code and the definition of contractionDims, it is more readable. But open to suggestions!

Code/Source/solver/mat_models.cpp
aabrown100-git: Ah I see. Maybe this is causing some slowdown. Do you recommend any changes to try?

Code/Source/solver/mat_models.cpp
aabrown100-git: See my response to your other comment about the meaning of underscore here.

Code/Source/solver/mat_models.cpp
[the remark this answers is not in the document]
aabrown100-git: This code is for active strain. Active strain isn't currently tested, but if/when we want to test and use it, I think this commented code will be useful. IMO, it doesn't hurt to keep this commented code, and it could be helpful to a developer in the future.

Code/Source/solver/mat_models.cpp
aabrown100-git: `EigenMatrix<nsd>` is an nsd x nsd array. `EigenMatrix` is an alias defined at the top of the file.

template<size_t nsd>
using EigenMatrix = Eigen::Matrix<double, nsd, nsd>;

Open to any suggestions for a different alias name!

Code/Source/solver/mat_models.cpp
```diff
@@ -1744,4 +1509,206 @@ void g_vol_pen(const ComMod& com_mod, const dmnType& lDmn, const double p,
...
+ * @param Nx Shape function gradient w.r.t. reference configuration coordinates (dN/dX)
+ * @param vx Velocity gradient matrix w.r.t reference configuration coordinates (dv/dX)
+ * @param F Deformation gradient matrix
+ * @param Svis Viscous 2nd Piola-Kirchhoff stress matrix
+ * @param Kvis_u Viscous tangent matrix contribution due to displacement
+ * @param Kvis_v Visous tangent matrix contribution due to velocity
+ */
+void get_visc_stress_pot(const double mu, const int eNoN, const Array<double>& Nx, const Array<double>& vx, const Array<double>& F,
+                        Array<double>& Svis, Array3<double>& Kvis_u, Array3<double>& Kvis_v) {
+
```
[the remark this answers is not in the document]
aabrown100-git: I used "get" to be consistent with `get_pk2cc()`. But happy to change this (or both) to something better like "compute" or "calculate".

Code/Source/solver/sv_struct.cpp
```diff
@@ -1097,135 +761,79 @@ void struct_3d(ComMod& com_mod, CepMod& cep_mod, const int eNoN, const int nFn,
...
-      // Viscous terms contribution
-      Tv = (2.0*(DdNx(0,a)*NxFi(0,b) - DdNx(0,b)*NxFi(0,a)) - (NxNx*VxFi(0,0) + NxFi(0,b)*VxNx(0,a) -  
-           r23*NxFi(0,a)*VxNx(0,b))) * rmu + (r13*NxFi(0,a)*NxFi(0,b) + NxNx) * rmv;
-
-      //dmsg << "Tv: " << Tv;
-      //dmsg << "BmDBm: " << BmDBm;
-      //dmsg << "NxNx: " << NxNx;
-
-      lK(0,a,b) = lK(0,a,b) + w*(T1 + afu*BmDBm + Tv);
+      lK(0,a,b) = lK(0,a,b) + w*( T1 + afu*(BmDBm + Kvis_u(0,a,b)) + afv*Kvis_v(0,a,b) );
```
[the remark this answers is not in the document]
aabrown100-git: Is lK += ... more efficient than lK = lK + ...? Also, if I use += here, should I use += everywhere else in this file?

tests/unitTests/test.h
```diff
@@ -429,36 +428,39 @@ class TestMaterialModel {
...
+        Array<double> S_arr(3,3);
+        Array<double> Dm_arr(6,6);
+        for (int i = 0; i < 3; i++) {
+            for (int J = 0; J < 3; J++) {
                     F_arr(i, J) = F[i][J];
-                }
             }
+        }
+
+        double J = 0; // Jacobian (not used in this testing)
```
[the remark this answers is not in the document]
aabrown100-git: `J` is not analyzed in these tests, but I need to create it in order to run `get_pk2cc()`

Code/Source/solver/sv_struct.cpp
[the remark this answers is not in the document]
ktbolt: `+=` would only be more efficient if temporary values were being created (e.g. Array objects) but this is not the case here.

Fortran does not have an `+=` operator so I left these statements to resemble the Fortran code for debugging. 

I find `lK(0,a,b) = lK(0,a,b) +` to be clearer than `+=` but it is more verbose. If we do change it then we should change it in all files.

Code/Source/solver/mat_models.cpp
ktbolt: @aabrown100-git The `double_dot_product_2323()` and `double_dot_product_3424()` functions are only used in `_get_pk2cc()` so I would define `prod_dims2323` and `prod_dims3424` locally and 
```
CC = CC. contract(CC_bar, prod_dims2323);
```

Or  if these functions are going to be use more often then generalize `double_dot_product()` to use product dimensions as parameters
```
double_dot_product(CC, {2,3}, CC_bar, {2,3});
```

Code/Source/solver/mat_models.cpp
```diff
@@ -375,57 +546,52 @@ void get_pk2cc(const ComMod& com_mod, const CepMod& cep_mod, const dmnType& lDmn
...
-
-      S  = S + p*J*Ci;
-      CC = CC + 2.0*(r1 - p*J) * ten_symm_prod(Ci, Ci, nsd) + (pl*J - 2.0*r1/nd) * ten_dyad_prod(Ci, Ci, nsd);
+      CC_bar += g1 * ten_dyad_prod_eigen<nsd>(RmRm_00, RmRm_00) + 
+                g2 * (ten_dyad_prod_eigen<nsd>(RmRm_11, RmRm_11) +
+                      ten_dyad_prod_eigen<nsd>(RmRm_22, RmRm_22) +
+                2.0 * ten_dyad_prod_eigen<nsd>(RmRm_12, RmRm_12)) +
+          2.0 * g3 * (ten_dyad_prod_eigen<nsd>(RmRm_01, RmRm_01) +
+                      ten_dyad_prod_eigen<nsd>(RmRm_20, RmRm_20));
+      CC_bar = r2 * CC_bar;
```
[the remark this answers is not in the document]
ktbolt: @aabrown100-git Be careful with Eigen expressions that have the same variable on the lhs and rhs.

Code/Source/solver/mat_models.cpp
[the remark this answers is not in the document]
ktbolt: @aabrown100-git It would be good to improve readability where possible, `compute` makes sense, but be careful not to change a naming convention that could cause confusion, thus change all `get_` to `compute_`.

Code/Source/solver/mat_models.cpp
ktbolt: @aabrown100-git You don't need a different name, can use `get_pk2cc()` for the template function, the compiler will figure out which one to use.

Code/Source/solver/mat_models.cpp
aabrown100-git: Sounds good, I will go with the generalized `double_dot_product()`

Code/Source/solver/mat_models.cpp
aabrown100-git: Will do

Code/Source/solver/mat_models.cpp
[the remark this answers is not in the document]
aabrown100-git: Got it, I changed all `get_` to `compute_` in this file. There are many other `get_`s in the repository that do a little more computing than getting, but I don't think this will cause any confusion.

Code/Source/solver/mat_models.cpp
ktbolt: @aabrown100-git I bet that the overhead of this copy is insignificant given all that is going on in the material code. Multiple results can also be returned in a `struct`, not as clean as returning a pair though, maybe comment this as a TODO.

Code/Source/solver/mat_models.cpp
aabrown100-git: Ah interesting!

Code/Source/solver/mat_models.cpp
ktbolt: @aabrown100-git If you are going to leave in commented-out code then be sure to document it so developers can understand why it is there.

Code/Source/solver/mat_models.cpp
aabrown100-git: I feel like this may be a more natural place to have this check. I'm not super familiar with the code that performs the file reading, so I'm not sure where to place this check. When running a ustruct simulation with one of these material models, the simulation fails immediately and the error message is fairly clear.

Code/Source/solver/mat_models.cpp
ktbolt: @aabrown100-git My mistake, confused `EigenMatrix<nsd>` with an actual Eigen object, looked like a vector to me.

I recommend using `Matrix` rather than `EigenMatrix`, same for `Tensor` and `Vector`.

Code/Source/solver/mat_models.cpp
aabrown100-git: Got it, will leave a TODO comment

Code/Source/solver/mat_models.cpp
aabrown100-git: Got it, will leave a detailed comment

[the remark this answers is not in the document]
aabrown100-git: After discussing with Dave, I added a detailed comment explaining what this is, why it's commented out, and why I'm leaving it in for reference. Does this work?

Code/Source/solver/mat_models.cpp
aabrown100-git: Sounds good, I changed EigenMatrix to Matrix, and EigenTensor to Tensor. I actually don't use an Eigen vector so didn't need to define a Vector alias.

Code/Source/solver/sv_struct.cpp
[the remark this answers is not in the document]
aabrown100-git: Thanks for explaining @ktbolt. @mrp089, is it okay to leave these as lK(0,a,b) = lK(0,a,b) + for now?

Code/Source/solver/mat_models.cpp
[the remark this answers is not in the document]
aabrown100-git: @mrp089, okay to keep this as `CC_bar = r2 * CC_bar;`?

mrp089 (approved):

Code/Source/solver/mat_models.cpp
aabrown100-git: @ktbolt Just pinging this when you get you get a chance. After resolving this conversation I think we can merge.

ktbolt: @aabrown100-git The problem is that when running on an HPC cluster your job can wait in a queue for a long time and then immediately fail because the input is not correct, would need something like a single core preprocessing check before running the main job on a larger number of cores. 

We can see if users are having any problems with this sort of thing before implementing a solution. Users creating a job from the SV GUI should be mostly OK. 

A check should be made in the `ConstitutiveModelParameters` class I think, calling a function defined in the `mat_models` namespace. I will open an Issue for this and implement the solution.

ktbolt (approved):
It all looks good to me!

aabrown100-git:
Awesome, I will merge now. Thanks @ktbolt and @mrp089!

Issue #320, opened 2024-12-28
Add a check for a valid constitutive model for the ustruct equation


ktbolt:
### Problem

Certain constitutive models are not valid for the `ustruct` equation. A check is currently made in the `mat_models::get_pk2cc()` function for this as several if-statements embedded in a `switch` statement. 

### Solution

The check for a valid `ustruct` equation constitutive model should be made in the `ConstitutiveModelParameters` class so an error can be caught early. 

A function can be created in the `mat_models` namespace to make this check and clearly show which constitutive models are not valid.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct and Contributing Guidelines

ktbolt:
I've added the `ustruct::bool constitutive_model_is_valid(consts::ConstitutiveModelType model)` function to check for a valid constitutive model for `ustruct` equations.

This function is called from ` ConstitutiveModelParameters::check_constitutive_model()` when setting the constitutive model for an equation in `EquationParameters::set_values()`.

Setting a constitutive model that is not valid will throw for example
```
The stVenantKirchhoff constitutive model is not valid for ustruct equations.
```

PR #321, opened 2025-01-06
Check constitutive model ustruct 320

ktbolt:
Move the check for a valid material model for ustruct equations from `mat_models::get_pk2cc()` to `ConstitutiveModelParameters::check_constitutive_model()`.

See https://github.com/SimVascular/svMultiPhysics/issues/320.


Code/Source/solver/ustruct.cpp
```diff
@@ -210,6 +210,21 @@ void b_ustruct_3d(const ComMod& com_mod, const int eNoN, const double w, const V
   }
 }
 
+/// @brief Check is a constitutive model is valid for the ustruct equation.
```
aabrown100-git, line 213: Should be "Check **if** a constitutive ...", right?

aabrown100-git (approved):
Apart from that one comment typo, it all looks good to me!

Code/Source/solver/ustruct.cpp
ktbolt: @aabrown100-git Yes, `Check if` !

Issue #322, opened 2025-01-06
check missing on the inflow profile name


dcodoni:
### Description

If the boundary condition `Add_BC` `Profile` parameter is not  "Flat", "Parabolic" or "User_defined" (case sensitive), the solver does not give any warnings and the results are unexpected giving an unrelated error: Face used for a BC has no non-zero node. This happens because there is not a default case or a warning in the if statement that selects the type of flux profile.

### Reproduction

Try to run any fluid case with a flux profile and set "parabolic" (lower case p) in the input file instead of "Parabolic" (capital P).

### Expected behavior

I would expect the solver to give a warning or a descriptive runtime error saying explicitly that a match for the profile type has not been found.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct and Contributing Guidelines

ktbolt:
Merged.

PR #325, opened 2025-01-07
Check on the flux profile type name #322

dcodoni:
## Current situation
The flux profile type names accepted by the solver are: "Flat", "Parabolic" and "User_defined" (case sensitive). If the name on the input file does not match, no error is provided and eventually the simulation will crash with a non-related error.

## Release Notes 
A check is added and a descriptive runtime error will be prompt whenever a match is not found in the profile type name.


ktbolt:
@dcodoni I am wondering if it would be better to move checks like these (e.g. testing string literals) to the `Parameters` class so the error would be identified when the solver XML file was being parsed.

Issue #264, opened 2024-08-30
Terminate called after throwing an instance of 'std::runtime_error'.   what():  [set_bc_cpl]  Follower pressure load must be used for 0D coupling with struct/ustruct


HadiWiputra:
### Description

Hi All,

I have script that I ran in svFSI normally, but it couldnt run in svFSI+. It is an FSI run of an aortic dissection. The simulation runs for 50 time points and then returned the following error:
Terminate called after throwing an instance of 'std::runtime_error'.   what():  [set_bc_cpl]  Follower pressure load must be used for 0D coupling with struct/ustruct

There is no sign of divergence in the previous time points. Also, the output at time point 50 looks reasonable.
https://drive.google.com/drive/folders/1WrGsZHV0feUgy4jai7dWDa53kLU7r2WX?usp=drive_link

Regards,
Hadi

### Reproduction

Run svfsi on the .inp file, and svfsiplus on the .xml file

### Expected behavior

I expect the svFSI plus version to replicate the output of svFSI fortran.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct and Contributing Guidelines

ktbolt:
@dcodoni Could someone in the solver group take a look at this? There are a couple of users in SV Forum that have run into this problem.

dcodoni:
@ktbolt I will bring it up tomorrow at our meeting!

aabrown100-git:
Just getting around to looking at this now. I believe the issue arises when running an FSI simulation with RCR boundary conditions (probably also when coupling to genBC, cplBC, or svZeroDSolver). Unfortunately, we do not currently have a test for FSI + RCR/0D coupling, which is why we didn't catch this error. I added an FSI + RCR simulation [here](https://github.com/aabrown100-git/svMultiPhysics/tree/fsi_cplbc_bug_fix/tests/cases/fsi/pipe_RCR_3d). Based on my preliminary testing, this simulation reproduces the error at timestep 13, but only when running with multiple processors. 

@HadiWiputra when you get a chance, could you run this test (it should take less than a minute) and verify this is the error you saw?

PR #333, opened 2025-01-28
Fix bug with coupled BCs in FSI simulation

aabrown100-git:
## Current situation
Resolves https://github.com/SimVascular/svMultiPhysics/issues/264

## Release Notes 
- Fix bug where `eq.dmn[cDmn].phys` was used instead of `eq.phys` to determine the current physics (e.g. fluid, FSI, struct, etc.) within `set_bc_cpl()`. For an FSI simulation, `eq.dmn[:].phys = [phys_fluid, phys_struct]`. Using `eq.dmn[:].phys` is wrong to begin with because for an FSI simulation, the current physics within `set_bc_cpl()` should be `phys_FSI`. In addition, this causes the bug discussed in https://github.com/SimVascular/svMultiPhysics/issues/264, explained as follows. For some reason, `cDmn` can seemingly randomly switch from 0 to 1, which causes causes `eq.dmn[:].phys` to switch from `phys_fluid` to `phys_struct`, which then yields the `follower pressure load` error discussed in https://github.com/SimVascular/svMultiPhysics/issues/264. Using `eq.phys` fixes these problems because for an FSI simulation, `eq.phys = phys_FSI`.
- Add a case testing FSI + cplBC. This case is the same as the other FSI pipe_3d cases, except uses an RCR boundary condition, which internally calls `set_cpl_bc()`.

## Testing
- Add fsi/pipe_RCR_3d test case to test this combination of FSI and coupled BCs

### DIFF
--- a/Code/Source/solver/set_bc.cpp
+++ b/Code/Source/solver/set_bc.cpp
@@ -84,8 +84,7 @@ void calc_der_cpl_bc(ComMod& com_mod, const CmMod& cm_mod)
   // if (ALL(cplBC.fa.bGrp .EQ. cplBC_Dir)) RETURN
 
   // Determine current physics
-  auto& cDmn = com_mod.cDmn;
-  auto cPhys = eq.dmn[cDmn].phys;
+  auto cPhys = eq.phys;
 
   // Mechanical configuration in which to compute flowrate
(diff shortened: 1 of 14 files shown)


dcodoni (approved):
Good job!

Issue #197, opened 2024-02-22
Slow nonlinear convergence in ustruct/LV_Guccione_active


mrp089:
### Description

While adding missing results to test cases (https://github.com/SimVascular/svFSIplus/pull/175), I noticed that the test `ustruct/LV_Guccione_active` has a very slow nonlinear convergence.

### Reproduction

In `./tests`, run `pytest -vrP -k LV_Guccione_active` or have a look at [this pipeline](https://github.com/SimVascular/svFSIplus/actions/runs/8001237589/job/21852062397#step:7:7804).

```
---------------------------------------------------------------------
 Eq     N-i     T       dB  Ri/R1   Ri/R0    R/Ri     lsIt   dB  %t
---------------------------------------------------------------------
 ST 1-1  2.720e-01  [0 1.000e+00 1.000e+00 9.031e-13]  [96 -126 3]
 ST 1-2  5.520e-01  [-27 4.192e-02 4.192e-02 8.019e-13]  [95 -125 2]
 ST 1-3  8.300e-01  [-42 7.443e-03 7.443e-03 4.504e-12]  [88 -102 2]
 ST 1-4  1.108e+00  [-55 1.616e-03 1.616e-03 1.866e-11]  [82 -85 3]
 ST 1-5  1.399e+00  [-68 3.657e-04 3.657e-04 8.086e-11]  [77 -70 2]
 ST 1-6  1.678e+00  [-81 8.364e-05 8.364e-05 3.345e-10]  [72 -56 2]
 ST 1-7  1.968e+00  [-94 1.919e-05 1.919e-05 1.418e-09]  [67 -39 1]
 ST 1-8  2.254e+00  [-107 4.404e-06 4.404e-06 8.073e-09]  [61 -23 1]
 ST 1-9  2.529e+00  [-119 1.011e-06 1.011e-06 3.121e-08]  [56 -10 1]
 ST 1-10  2.793e+00  [-132 2.321e-07 2.321e-07 1.121e-07]  [50 -160 2]
 ST 1-11  3.063e+00  [-145 5.326e-08 5.326e-08 5.740e-07]  [45 -144 1]
 ST 1-12  3.331e+00  [-158 1.222e-08 1.222e-08 2.805e-06]  [40 -128 1]
 ST 1-13  3.597e+00  [-171 2.805e-09 2.805e-09 1.244e-05]  [35 -113 1]
[...]

PR #337, opened 2025-02-07
Improve ustruct/LV_Guccione_active convergence

aabrown100-git:
## Current situation
Resolve https://github.com/SimVascular/svMultiPhysics/issues/197

## Release Notes 
- Reduces momentum and continuity stabilization coefficients to improve nonlinear convergence in ustruct/LV_Guccione_active test case
- Update reference solution `result_001.vtu` accordingly
- Add README.md and animation to describe case and visualize results

## Testing
No testing required. Note however that this modifies the reference solution in ustruct/LV_Guccione_active

### DIFF
--- a/tests/cases/ustruct/LV_Guccione_active/README.md
+++ b/tests/cases/ustruct/LV_Guccione_active/README.md
@@ -1,0 +1,8 @@
+This test case simulates an idealized left ventricle (LV) with a Guccione material model
+contracting due to time-dependent active stress, and subject to a time-dependent
+pressure load on the endocardial surfaces. The problem set up can be found in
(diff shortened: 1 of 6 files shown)


dcodoni:
I think because you are adding non-linearity to the system. The supg term that control convection stability is a term: tau*delta W * U* Residual. Having a higher tau makes this non lienar term higher which can impact your convergence.

That is my opinion at least.

David
________________________________

aabrown100-git:
Building on that, it may be that the tangent matrix does not include linearizations of the stabilization terms (an inconsistent tangent), which would explain why larger stabilizations coefficients worsen nonlinear convergence.

Issue #300, opened 2024-11-08
Missing check on the number of spatial dimension


msbazzi:
### Description

Setting  the number of spatial dimensions that is not  consistent with the dimension of the mesh may cause the solver to fail or produce wrong results.

### Reproduction

Set `<Number_of_spatial_dimensions> 2 </Number_of_spatial_dimensions>` in `svMultiPhysics/tests/cases/struct/LV_NeoHookean_passive/solver.xml` causes the failure
```
libc++abi: terminating due to uncaught exception of type std::runtime_error: There may be a duplicate nodes within the mesh.
```

### Expected behavior

A check is needed to make sure the spatial dimension set for the simulation is consistent with the dimension of mesh used in the simulation.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct and Contributing Guidelines

ktbolt:
@msbazzi I modified the svMultiPhysics/tests/cases/fluid/pipe_RCR_3d/solver.xml file with 
```
  <Number_of_spatial_dimensions> 2 </Number_of_spatial_dimensions>
```
but did not see an error.

How did you test this and where in the solver code was this failing ?

msbazzi:
@ktbolt i noticed the problem when solver the struct. Here is the error i get when running svMultiPhysics/tests/cases/struct/LV_NeoHookean_passive

![Image](https://github.com/user-attachments/assets/aa584468-8a39-4968-8982-e1069a498b8c)

ktbolt:
The problem is that the spatial dimension is used to set the integration dimension in `nn::select_ele()` and the element type is then set in that function using the integration dimension and the element number of nodes

```
void select_ele(const ComMod& com_mod, mshType& mesh)
{
  // Set integration dimension.
  int insd;
  if (mesh.lShl) {
    insd = com_mod.nsd - 1;
  } else if (mesh.lFib) {
    insd = 1;
  } else {
    insd = com_mod.nsd;
  }

  // Set element properties based on integration dimension and number of element nodes.
  try {
    if (insd == 3) {
      set_3d_element_props[mesh.eNoN](insd, mesh);
    } else if (insd == 2) {
      set_2d_element_props[mesh.eNoN](insd, mesh);
    } else if (insd == 1) {
      set_1d_element_props[mesh.eNoN](insd, mesh);
    }
```
For a tet mesh
```
nsd=3 and element number of nodes=4 -> tet
nsd=2 and element number of nodes=4 -> quad
```

[...]

PR #339, opened 2025-02-10
Check spatial dimension 

ktbolt:
Check for consistency with the `Number_of_spatial_dimensions` parameter set in the solver XML file and the dimension of the elements used for the finite element mesh. https://github.com/SimVascular/svMultiPhysics/issues/300


dcodoni (approved):
Thanks @ktbolt for doing this!

ktbolt:
I messed up 1d and shells in 3d, and setting element type for a mesh defined in a vtp file. I think I got it right this time.

dcodoni (approved):

Issue #310, opened 2024-12-09
Automatically build and upload Docker containers


mrp089:
### Problem

Sometimes, we will need to rebuild the Docker images that are hosted on [Docker Hub](https://registry.hub.docker.com/u/simvascular), e.g., when external versions change. This is currently done manually.

### Solution

Have a GitHub workflow that automatically runs the [Dockerfiles](https://github.com/SimVascular/svMultiPhysics/tree/main/Docker) and uploads the images to Docker hub **when triggered**. This doesn't need to happen with every pull request since it's unnecessary and resource-consuming. Rather, the action would need to figure out for itself when it's necessary to trigger a rebuild.

`4C` has an action to [check if a Docker rebuild is required](https://github.com/4C-multiphysics/4C/blob/main/.github/actions/check_docker_build_required/action.yml). I think they do that by [comparing the hashes of their dependencies](https://github.com/4C-multiphysics/4C/blob/main/.github/actions/compute-and-check-dependencies-hash/action.yml).rebuild

### Additional context

See the discussion in #308.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct and Contributing Guidelines

PR #344, opened 2025-02-11
Only test ubuntu-latest

### LINKED ISSUE #343: Update ubuntu runners
### Problem

We currently test on the following runners:
- `ubuntu-20.04`
- `ubuntu-22.04`
- `macos-latest`

The runner `ubuntu-20.04` will be shut down on April 1, 2025.

### Solution

Remove the `ubuntu-20.04` runner from our [test matrix](https://github.com/SimVascular/svMultiPhysics/blob/c3fac1e81c774a7b1a7ca70657f43492a368b29e/.github/workflows/test_ubuntu.yml#L6-L14). In addition, it might be a nice occasion to switch our testing to the following:
- `ubuntu-latest` (currently `ubuntu-24.04`)
- `macos-latest`

### Additional context

Is there a reason we need to explicitly specify an `ubuntu` version? Using `latest` (as we do for `macos`) would prevent us from running into this problem periodically and would tell us immediately if there is a problem with upgrading.

### Code of Conduct

mrp089:
## Current situation
Closes #343 

## Release Notes 
Only testing on `ubuntu-latest` (currently `ubuntu-24.04`). Dropping `ubuntu-20.04` and `ubuntu-22.04`.

## Documentation
None for now.

## Testing
Let's see if the tests run. I have a feeling we might rename some Docker containers

### DIFF
--- a/.github/workflows/test_macos.yml
+++ b/.github/workflows/test_macos.yml
@@ -1,4 +1,4 @@
-name: macOS
+name: macos-latest
 on: [push, pull_request]
 jobs:
   test:
--- a/.github/workflows/test_ubuntu.yml
+++ b/.github/workflows/test_ubuntu.yml
@@ -3,7 +3,7 @@ on: [push, pull_request]
 jobs:
   test:
     runs-on: ubuntu-latest
-    container: simvascular/libraries:ubuntu-22.04
+    container: simvascular/libraries:ubuntu22
     steps:
       - uses: actions/checkout@v4
       - name: Build svZeroDSolver
--- a/.github/workflows/test_ubuntu.yml
+++ b/.github/workflows/test_ubuntu.yml
@@ -1,9 +1,9 @@
-name: test Ubuntu
+name: ubuntu-latest
 on: [push, pull_request]
 jobs:
   test:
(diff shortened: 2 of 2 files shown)


mrp089:
This fails as expected because there is no container for `ubuntu-latest`. I'll see if the `ubuntu-latest` test runs with the `ubuntu-22.04` container. @dcodoni, what do we need to do to generate an `ubuntu-latest` container?

dcodoni:
@mrp089 the container should run no matter what is the OS is based on. I wanted to create an Ubuntu-latest based container last time, but we decided for single versions. 
Anyway, the changes are very simple. At the top of the dockerfile we need to change FROM ubuntu:22.04 AS base
 to FROM ubuntu:latest AS base, and build the whole container again. Hopefully all the libraries versions will work fine on the latest Ubuntu.

mrp089:
Thank you, @dcodoni! Might be a good time for #310 now

mrp089:
Seems like the tests run fine! I would merge this now and then build an `ubuntu-latest` container later, maybe with an automated workflow as proposed in #310.

ktbolt (approved):
Looks good!

PR #352, opened 2025-03-05
GenBC coupling in FSI and effective direction

zinanhu0810:
## Current situation
The GenBC.int file, which is used to communicate between the 3D solver and the 0D genBC, cannot be located if the FSI equation is used. Moreover, while we have genBC coupled as a Dirichlet boundary condition and have set up an effective direction for this surface, the area of the surface may change, but the area reading in 0D doesn't, which causes incorrect averaged pressure and communication inconsistency.

## Release Notes 
Implemented an if condition to check the communication path and resized it to ensure GenBC.int can be located for all equations. The area of the Dirichlet surface coupled to genBC is updated every timestep so that the proper averaged pressure can be transferred to genBC.

## Documentation
 
*Please ensure that you properly document any additions.*

## Testing
 
*In addition, please ensure all modified and new code is covered by tests.*


ktbolt (approved):
Looks good!

PR #386, opened 2025-05-22
Reorganize readme 385

ktbolt:
Reorganize the README to follow the table of contents.


dcodoni (approved):

dcodoni:
@ktbolt would it be necessary to add something about git-lfs in the README file? For example something like: run _git-lfs install_ after cloning the repository and before compiling, if test cases need to be run.

ktbolt:
@dcodoni git-lfs and all that is described in the `testing` section but it would probably be helpful to new users to put something in the Introduction about the contents of the testing directory.

Issue #316, opened 2024-12-16
Test Linear Elastic Model and Remove Redundancy


divyaadil23:
### Problem

There seem to be 2 options for the linear elastic model - lElas and stIso_Lin. Neither of them have any test cases.

### Solution

Look at how the two models are different, and combine them into a single model. Add unit tests and/or integration tests. 

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct and Contributing Guidelines

ktbolt:
@divyaadil23 `phys_lElas` is an equation type, `stIso_LS` is a material model. 

There are tests [here](https://github.com/SimVascular/svFSIplus-Tests/tree/main/02-lelas/01-beam_Euler-Bernoulli) for `phys_lElas`.

I believe that the `stIso_LS`  material model is used for the `mesh` equation type.

mrp089:
The linear elasticity test cases are still missing in #95 and stIso_LS is [not covered](https://app.codecov.io/github/SimVascular/svMultiPhysics/blob/main/Code%2FSource%2Fsolver%2Fmat_models.cpp). It would be nice if we could finally close that. @divyaadil23 do you want to add those as a first step?

As @ktbolt said, we use a linear elastic material for mesh deformation in ALE-FSI. It looks like the `mesh` equation type uses its [own linear elastic material model](https://github.com/SimVascular/svMultiPhysics/blob/6ecb75f7144fe96ddbdcbf6c7f31c8d1b17af3ec/Code/Source/solver/l_elas.cpp#L353-L359) that [is covered](https://app.codecov.io/github/SimVascular/svMultiPhysics/blob/main/Code%2FSource%2Fsolver%2Fl_elas.cpp).

ktbolt:
I will add the test case.

ktbolt:
`stIso_LS` is for a Lee-Sacks material model for valve-like tissues that are modeled as shells. I could add a test case for this if it would be usefull.

ktbolt:
I've added a `tests/cases/linear-elasticity/beam` test taken from https://github.com/SimVascular/svFSI-Tests/tree/master/02-lelas/01-beam_Euler-Bernoulli/01-tet4/N008 even though it is not that great of a mesh, many elements with a poor aspect ratio.

PR #392, opened 2025-05-29
Add linear elastic test 

ktbolt:
I've added a linear elasticity test case. See https://github.com/SimVascular/svMultiPhysics/issues/316.


tests/cases/linear-elasticity/beam/README.md
```diff
@@ -0,0 +1,3 @@
+
+Solve a Eulear-Bernoulli beam problem with a traction boundary condition.
```
aabrown100-git, line 2: Why are we calling this an Euler-Bernoulli beam problem? I think Euler-Bernoulli implies we're using [Euler-Bernoulli beam theory](https://en.wikipedia.org/wiki/Euler%E2%80%93Bernoulli_beam_theory), but this test case is solving standard 3D linear elasticity with a beam-shaped body.

tests/cases/linear-elasticity/beam/solver.xml
```diff
@@ -0,0 +1,96 @@
...
+
+</Add_mesh>
+
+
+<Add_equation type="lElas" > 
+
+   <Coupled> true </Coupled>
+   <Min_iterations> 1 </Min_iterations>  
+   <Max_iterations> 20 </Max_iterations> 
+   <Tolerance> 1e-5 </Tolerance>
```
aabrown100-git, line 60: In other test cases, I think we're typically using a smaller tolerance (1e-9 or smaller).

tests/test_linear_elasticity.py
```diff
@@ -0,0 +1,20 @@
+from .conftest import run_with_reference
+import os
+import subprocess
+
+# Common folder for all tests in this file
+base_folder = "linear-elasticity"
+
+# Fields to test
+fields = [
```
aabrown100-git, line 9: Are other fields available to be output for linear elasticity? It looks like we're testing all possible outputs for other physics.

aabrown100-git (changes requested):

tests/cases/linear-elasticity/beam/README.md
ktbolt: @aabrown100-git This description was taken from `svFSI-Tests/02-lelas/01-beam_Euler-Bernoulli` which this test is based on. I assume it was called this because it reproduces the solution to the Euler-Bernoulli beam  equation.

tests/test_linear_elasticity.py
ktbolt: @aabrown100-git I will add strain and stress.

tests/cases/linear-elasticity/beam/solver.xml
aabrown100-git: What does TRT mean?

aabrown100-git: I agree it's good to avoid confusion, but I worry that if tolerances are set too high, then our tests may not detect bugs introduced by future commits.

mrp089:
Loosening (linear or nonlinear) tolerances can prevent us from detecting bugs or cause different results on different machines (that's why we made them stricter in existing tests).

We could introduce examples for more ready-to-use simulations. These could still be executed to ensure they always run, but not compare the results to a reference.

tests/cases/linear-elasticity/beam/solver.xml
mrp089: Loosening (linear or nonlinear) tolerances can prevent us from detecting bugs or cause different results on different machines (that's why we made them stricter in existing tests).

We could introduce examples for more ready-to-use simulations. These could still be executed to ensure they always run, but not compare the results to a reference.

ktbolt:
@aabrown100-git Let's merge this.

Issue #382, opened 2025-05-20
Move svZeroD_interface.dat information into the solver input XML file


ktbolt:
### Problem

The `svZeroD_interface.dat` file is used to specify the parameters used to interface to the svZeroDSolver
- Location of the svZeroDSolver shared library
- Input JSON file
- Initialization parameters

This file has its own (undocumented) format that is not obvious: parameter name and value are separated by a `:` but the value must be on the next line.
```
<PARMERTER>:
<VALUE>
```
There is no error checking for mis-spelled parameter names or invalid types
```
External coupling block initial flows (one number is provided, it is applied to all coupling blocks):
"A"

```
does not produce an error.

### Solution

Move the parameters used to interface to the svZeroDSolver into the solver input XML file.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct and Contributing Guidelines

ktbolt:
I've replaced the `svZeroD_interface.dat` file
```
interface library path:
/usr/local/sv/svZeroDSolver/2024-09-26/lib/libsvzero_interface.dylib

svZeroD input file:
svzerod_3Dcoupling.json

svZeroD external coupling block names to surface IDs (where surface IDs are from *.svpre file):
RCR_coupling 0

Initialize external coupling block flows:
0

External coupling block initial flows (one number is provided, it is applied to all coupling blocks):
0.0

Initialize external coupling block pressures:
1

External coupling block initial pressures (one number is provided, it is applied to all coupling blocks):
0.0
```

with the `svZeroDSolver_interface` XML element
```
   <svZeroDSolver_interface name="svzerod-interface">
     <Coupling_type> SI </Coupling_type>
     <Configuration_file> svzerod_3Dcoupling.json </Configuration_file>
     <Shared_library> /usr/local/sv/svZeroDSolver/2024-09-26/lib/libsvzero_interface.dylib </Shared_library>
     <Block_to_surface_map>
        RCR_coupling 0
     </Block_to_surface_map>
     <Initial_flows> 0.0 </Initial_flows>
     <Initial_pressures> 0.0 </Initial_pressures>
   </svZeroDSolver_interface>
```

This now provides all of the standard XML error checking: required parameters, unknown keyword names, improper values, etc. And the order of the parameters are not fixed like for the `svZeroD_interface.dat` file. 

These parameters are stored in the `svZeroDSolverInterfaceType` class defined in `ComMod.h` and an object of this type as added to `cplBCType`.

[...]

PR #396, opened 2025-06-04
Move svzerodinterface file into solver xml

ktbolt:
I've moved the information in the `svZeroD_interface.dat` file into the solver XML `svZeroDSolver_interface` parameter and added a `svZeroDSolver_block` to `Add_BC` to identify the svZeroDSolver coupling block. https://github.com/SimVascular/svMultiPhysics/issues/382


Code/Source/solver/Parameters.h
```diff
@@ -1450,6 +1457,7 @@ class Parameters {
     void set_mesh_values(tinyxml2::XMLElement* root_element);
     void set_precomputed_solution_values(tinyxml2::XMLElement* root_element);
     void set_projection_values(tinyxml2::XMLElement* root_element);
+    void set_szsolver_interface_values(tinyxml2::XMLElement* root_element);
```
chlookaburra, line 1460: I am wondering if `szsolver` should be written as `svzerodsolver` to stay consistent with the full name functions

chlookaburra (changes requested):
Thank you @ktbolt ! I left one comment in the following line to see what you think.

chlookaburra (approved):

Issue #338, opened 2025-02-10
Split up unit tests


mrp089:
### Problem

All our unit tests are in one long (>2000 lines) [`.cpp` file](https://github.com/SimVascular/svMultiPhysics/blob/main/tests/unitTests/test.cpp). This makes it hard to see what we're currently testing and intimidating to add new tests.

### Solution

Split up the test file logically into many `.cpp` files (like we do for integration tests). @arsLibera [added code](https://github.com/SimVascular/svOneDSolver/blob/4aac52ce70115d0274d99a54f2c69ad5fb928354/CMakeLists.txt#L114-L181) for the 1D solver to collect several unit test files.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct and Contributing Guidelines

aabrown100-git:
I agree. Also, currently our "unit tests" are rather complicated tests of the solid material models (this was my doing). These might better be split up and placed in a folder called something more appropriate, and leave "unit tests" for true unit tests.

PR #413, opened 2025-07-10
Split up unit tests

aabrown100-git:
## Current situation
Resolves https://github.com/SimVascular/svMultiPhysics/issues/338

## Release Notes 
- Split up code originally in `unitTests/test.h` and `unitTests/test.cpp` into separate files. These contained tests for different material models. I create separate `test_material_xxx.h` and `test_material_xxx.cpp` files for each material model.
- Put all material model tests into a subfolder `material_model_tests`
- Added `README.md` describing unit tests
- Modified `CMakeLists.txt` to automatically find all `*.cpp` files under `unitTests/` and compile and run them when unit testing is performed.

## Documentation
- `README.md` provides documentation for unit testing

## Testing
- Verified that all original tests (127 tests from 16 test suites) in `test.cpp` were copied into the new `test_material_xxx.cpp` files.


tests/unitTests/material_model_tests/test_material_CANN_holzapfel_ogden.h
```diff
@@ -0,0 +1,192 @@
...
+
+
+// Class to contain CANN model with Holzapfel-Ogden material parameters
+class CANN_HO_Params : public MatParams {
+public:
+    std::vector<CANNRow> Table;
+    // Define fiber directions
+    double f[3];    // Fiber direction
+    double s[3];    // Sheet direction
+
```
ktbolt, line 44: If you are defining a lot of variables of the same type then you can use 
```
using Direction = double[3];
Direction f;
Direction s;
```

tests/unitTests/material_model_tests/test_material_CANN_holzapfel_ogden.h
```diff
@@ -0,0 +1,192 @@
...
+    // Default constructor
+    CANN_HO_Params() {
+
+        // Resize Table to ensure there's at least 4 elements
+        Table.resize(4);  // Ensure there's space for 4 rows
```
ktbolt, line 49: It is better to give `4` a name.

tests/unitTests/material_model_tests/test_material_CANN_neohookean.cpp
```diff
@@ -0,0 +1,316 @@
...
+        }
+    }
+
+    // TearDown method to clean up after each test, if needed
+    void TearDown() override {
+        // Clean up the test objects
+        delete TestNH;
+        TestNH = nullptr;
+        delete TestCANNNH;
+        TestCANNNH = nullptr;
```
ktbolt, line 85: Why do we need to set these guys to nullptr ? Isn't `SetUp()` always called to assign them a value ?

tests/unitTests/material_model_tests/test_material_CANN_neohookean.cpp
```diff
@@ -0,0 +1,316 @@
...
+        TestCANNNH->compute_pk2cc(F,S_ref,Dm); // Computing S_ref from CANN
+        TestNH->testPK2StressConvergenceOrderAgainstReference(F,S_ref, delta_max, delta_min, order, convergence_order_tol, verbose);
+    }
+}
+
+// Test order of convergence of consistency of material elasticity for random F (small)
+TEST_F(STRUCT_CANNNeoHookeanTest, TestMaterialElasticityConsistencyConvergenceOrderAgainstNHRandomFSmall) {
+    verbose = false; // Show order of convergence, errors, F, S
+
+    // Loop over F in F_small_list
```
ktbolt, line 206: A comment of the totally obvious.

tests/unitTests/material_model_tests/test_material_CANN_neohookean.cpp
```diff
@@ -0,0 +1,316 @@
...
+    for (auto F_std : F_small_list) {
+        // Convert to C array
```
ktbolt, line 208: This is creating a copy of the elements of `F_small_list`. 

And another comment of the obvious. The function name `convertToArray` says it all.

tests/unitTests/material_model_tests/test_material_CANN_neohookean.h
```diff
@@ -0,0 +1,156 @@
...
+
+    // Constructor with parameters
+    CANN_NH_Params(std::vector<CANNRow> TableValues) {
+        for (int i = 0; i < 1; i++){
+            this -> Table[i].invariant_index = TableValues[i].invariant_index;
+            this -> Table[i].activation_functions = TableValues[i].activation_functions;
+            this -> Table[i].weights = TableValues[i].weights;
+        }     
+    };
+
```
ktbolt, line 61: Use assignment here.

tests/unitTests/material_model_tests/test_material_common.h
```diff
@@ -303,7 +277,7 @@ solidMechanicsTerms calcSolidMechanicsTerms(const Array<double> &F) {
  * @param y y data points.
  * @return std::pair<double, double> A pair containing the slope (m) and the y-intercept (b).
  */
-std::pair<double, double> computeLinearRegression(const std::vector<double>& x, const std::vector<double>& y) {
+inline std::pair<double, double> computeLinearRegression(const std::vector<double>& x, const std::vector<double>& y) {
```
ktbolt, line 280: All of these `inline` qualifications can be removed.

tests/unitTests/material_model_tests/test_material_holzapfel_ogden_MA.cpp
```diff
@@ -0,0 +1,499 @@
...
+        params.b_fs = 11.436; // no units
+        params.k = 100.0; // no units
+
+        // Set random values for f between 0 and 1 and normalize
+        params.f[0] = getRandomDouble(0.0, 1.0);
+        params.f[1] = getRandomDouble(0.0, 1.0);
+        params.f[2] = getRandomDouble(0.0, 1.0);
+        double norm_f = sqrt(params.f[0]*params.f[0] + params.f[1]*params.f[1] + params.f[2]*params.f[2]);
+        params.f[0] /= norm_f; params.f[1] /= norm_f; params.f[2] /= norm_f;
+
```
ktbolt, line 72: Using `std::array` for `params.f` would allow more compact and cleaner code, get rid of all of the indexing
```
auto norm = sqrt(std::inner_product(f.begin(), f.end(), f.begin(), 0));
std::for_each(f.begin(), f.end(), [prod](double& n) { n /= prod; });
```

tests/unitTests/material_model_tests/test_material_holzapfel_ogden_MA.cpp
```diff
@@ -0,0 +1,499 @@
...
+    TestHO_ma->testPK2StressConvergenceOrder(F, delta_max, delta_min, order, convergence_order_tol, verbose);
+}
+
+// Test order of convergence between finite difference PK2 stress and compute_pk2cc() PK2 stress for random F (small)
+TEST_F(STRUCT_HolzapfelOgdenMATest, TestPK2StressConvergenceOrderRandomFSmall) {
+    //verbose = true; // Show order of convergence, errors, F, S
+
+    // Loop over F in F_small_list
+    for (auto F_std : F_small_list) {
+        // Convert to Array
```
ktbolt, line 196: `F_std` is a copy, use `const auto& F_std`.

tests/unitTests/material_model_tests/test_material_holzapfel_ogden_MA.cpp
```diff
@@ -0,0 +1,499 @@
...
+
+    // Loop over F in F_medium_list
+    for (auto F_std : F_medium_list) {
+        // Convert to Array
+        convertToArray(F_std, F);
+
+        // Set convergence order tolerances larger and special delta_max/delta_min to get this test to pass
+        convergence_order_tol = 0.15;
+        delta_max = 8e-6;
+        delta_min = 2e-6;
```
ktbolt, line 297: These should be placed somewhere visible in the class where developers can see them. You might should have a class containing all of the tolerances and such.

tests/unitTests/material_model_tests/test_material_holzapfel_ogden_MA.cpp
```diff
@@ -0,0 +1,499 @@
...
+                        {0.0, 0.8, 0.0},
+                        {0.0, 0.0, 1.0}};
+
+    // Check order of convergence of consistency of material elasticity
+    TestHO_ma->testMaterialElasticityConsistencyConvergenceOrder(F, delta_max, delta_min, order, convergence_order_tol, verbose);
+}
+
+// Test order of convergence of consistency of material elasticity for random F (small)
+TEST_F(USTRUCT_HolzapfelOgdenMATest, TestMaterialElasticityConsistencyConvergenceOrderRandomFSmall) {
+    //verbose = true; // Show order of convergence, errors, F, S
```
ktbolt, line 461: A lot of these functions have the same structure and are repeated all over.

ktbolt (changes requested):
@aabrown100-git Very good work !

ktbolt:
@aabrown100-git A few general comments about the unit tests.

1) I would be good to see more comments describing what the tests are doing and why, and any assumptions, and what criteria is used to pass or fail the test.

2) The tolerances used to pass a test should be constants set in a visible location, like at the top of the class.

3) Function and variable names do not conform to our coding standards: they should use snake case.

4) Function names are much too long (e.g. `TestPK2StressConvergenceOrderTriaxialStretch`) 
    
5) There are lots of useless comments
```
    // Loop over F in F_small_list
    for (auto F_std : F_small_list) {
        // Convert to Array
        convertToArray(F_std, F);
```
 
6) Why is `convertToArray` need to be used ?

tests/unitTests/material_model_tests/test_material_CANN_neohookean.cpp
aabrown100-git: Removing extraneous comment

tests/unitTests/material_model_tests/test_material_common.h
aabrown100-git: When I remove `inline`, I get a linker error `duplicate symbol 'computeLinearRegression(...)`. This is because `test_material_common.h` is included in all material test header files, so without the `inline` the function is defined multiple times.

ktbolt: @aabrown100-git Why are function implementations defined in a header file rather than an implementation file ?

tests/unitTests/material_model_tests/test_material_CANN_neohookean.cpp
aabrown100-git: Modified all instances to create a const reference of elements of F_small_list

tests/unitTests/material_model_tests/test_material_holzapfel_ogden_MA.cpp
aabrown100-git: Done, modified all other instances

tests/unitTests/material_model_tests/test_material_holzapfel_ogden_MA.cpp
```diff
@@ -0,0 +1,499 @@
...
+}
+
+// Test order of convergence between finite difference PK2 stress and compute_pk2cc() PK2 stress for random F (medium)
+TEST_F(STRUCT_HolzapfelOgdenMATest, TestPK2StressConvergenceOrderRandomFMedium) {
+    //verbose = true; // Show order of convergence, errors, F, S
+
+    // Loop over F in F_medium_list
+    for (auto F_std : F_medium_list) {
+        // Convert to Array
+        convertToArray(F_std, F);
```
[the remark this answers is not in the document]
aabrown100-git: Thanks for the suggestion, did this. Looks much cleaner now

tests/unitTests/material_model_tests/test_material_CANN_holzapfel_ogden.h
aabrown100-git: Defined these as Vector<double> to address a different comment. I think this is concise and clear enough

tests/unitTests/material_model_tests/test_material_holzapfel_ogden_MA.cpp
ktbolt: @aabrown100-git I've not used Google Test before so I don't know how tests can be organized. 

If each material model has essentially the same test then that would suggest using templates (somehow!) and being able to perform a test on a list of fixtures I guess.

You can think about this as what new code is required when adding (1) a new material model or (2) a new test. 

But it all works so we can probably think about this another time.

tests/unitTests/material_model_tests/test_material_holzapfel_ogden_MA.cpp
aabrown100-git: Yeah, I had to set these special tolerances for this particular test to get the test to pass. Totally hacky I know, but can't think of a better way to do it. In the most recent code, the default value of `convergence_order_tol` is 0.02, and for a few tests I have to set `convergence_order_tol` to 0.03 or 0.04. [The default value is defined in the generic class `MaterialTestFixture`](https://github.com/aabrown100-git/svMultiPhysics/blob/eb511b7a0110783c284fd9529d6b3185a916f50d/tests/unitTests/material_model_tests/test_material_common.h#L1764C5-L1764C106), from which each material model test fixture class is derived. Is this what you had in mind for a class containing all these tolerances and related values?

tests/unitTests/material_model_tests/test_material_holzapfel_ogden_MA.cpp
```diff
@@ -0,0 +1,499 @@
...
+        }
+
+        // Normalize s
+        double norm_s = sqrt(params.s[0]*params.s[0] + params.s[1]*params.s[1] + params.s[2]*params.s[2]);
+        params.s[0] /= norm_s; params.s[1] /= norm_s; params.s[2] /= norm_s;
+
+        // Check f.s = 0
+        double dot_fs = params.f[0]*params.s[0] + params.f[1]*params.s[1] + params.f[2]*params.s[2];
+        if (fabs(dot_fs) > 1e-6) {
+            std::cout << "f.s = " << dot_fs << std::endl;
```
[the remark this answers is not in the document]
aabrown100-git: Thanks for pointing this out. I've replaced `1e-6` with `std::numeric_limits<double>::epsilon()`

tests/unitTests/material_model_tests/test_material_CANN_neohookean.h
aabrown100-git: Could you sketch out what it would look like to use assignment? I don't fully understand.

tests/unitTests/material_model_tests/test_material_CANN_holzapfel_ogden.h
aabrown100-git: @divyaadil23 Could you comment on how to address?

tests/unitTests/material_model_tests/test_material_holzapfel_ogden_MA.cpp
ktbolt: This is what I had in mind!  

Maybe declare them `const` to signal that the values of these guys are only set in the class. 

It is also a C++ convention to use uppercase letters for constants: CONVERGENCE_ORDER_TOL

And you might could group some of these into classes
```
class Tolerance {
  public:
    static constexpr double RELATIVE = 1e-3; // relative tolerance for comparing values
    static constexpr double ABSOLUTE = 1e-11; // absolute tolerance for comparing values
    static constexpr double CONVERGENCE_ORDER_TOL = 0.02; // Tolerance for comparing convergence order with expected value
};

class DeformationPerturbation {
  public:
    static constexpr double SMALL = 0.003; // Small perturbation factor
    static constexpr double MEDIUM = 0.03; // Medium perturbation factor
    static constexpr double LARGE = 0.3; // Large perturbation factor
};

if (value > Tolerance::RELATIVE) { 
}
```
Then it is clear what these guys are and you don't need to use a suffix like _TOL.

tests/unitTests/material_model_tests/test_material_CANN_neohookean.h
ktbolt: `Table` and `TableValues` are `std::vectors` so they have a copy assignment operator which will copy the data from one to another using `=`
```
Table = TableValues;
```
assuming that `CANNRow`  contains objects that can be copied (i.e. no pointers).

tests/unitTests/material_model_tests/test_material_CANN_neohookean.cpp
divyaadil23: I followed what was done for the other tests (e.g. holzapfel_ogden tests)

tests/unitTests/material_model_tests/test_material_CANN_holzapfel_ogden.h
```diff
@@ -0,0 +1,192 @@
...
+            dmn.stM.paramTable.activation_functions(i,1) = params.Table[i].activation_functions.value_[1];
+            dmn.stM.paramTable.activation_functions(i,2) = params.Table[i].activation_functions.value_[2];
+
+            // Store weight values
+            dmn.stM.paramTable.weights(i,0) = params.Table[i].weights.value_[0];
+            dmn.stM.paramTable.weights(i,1) = params.Table[i].weights.value_[1];
+            dmn.stM.paramTable.weights(i,2) = params.Table[i].weights.value_[2];
+
+        }
+
```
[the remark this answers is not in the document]
divyaadil23: I think there was a data type mismatch in the assignment if I didn't use the field value_  for VectorParameter

tests/unitTests/material_model_tests/test_material_CANN_holzapfel_ogden.h
aabrown100-git: Nice!

tests/unitTests/material_model_tests/test_material_CANN_holzapfel_ogden.h
```diff
@@ -0,0 +1,192 @@
...
+        Vector<double> s = {params.s[0], params.s[1], params.s[2]};
+        fN.set_col(0, f);
+        fN.set_col(1, s);
+    }
+
+/**
+     * @brief Prints the CANN HO material parameters.
+     */
+    void printMaterialParameters() {
+        int nrows = 4;
```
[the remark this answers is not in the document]
aabrown100-git: Replaced with params.nrows, which is set to 4 for the CANN HO model

tests/unitTests/material_model_tests/test_material_CANN_neohookean.cpp
```diff
@@ -0,0 +1,316 @@
...
+        params_CANN_NH.Table[0].activation_functions.value_ = {1,1,1};
+        params_CANN_NH.Table[0].weights.value_ = {1.0,1.0,C10};
+
+        // Initialize the test objects
+        TestNH = new TestNeoHookean(params_NH);
+        TestCANNNH = new TestCANN_NH(params_CANN_NH);
+
+        if (TestCANNNH == nullptr) {
+            throw std::runtime_error("TestCANNNH is not properly initialized");
+        }
```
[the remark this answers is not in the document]
aabrown100-git: Removing these if statements, also in `CANN_holzapfel_ogden.cpp`

tests/unitTests/material_model_tests/test_material_CANN_neohookean.cpp
aabrown100-git: Yes, I don't think its strictly necessary. I asked Cursor and it says its a defensive programming practice but not strictly necessary. @ktbolt happy to remove, just let me know what you think is best.

tests/unitTests/material_model_tests/test_material_CANN_holzapfel_ogden.h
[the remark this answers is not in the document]
aabrown100-git: This seems to be specific to @divyaadil23 CANN material model implementation. Maybe we can push this to a separate issue and merge the current PR?

tests/unitTests/material_model_tests/test_material_CANN_neohookean.cpp
aabrown100-git: Okay makes sense, so remove just the `TestNH = nullptr` and all similar lines?

ktbolt: Yes please.

tests/unitTests/material_model_tests/test_material_CANN_holzapfel_ogden.h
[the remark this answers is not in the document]
aabrown100-git: Sounds good, resolving this conversation

Issue #347, opened 2025-02-12
Add cardiac mechanics benchmark


mrp089:
### Problem

Fortran svFSI participated in an [elastodynamic software benchmark for cardiac mechanics](https://doi.org/10.1016/j.cma.2024.117485).

### Solution

It would be nice to have one or all of the test setups as a part of our test cases (or as our first example/tutorial). Since this example is well documented, it would be a nice resource to point new users to.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct and Contributing Guidelines

aabrown100-git:
@dseyler has already set up this case in svMultiPhysics. I believe he also wrote up something to be added to the svMultiPhysics User Guide. 

@dseyler, could you share the case here, and we can discuss where to put it?

PR #419, opened 2025-07-29
Cardiac mechanics benchmark

dseyler:
Cardiac Mechanics Benchmark test cases

## Current situation
Added two new test cases (struct + ustruct) based on [this cardiac elastodynamics benchmark paper ](https://doi.org/10.1016/j.cma.2024.117485). This addresses [issue 347] (https://github.com/SimVascular/svMultiPhysics/issues/347).

## Release Notes 

- Added struct/LV_HolapfelOgden_active and uruct/LV_HolapfelOgden_active test case
- Test cases include plots from the benchmark paper as well as a PV loop.
- Includes scripts to generate pressure and active stress waveforms as well as post processing script.


tests/cases/struct/LV_HolzapfelOgden_active/README.md
```diff
@@ -0,0 +1,19 @@
+This test case simulates an idealized left ventricle with a Holzapfel-Ogden material model
+contracting due to time-dependent active stress, and subject to a time-dependent
+pressure load on the endocardial surface. The full problem is described in
+case 1A of the cardiac elastodynamcis benchmark paper by  Aróstica et al. (2025)[1]. A comparison of the displacement of two points throughout the cardiac cycle as computed by multiple solvers including svMultiphysics (SimVascular) is shown below:
+
+![Displacement Benchmark](1-s2.0-S0045782524007394-gr8_lrg.jpg)
+
+Aditionally, we present a pressure-volume loop of the idealied left ventricle. Note that the time-dependent pressure load in this problem is not intended to reflect physiological conditions.
```
aabrown100-git, line 8: should be idealized

tests/cases/ustruct/LV_HolzapfelOgden_active/README.md
```diff
@@ -0,0 +1,19 @@
+This test case simulates an idealized left ventricle with a Holzapfel-Ogden material model
+contracting due to time-dependent active stress, and subject to a time-dependent
+pressure load on the endocardial surface. The full problem is described in
+case 1A of the cardiac elastodynamcis benchmark paper by  Aróstica et al. (2025)[1]. A comparison of the displacement of two points throughout the cardiac cycle as computed by multiple solvers including svMultiphysics (SimVascular) is shown below:
+
+![Displacement Benchmark](1-s2.0-S0045782524007394-gr8_lrg.jpg)
+
+Aditionally, we present a pressure-volume loop of the idealied left ventricle. Note that the time-dependent pressure load in this problem is not intended to reflect physiological conditions.
```
aabrown100-git, line 8: should be idealized

tests/cases/struct/LV_HolzapfelOgden_active/README.md
aabrown100-git, line 8: typo

tests/cases/struct/LV_HolzapfelOgden_active/README.md
```diff
@@ -0,0 +1,19 @@
+This test case simulates an idealized left ventricle with a Holzapfel-Ogden material model
+contracting due to time-dependent active stress, and subject to a time-dependent
+pressure load on the endocardial surface. The full problem is described in
+case 1A of the cardiac elastodynamcis benchmark paper by  Aróstica et al. (2025)[1]. A comparison of the displacement of two points throughout the cardiac cycle as computed by multiple solvers including svMultiphysics (SimVascular) is shown below:
```
aabrown100-git, line 4: typo

tests/cases/struct/LV_HolzapfelOgden_active/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+![Displacement Benchmark](1-s2.0-S0045782524007394-gr8_lrg.jpg)
```
aabrown100-git, line 6: Can you also show displacements.png? Actually, not sure if the figure with all the solver solutions is necessary at all. Is this just from the benchmark paper?

tests/cases/struct/LV_HolzapfelOgden_active/Step_1_US_P1_h5.npz
aabrown100-git, line 1: where is this data from? Is this provided by the benchmark organizers?

tests/cases/struct/LV_HolzapfelOgden_active/solver.xml
```diff
@@ -0,0 +1,132 @@
...
+     <afs> 216.0 </afs> <!-- Pa -->
+     <bfs> 11.436 </bfs>
+     <k> 100.0 </k>
+   </Constitutive_model>
+
+   <Dilational_penalty_model> ST91 </Dilational_penalty_model>
+   <Penalty_parameter> 1e6 </Penalty_parameter> <!-- Pa -->
+   
+   <Momentum_stabilization_coefficient> 1e-5 </Momentum_stabilization_coefficient>
+   <Continuity_stabilization_coefficient> 1e-5 </Continuity_stabilization_coefficient>
```
aabrown100-git, line 82: You can remove this, these stabilization coefficients are not used in a struct simulation

tests/cases/ustruct/LV_HolzapfelOgden_active/README.md
aabrown100-git, line 8: typo

tests/cases/struct/LV_HolzapfelOgden_active/README.md
aabrown100-git, line 4: Strictly speaking, svMultiPhysics was not used to run the data presented in the benchmark. svFSI was used, and the results are not exactly the same with svMultiPhysics. I think you should say this to avoid confusion.

tests/cases/ustruct/LV_HolzapfelOgden_active/README.md
```diff
@@ -0,0 +1,19 @@
+This test case simulates an idealized left ventricle with a Holzapfel-Ogden material model
+contracting due to time-dependent active stress, and subject to a time-dependent
+pressure load on the endocardial surface. The full problem is described in
+case 1A of the cardiac elastodynamcis benchmark paper by  Aróstica et al. (2025)[1]. A comparison of the displacement of two points throughout the cardiac cycle as computed by multiple solvers including svMultiphysics (SimVascular) is shown below:
+
+![Displacement Benchmark](1-s2.0-S0045782524007394-gr8_lrg.jpg)
```
aabrown100-git, line 6: This file is not in this folder, so it doesn't show up when the README.md is rendered.

tests/cases/struct/LV_HolzapfelOgden_active/solver.xml
```diff
@@ -0,0 +1,132 @@
...
+  <Fiber_direction_file_path> mesh/fibersSheet.vtu </Fiber_direction_file_path>
+
+  <Mesh_scale_factor> 1.0 </Mesh_scale_factor>
+</Add_mesh>
+
+
+<Add_equation type="struct" > 
+
+   <Coupled> true </Coupled>
+   <Min_iterations> 6 </Min_iterations>
```
aabrown100-git, line 54: I looked at the convergence history for this case and the residual plateaus at around 1e-9 for several iterations. The linear solver isn't doing anything for the last several iterations. I think you can make <Min_iterations> less, maybe 3. Also, try reducing the linear solver tolerance to 1e-12

aabrown100-git (changes requested):
Looks good! Just made a few suggestions for things to change.

tests/cases/struct/LV_HolzapfelOgden_active/README.md
aabrown100-git: Could you check the benchmark paper? I'm pretty sure the data is available as supplementary information. Maybe you can link to and plot that data instead.

tests/cases/struct/LV_HolzapfelOgden_active/README.md
ktbolt: @aabrown100-git How do the svFSI and svMultiPhysics results differ, around 1% or more ?

tests/cases/struct/LV_HolzapfelOgden_active/solver.xml
ktbolt: In `gmres_v` the initial norm is compared to `ls.absTol` so maybe you can set `Absolute_tolerance` parameter.

tests/cases/struct/LV_HolzapfelOgden_active/README.md
dseyler: The data linked in the paper is the same is in the .jpg from the paper that is currently shown in the ReadMe. I suppose I could download that and plot our results on top of it

tests/cases/struct/LV_HolzapfelOgden_active/README.md
dseyler: Would it be better to download the results from the benchmark paper and plot them alongside the results of the svMultiPhysics results?

tests/cases/struct/LV_HolzapfelOgden_active/solver.xml
dseyler: Updated the solver file to include both these suggestions and it no longer plateaus

tests/cases/struct/LV_HolzapfelOgden_active/solver.xml
```diff
@@ -11,7 +11,7 @@
 
   <Save_results_to_VTK_format> 1 </Save_results_to_VTK_format> 
   <Name_prefix_of_saved_VTK_files> result </Name_prefix_of_saved_VTK_files> 
-  <!-- <Save_results_in_folder> results </Save_results_in_folder> -->
+  <!-- <Save_results_in_folder> results </Save_results_in_folder>  -->
```
ktbolt, line 14: It is better not to have commented out lines.

ktbolt (commented):
Remove the `Mesh_scale_factor` parameter, it does not do anything.

tests/cases/struct/LV_HolzapfelOgden_active/README.md
aabrown100-git: Yes, I think it would be better to plot the public ally available benchmark results alongside our own.

aabrown100-git: @ktbolt when we compare svMultPhysics with the benchmark data, we'll be able to see how much it differs from svFSI (labeled SimVascular in the benchmark data)

ktbolt: @aabrown100-git Very good, let me know. If the results are different then I can dig into it.

dseyler: @ktbolt @aabrown100-git Updated the test cases to plot the results alongside the publicly available benchmark data. There's definitely a big discrepancy between our results and the others. Unless I'm missing something, I'm pretty sure our parameters match exactly what's specified in the paper

ktbolt: @dseyler @aabrown100-git  Was svMultiPhysics updated with the latest svFSI fixes ? Did you do a run using the latest svFSI ?

svFSI also used Trilinos in the paper.

aabrown100-git: If I recall correctly, using the exact same mesh as the benchmark organizers was important. @dseyler can you check if the mesh you're using is exactly the same as the one provided by the bechmark?

@vvedula22 could you comment on possible sources of the discrepancy? I recall storing fiber directions at Gauss points instead of at cells.

vvedula22: In our sensitivity analysis, the biggest source of discrepancy was the mesh and its resolution, whether we created it or are using the mesh the organizers provided. I'd recommend using the exact mesh the organizers provided and the 5mm case.

To some extent, fiber direction computation also matters. We tried using a semi-analytical approach that relies on a Newton method to estimate transmural distance versus using the Laplace solution. The latter was the one proposed by the organizers.

Other options, including the location of where the fibers are stored, linear solver choices (as long as the nonlinear solver converges to the set tolerance), stabilization parameters (as long as they are reasonably small), etc., have minimal effects.

dseyler: Thanks @vvedula22. I'm using the same mesh as provided by the organizers but am currently computing the fiber directions using your semi-analytical approach. I haven't tried the method provided by the organizers yet though. Did you ever compare that method to your own and how much of a difference did it seem to make?

vvedula22: Yes, we did such a comparison. The differences can be substantial, especially at coarser meshes (e.g., 5mm case). You should change it to using the Laplace solution for computing transmural thickness and fiber directions.

aabrown100-git: Thanks @vvedula22. Do you think the discrepancy in the plots below could be explained by fiber locations at cells vs. Gauss points, or other smaller variations?
comparison_plots_p0_p1_step_1_nonblinded

tests/cases/struct/LV_HolzapfelOgden_active/figures.py
```diff
@@ -0,0 +1,475 @@
...
+        filename = Path(filename)
+
+    if not filename.exists():
+        raise FileNotFoundError(f"File {filename} not found")
+
+    with open(filename.as_posix(), "rb") as fl:
+        return pickle.load(fl)
+
+
+def get_array_at_point(mesh, point, array_name):
```
aabrown100-git, line 26: This might be an issue. The sample point is not necessarily on a mesh point. More likely, it's inside a cell. Instead of finding the closest point, I think it might be better to use the PyVista sample method, which would interpolate the array values from mesh points to the sample point.

tests/cases/struct/LV_HolzapfelOgden_active/figures.py
```diff
@@ -0,0 +1,475 @@
...
+import numpy as np
+import matplotlib.pyplot as plt
+import seaborn as sns
+import os
+import pyvista as pv
+from scipy.interpolate import interp1d
+
+sns.set_theme()
+
+DATA_PATH = Path(__file__).parent / "data"
```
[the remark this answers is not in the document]
dseyler: I could include the benchmark data but the files are quite large. The solver.xml file would also have to be changed to generate the data in the plot (it only runs for one time step for the benchmark).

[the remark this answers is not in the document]
dseyler: Provided a link to the benchmark data folder in the README and updated the plotting script to throw an error if the dataset is not in the directory.

aabrown100-git:
The code itself looks good. It's still failing the macos and linux tests. The macos issue seems to be affecting all current PRs. Not sure about the linux issue, I'm rerunning the linux test now.

dseyler:
@aabrown100-git @ktbolt Looks like it's still failing macos and linux. Any ideas why?

PR #420, opened 2025-07-30
Fix boundary writing issue

zinanhu0810:
For simulation with multiple meshes, the output txt file should be closed after it loops over all meshes.

## Current situation
The boundary‐integral output file was being closed inside the per‑mesh loop, triggering a “double free or corruption” error and crashing the solver on the first iteration. 

## Release Notes 
The file is now closed only after iterating over all meshes.


ktbolt (approved):
Looks good.

PR #424, opened 2025-08-25
Trilinos refactoring and updating

dcodoni:
The Trilinos implementation is updated to use the latest packages available in the library.

## Current situation
Currently the Trilinos implementation relies on archived, not supported or deprecated packages. In particular, it is based on the following packages:
- Epetra: legacy linear algebra library in Trilinos for dense and sparse matrices and vectors. It has limited flexibility, making it harder to maintain and extend, for example if a custom preconditioner needs to be implemented in the trilinos environment (Planned Archival/Deprecation)
- AztecOO: iterative linear solver package in Trilinos, it lacks some linear solvers such as FGMRES (Planned Archival/Deprecation)
- Ifpack: legacy preconditioner package for Epetra-based matrices (Planned Archival/Deprecation)
- ML: multilevel preconditioner package (Planned Archival/Deprecation)
Since Ifpack and ML are deprecated, they are not developed anymore and new optimized preconditioners will not be available in our solver.

## Release Notes 
Packages updated:
- Epetra -> Tpetra: actively developed, templated, and Kokkos-enabled for GPU/CPU portability.
- AztecOO -> Belos: supported modern iterative solver framework with flexible operator support.
- Ifpack -> Ifpack2: supported templated preconditioner package compatible with Tpetra.
- ML -> MueLu: actively developed multigrid preconditioner framework with advanced scalability (parameters are tuned for working on a general fsi problem).
The core package Teuchos is used for memory management, communication utilities, parameter lists, and interoperability across Trilinos packages.

All updated packages are wrapped under a Kokkos Node abstraction, enabling the same code to target different execution spaces. This provides the possibility of running for example on GPUs in the future, while still supporting efficient CPU execution when GPU usage is not required.

The workflow files are updated to use the latest docker image which contains the updated Trilinos built.
The dockerfiles are updated also for reference.

## Documentation
Documentation about Trilinos linear algebra should be added.

## Testing
- existing tests using Trilinos are updated. 
- a new test is added for the fluid iliac artery to use Trilinos ML preconditioner


Code/Source/solver/TrilinosLinearAlgebra.cpp
```diff
@@ -140,6 +140,15 @@ void TrilinosLinearAlgebra::initialize(ComMod& com_mod, eqType& lEq)
...
 
+/// @brief Finalize Trilinos.
+void TrilinosLinearAlgebra::finalize()
+{
+  if (Kokkos::is_initialized())
+  {
+    Kokkos::finalize();
+  }
+}
+
```
ktbolt, line 151: This reference to `Kokkos` should be in trilinos_impl.cpp.

ktbolt:
@dcodoni What's the plan here ?

dcodoni:
@ktbolt macos now is failing the ustruct tests even though they were succeeding before. The coverage problem I honestly have no idea yet!

ktbolt:
@dcodoni MacOS tests failed because of an `MPI error` related (maybe) to a network time out. 

Ubuntu is failing because of a `make error` related (I think) to `computing coverage`.

dcodoni:
@ktbolt the ubuntu test fails during generation of code coverage with error: geninfo: ERROR: mismatched end line. I tried then to ignore this type of errors by including in the CmakeLists the following flags: --ignore-errors mismatch --rc geninfo_unexecuted_blocks=1. This made it go further but at the end I got another make error, which I haven't investigated still.

dcodoni:
@mrp089 that is actually a good idea, and I think it may solve the problem I have been having

ktbolt:
@dcodoni I think it is cleaner to exclude the unit tests from the coverage step, the macros in the unit tests are probably causing problems.

dcodoni:
@ktbolt I found some allocation problems in the Trilinos implementation that showed up only at high number of cores and for CFD problems. I am trying to investigate it. Let's not merge it now.

dcodoni:
@ktbolt I fixed the bug in the LHS matrix graph creation. The problem was in the allocation of the non-zero elements per row during graph creation. The local to global map of the matrix used the sorted ordering while the graph was using information in unsorted ordering. This mismatch caused the wrong allocation problem, since Tpetra does not allow for dynamic allocation unlike Epetra.

Code/Source/solver/TrilinosLinearAlgebra.cpp
ktbolt: Looks good !

PR #427, opened 2025-08-27
Adding test cases for reading of fourier coefficient file

### LINKED ISSUE #418: Undefined behavior when reading in Fourier coefficients files - indexing issue
### Description

Fourier coefficient files (.fcs) are an alternative way to provide temporal flow data (.flow) that is prescribed to the inlet of a simulation. .fcs files typically contain the number of Fourier modes, followed by the real and imaginary components of the mode for the interpolated flow field. For example, 

```inlet_flow.fcs 
<first_time_point>    <last_time_point>
<initial_value>    <linear_slope>
<number_of_Fourier_modes>
<real_Fourier_mode_1>    <imag_Fourier_mode_1>
<real_Fourier_mode_2>   <imag_Fourier_mode_2>  
...
<real_Fourier_mode_N>   <imag_Fourier_mode_N>  
```

An example `inlet_flow.fcs` file for n = 8 modes: 
```
0.0 0.43904
-44.6 4.73760
8
1.437551e+00    0.000000e+00
6.633563e-02    -4.090099e+00
-1.793041e+00    7.981296e-01
5.613653e-01    -1.145740e+00
4.747748e-01    2.411945e-01
5.653279e-02    2.779043e-01
-3.934419e-01    2.611582e-01
4.161334e-02    -1.333616e-01 
```

There is a possible issue with the reading of this file inside of the `read_fourier_coeff_values_file` function in `read_files.cpp`. 

```
while (std::getline(temporal_values_file, line)) { 
    line.erase(std::remove(line.begin(), line.end(), '\r'), line.end());
    if (line == "") {
      continue;
    }
    std::istringstream line_input(line);
    std::vector<double> values;
    while (line_input >> value) {
      values.push_back(value);
    }
int num_vals = values.size();
for (int i = 0; i < values.size(); i++) {
      lBc.gt.r(i,j) = values[i]; 
      lBc.gt.i(i,j) = values[i+num_vals];
    }
```

This is because (typically) there is only one column of real components and one column of imaginary components (n = 1 dimension, each row contains r1, i1). The code is designed to account for multi-dimensional modes. (For example in n = 3 dimensions, each row contains r1, r2, r3, i1, i2, i3). From my understanding, `num_vals` accesses the size of the first set of Fourier modes and then applies an offset when the real and imaginary components are being read. 

So, in our 1D example, `num_vals = 2` but because we only have 2 columns inside of the .fcs file, it results in the following memory allocation error: 
Image

### Reproduction

The simulation case provided here should be run multiple times to produce the memory allocation error: [nitros.zip](https://github.com/user-attachments/files/21415223/nitros.zip)

### Expected behavior

The real and imaginary Fourier coefficients should be read in using the correct indices. `num_vals = 1` for 1D temporal data. 

One possible solution is to take the `num_vals` parameter and divide it by 2 and then apply the offset. This allows the code to be generalizable to n dimensions, if necessary. Here is my idea:
```
    int num_vals = values.size();
    // divide the num_vals by 2 
    if (num_vals % 2 != 0) {
      throw std::runtime_error("Fourier coefficients must be provided in real-imaginary pairs.");
    }
    num_vals /= 2;
    for (int i = 0; i < num_vals; i++) {
      lBc.gt.r(i,j) = values[i]; 
      lBc.gt.i(i,j) = values[i+num_vals];
    }
```

Once this is fixed, then we can include this example as a test case. It is equally important to compare this to the temporal values approach (i.e., simulating the same test case using `inlet_inflow.flow` instead of `inlet_flow.fcs` file). 

### Additional context

This feature has existed since the svFSI solver, but this was not tested when the code was translated [#227](https://github.com/SimVascular/svMultiPhysics/issues/227). Looking at the svFSI tests, a case has existed [01-pipe3D_RCR](https://github.com/SimVascular/svFSI-Tests/tree/master/04-fluid/01-pipe3D_RCR) before, but was never tested.

In terms of practicality, the Fourier coefficients is probably not that useful. This is because users will typically provide a `inflow.flow` instead and the `fft` interpolation function is called after the temporal values file is read. However, I am working on a python script that allows users to select the number of Fourier modes for a temporal file and visualize the interpolation. For example, the plot below shows scatter plots of the temporal data that is provided in the `.flow` file, and the continuous line represents the interpolation. (Adjusting the number of modes will either underfit or overfit the flow profile.)

Image

### Code of Conduct

kko27:
Adds a test case that reads in Fourier coefficients for boundary condition and fixes the read io bug. 

## Current situation
Resolves #418 (indexing issue when reading in Fourier coefficient files) 

## Release Notes 
- Added pipe_RCR_3d_fourier_coeff test case
- Test case includes python script fft_temporal_values.py that can convert the .flow file into a .fcs (fourier coefficients) file
- Added unit test (test_fft) that tests the fft function
- Fixed the indexing issue when reading in .fcs files

## Documentation
 
Code is documented with comments. Test case includes an updated README. 

## Testing
Test case is included as pipe_RCR_3d_fourier_coeff. Unit test has also been included. 
 


tests/cases/fluid/pipe_RCR_3d_fourier_coeff/fft_temporal_values.py
```diff
@@ -0,0 +1,187 @@
...
+    print("Slopes (qs):", result['qs'])
+    print("Initial time (ti):", result['ti'])
+    print("Total duration (T):", result['T'])
+
+    # After computing `result` with your fft function:
+    output_filename = os.path.join('./lumen_inlet.fcs')
+    write_fourier_coeff_file(output_filename, result, d, nf)
+
+    print(f"Fourier coefficient file written to {output_filename}")
+    visualize_fft(result, temporal_values, nf)
```
ktbolt, line 187: @kko27 Yet another Python script added to the test suite. Does it matter how the .fcs file was created ? 

Maybe we should add a `utilities` directory to store Python scripts.

tests/unitTests/math_tests/test_fft.cpp
```diff
@@ -0,0 +1,95 @@
...
+        double step = (x_end - x_start) / (N - 1);
+        for (int i = 0; i < N; ++i) {
+            double t = x_start + i * step;
+            double y = std::sin(t) + std::cos(t) + 0.1 * t;
+            temporal_values.push_back({t, y});
+        }
+    }
+
+    // Helper method to initialize Fourier coefficients data structure
+    void InitializeFourierCoefficients(fcType& gt, int d = 1, int n = 16) {
```
ktbolt, line 57: An obvious comment.

tests/cases/fluid/pipe_RCR_3d_fourier_coeff/README.md
```diff
@@ -0,0 +1,55 @@
...
+```
+<Line 1>  <first_time_point>    <last_time_point>
+<Line 2>    <initial_value>     <linear_slope>
+<Line 3>  <number_of_Fourier_modes>
+<Line 4>  <real_Fourier_mode_1>    <imag_Fourier_mode_1>
+<Line 5>  <real_Fourier_mode_2>   <imag_Fourier_mode_2>  
+...
+<Line N+3>  <real_Fourier_mode_N>   <imag_Fourier_mode_N>  
+```
+
```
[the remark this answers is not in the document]
aabrown100-git: Yes, I tend to agree. I think you can remove the xml code examples, and just link to both the solver.xml file, as well as [this documentation](https://simvascular.github.io/documentation/multi_physics.html). @kko27 could you also double check that the information in the documentation regarding Fourier coefficients and their file format is correct?

tests/cases/fluid/pipe_RCR_3d_fourier_coeff/fft_temporal_values.py
aabrown100-git: I think this Python script is useful to include. It makes the test case more reproducible (somebody can double check that the .fcs indeed matches the .flow files). Also, a user could reuse this script to prescribe Fourier coefficients instead of flow data in their own simulations. This would help them check that the Fourier coefficient interpolation reproduces the flow data well. I agree about the `utilities` directory though.

tests/cases/fluid/pipe_RCR_3d_fourier_coeff/result_002.vtu
aabrown100-git, line 1: Is this identical to the result file for pipe_RCR_3d? Or did you regenerate it for this Fourier coefficient test case?

tests/unitTests/math_tests/test_fft.cpp
```diff
@@ -0,0 +1,95 @@
...
+ * OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+ * EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+ * PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+ * PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+ * LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+ * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+ #include "fft.h"
```
aabrown100-git, line 31: Extra space here

tests/cases/fluid/pipe_RCR_3d_fourier_coeff/solver.xml
```diff
@@ -0,0 +1,119 @@
...
+      <NS_GM_tolerance> 1e-3 </NS_GM_tolerance>
+      <NS_CG_tolerance> 1e-3 </NS_CG_tolerance>
+      <Absolute_tolerance> 1e-17 </Absolute_tolerance>
+      <Krylov_space_dimension> 250 </Krylov_space_dimension>
+   </LS>
+
+   <Add_BC name="lumen_inlet" > 
+      <Type> Dir </Type> 
+      <Time_dependence> Unsteady </Time_dependence> 
+     <Fourier_coefficients_file_path> lumen_inlet.fcs </Fourier_coefficients_file_path>
```
aabrown100-git, line 92: Can you add a space in this line so `<Fourier_coefficients_file_path>` is aligned with the other lines?

tests/cases/fluid/pipe_RCR_3d_fourier_coeff/lumen_inlet.fcs
aabrown100-git, line 1: The very small values in this file (e-16, e-17, etc..) are slightly different from the existing `lumen_inlet.fcs` in `pipe_RCR_3d` (which I guess is not even used). Do you know why they are slightly different? 

Also, the format for the first four values is different. Do you know why?

Also, since that file isn't actually used in `pipe_RCR_3d`, maybe we can remove it to avoid confusion.

tests/unitTests/math_tests/test_fft.cpp
```diff
@@ -0,0 +1,95 @@
...
+    double x_start = 0.0;   // start time
+    double x_end = 10.0;    // end time 
+    std::vector<std::vector<double>> temporal_values;
+    
+    // Create the temporal values using helper method
+    CreateTemporalValues(N, x_start, x_end, temporal_values);
+    
+    // Initialize the Fourier coefficients data using helper method
+    fcType gt;
+    InitializeFourierCoefficients(gt, 1, 16);
```
aabrown100-git, line 80: Why 16 here? What does it mean?

tests/unitTests/math_tests/test_fft.cpp
```diff
@@ -0,0 +1,95 @@
...
+    
+    // Compute the Fourier coefficients
+    fft(N, temporal_values, gt);
+
+    // Check the slope (first Fourier coefficient)
+    ASSERT_NEAR(gt.qs[0], -0.13830, 1e-2) << "Expected slope ~-0.13830";
```
aabrown100-git, line 86: How did you calculate these expected values?

aabrown100-git (changes requested):
Nice job with this! Just added a few suggestions and questions for clarification.

tests/cases/fluid/pipe_RCR_3d_fourier_coeff/fft_temporal_values.py
kko27: @ktbolt and @aabrown100-git, yes I think that makes sense. Where in the codebase should we add the `utilities` directory?

aabrown100-git: Top level? (parallel to `tests/`). I guess this could also contain various pre and post processing scripts as well?

tests/unitTests/math_tests/test_fft.cpp
kko27: @aabrown100-git, this refers to 16 Fourier modes. I will add that as a comment

tests/unitTests/math_tests/test_fft.cpp
kko27: @aabrown100-git I think I used my python script, or some version of it

tests/cases/fluid/pipe_RCR_3d_fourier_coeff/lumen_inlet.fcs
aabrown100-git: Thanks for the detailed response!

aabrown100-git: I think you should remove the Fourier coefficients file in the other case since it's not used

tests/cases/fluid/pipe_RCR_3d_fourier_coeff/result_002.vtu
aabrown100-git: Hm yes could you double check this? It would be great if the same results file works for both cases, since the two cases should produce the same results

ktbolt:
@aabrown100-git Having a script in a test directory means that it does something else than just testing the solver. 

The Python script is indeed useful so let's put it someplace users can find it ! I can add documentation for a `utilities` directory.

aabrown100-git (approved):
Looks great, love the extensive documentation!

PR #429, opened 2025-09-09
Remove redundant brew install cmake in MacOS workflow file

aabrown100-git:
Removed redundant installation of cmake from MacOS dependencies.

## Current situation
MacOS tests were failing because of 

```
Error: cmake was installed from the local/pinned tap
but you are trying to install it from the homebrew/core tap.
Formulae with the same name from different taps cannot be installed at the same time.

To install this version, you must first uninstall the existing formula:
  brew uninstall cmake
Then you can install the desired version:
  brew install cmake
```

## Release Notes 
- Remove `cmake` from list of packages installed by `brew`

## Testing
- No testing, will see if macos tests pass with this change.

### DIFF
--- a/.github/workflows/test_macos.yml
+++ b/.github/workflows/test_macos.yml
@@ -18,7 +18,7 @@ jobs:
       - name: Install MacOS dependencies
         run: |
           brew reinstall -v gcc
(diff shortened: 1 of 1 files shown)


ktbolt (approved):
Looks good.

PR #416, opened 2025-07-20
Update REAME with correct command for using local VTK install

menon-karthik:
## Current situation
Closes #415 

## Release Notes 
Updated the README

## Documentation
This is the documentation. 

## Testing
N/A

## Code of Conduct & Contributing Guidelines 
- [x] I agree to follow the [Code of Conduct](https://github.com/SimVascular/.github/blob/main/CODE_OF_CONDUCT.md) and [Contributing Guidelines](https://github.com/SimVascular/.github/blob/main/CONTRIBUTING.md).

PR #430, opened 2025-09-11
Fix conditional GitHub testing

aabrown100-git:
## Current situation
Currently, [this PR](https://github.com/SimVascular/svMultiPhysics/pull/416) is stuck because it only changes a `README.md` file, and the MacOS and Ubuntu tests do not run if only a `README.md` is changed. This is intended, but GitHub needs a success status for a test action before we can merge the PR. Since no tests are ever run, the test action is perpetually pending, preventing merging.

## Release Notes 
- Reorganizes GitHub testing to deal with this issue properly
- New `tests.yml` is run first. This contains a job `check-for-changes`, which checks if files in `Code/`, `tests/`, or `.github/` were changed
- If there were file changes in these folders, the original integration tests, now refactored as Actions under `.github/actions/test_macos/action.yml` and `.github/actions/test_ubuntu/action.yml` are run. If these pass, then a coordinating job `all-tests` (named "All Tests" in the GitHub UI) returns with a success.
-  I will modified the [branch protection rules](https://github.com/SimVascular/svMultiPhysics/settings/branch_protection_rules/36012339) to require this "All Tests" instead of the previous "test".
- If important files were not changed,  "All Tests" still runs and returns successfully. This will satisfy the branch protection rules requirement.

Note: For an open PR, if you make a new commit to the feature branch associated with the PR, and that commit doesn't modify important files ( i.e. in `Code/`, `tests/`, or `.github/`), the integration tests **will not** run on your feature branch, but they **will** run within the PR. This is a bit annoying, but I think it's safer this way. Otherwise, someone could make a documentation-only commit to the PR and pass all tests, even if the integration tests failed in previous commits.

## Testing
This is testing. We may want to make a test pull request to check that this is functioning correctly.


aabrown100-git:
@ktbolt @mrp089 any thoughts on this?

ktbolt:
@aabrown100-git I really don't have any experience in all of the yaml coding. But I would think that the tests would work the same way as they do on your fork, must be a difference in the files, maybe something not being merged the same way.

I do not see how the yaml code distinguishes between README and other files.

Issue #432, opened 2025-09-17
Fiber_direction spatial output causes seg fault


aabrown100-git:
### Description

There is some code to output fiber directions in the `result.vtu` files. This can be done by using the `Fiber_direction` keyword in the input file
`<Output type="Spatial" >
     <Fiber_direction> true </Fiber_direction>
   </Output>`

However, this causes a segfault when svMultiPhysics attempts to write a `result.vtu` file.

The fiber direction should be identical to that provided by the user in the first place. So this feature is not that important. But it would still be nice to have to verify that the fiber directions provided by the user are correctly read and processed by svMultiPhysics.

This has never been tested, and I think there is probably a translation bug in [`fib_dir_post()`](https://github.com/SimVascular/svMultiPhysics/blob/6ae45722e0579c75eb6878eefbaa3fe56fa72787/Code/Source/solver/post.cpp#L643)

### Reproduction

Add `<Fiber_direction> true </Fiber_direction>` as a spatial output in the [struct/LV_HolzapfelOgden_active solver.xml](https://github.com/SimVascular/svMultiPhysics/blob/main/tests/cases/struct/LV_HolzapfelOgden_active/solver.xml).

On my MacBook pro, this produces a segfault error

```
(base) aaronbrown@Aarons-MacBook-Pro LV_HolzapfelOgden_active % mpirun -np 1 /Users/aaronbrown/GitHub/svMultiPhysics_aabrown100-git/build/svMultiPhysics-build/bin/svmultiphysics solver.xml
[Array] **********************************
[Array] WARNING: Index checking is enabled 
[Array] **********************************
[Array] **********************************
[Array] WARNING: Index checking is enabled 
[Array] **********************************
[Vector] **********************************
[Vector] WARNING: Index checking is enabled 
[Vector] **********************************
[Array3] **********************************
[Array3] WARNING: Index checking is enabled 
[Array3] **********************************
[Vector] **********************************
[Vector] WARNING: Index checking is enabled 
[Vector] **********************************
[Vector] **********************************
[Vector] WARNING: Index checking is enabled 
[Vector] **********************************
[Vector] **********************************
[Vector] WARNING: Index checking is enabled 
[Vector] **********************************
---------------------------------------------------------------------
 Eq     N-i     T       dB  Ri/R1   Ri/R0    R/Ri     lsIt   dB  %t
---------------------------------------------------------------------
[...]

PR #435, opened 2025-09-18
Fiber direction spatial output seg fault 432

ktbolt:
This functionality was never tested. I have found and fixed several bugs (see https://github.com/SimVascular/svMultiPhysics/issues/432).


aabrown100-git:
Looks good!

PR #439, opened 2025-09-25
RIS valve mesh scale factor bug fix

hanzhao2020:
Fix bugs in the RIS valve code when the mesh scale factor is not one

## Current situation
The current RIS code always uses the unit cm when using the signed distance function to determine if the valve is opening or closing. These parameters should be consistent with the mesh scale factor.

## Release Notes 
All signed distance function related parameters are using consistent units now.

## Testing
This update is tested by the URIS tests.


Code/Source/solver/ComMod.h
```diff
@@ -1485,13 +1485,13 @@ class urisType
     Array<double> Yd;
 
     // Default signed distance value away from the valve.
-    double sdf_default = 10.0;
+    double sdf_default = 10.0; // in cm
```
ktbolt, line 1488: @hanzhao2020 What does `in cm` mean ?

hanzhao2020: It means the default value is in the unit of centimeters.

hanzhao2020:
@ktbolt Thanks for the reminder. An issue has been created and linked to this pull request.

aabrown100-git (commented):
@hanzhao2020 I tested this with my left heart FSI case with `mesh_scale_factor != 1`, and it looks like the SDF fields are being computed correctly. The nonlinear convergence is pretty bad though. I also noticed that convergence is bad for the `pipe_uris_3d/` test case as well. Do you know why?

hanzhao2020:
@aabrown100-git I think it's because the added resistive forces are similar to penalty terms, affecting the conditioning of the linear system. Using smaller resistance values can improve the nonlinear convergence a bit, but the convergence is still pretty slow.

aabrown100-git (approved):
Looks good to me now, thanks.

Issue #423, opened 2025-08-15
NS solver fails with divide by zero in gmres


ktbolt:
### Description

A user is trying to run the solver with incorrectly specified boundary conditions 
```
        <Add_BC name="inlet_cylinder">
            <Type>Neumann</Type>
            <Time_dependence>Resistance</Time_dependence>
            <Value>0</Value>
        </Add_BC>

        <Add_BC name="outlet_cylinder">
            <Type>Neumann</Type>
            <Time_dependence>Resistance</Time_dependence>
            <Value>41241</Value>
        </Add_BC>
```
created an ill-defined problem. 

The `gmres()` function fails with a dived by zero exception here https://github.com/SimVascular/svMultiPhysics/blob/3629164c21e55f945faa014108d5dc54581a3c8e/Code/Source/liner_solver/gmres.cpp#L169 because err(0) is 0.

### Reproduction

Run the `svMultiPhysics/tests/cases/cmm/pipe_3d` problem with the above BCs.

### Expected behavior

Ill-defined problems need to be recognized and an appropriate message displayed to help the user understand the problem.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct and Contributing Guidelines

PR #453, opened 2025-10-15
Fix divide by zero in gmres 423

ktbolt:
Ill-posed boundary conditions can cause a divide by zero fault https://github.com/SimVascular/svMultiPhysics/issues/423.

I've added a check for this that throws an exception if a zero norm is computed.


dcodoni (commented):
@ktbolt I have some doubts about this problem. 

I am skeptical about the fact that changing boundary conditions we get exactly zero as a result from some function in gmres. It makes me think that some variables enter the function zero-initialized. Also in the commit bf76a32, I commented a portion that would have prevented this problem. 

Did you try to set the same incorrect BCs to any other case in fluid dynamics and replicate this problem? Also does this problem relates to gmres or bipn?

ktbolt:
@dcodoni I am not as familiar as you are with what is going on in  `gmres`.

But as stated in the Issue setting an inlet boundary condition for svMultiPhysics/tests/cases/fluid/pipe_RCR_3d like so
```
<Type>Neumann</Type>
<Time_dependence>Resistance</Time_dependence>
<Value>0</Value>
```
causes the `R` array passed to `gmres` to be all zeros which causes a divide by zero in the first iteration.

Issue #387, opened 2025-05-23
Add reading in an XML file within the solver XML file


ktbolt:
### Problem

It would be useful to be able to read an XML file containing solver parameters from within the solver XML file. This would be useful to define a base simulation using different sets of parameters. 

For example, the linear solver for a simulation could be set by reading in a file containing `LS` parameters.

### Solution

I will add a `Include_xml` parameter used to read in an XML file 
```
   <Include_xml> linear.xml </Include_xml>
```
and modify certain parameters like `Add_equation` to process the parameters read from an XML file.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct and Contributing Guidelines

ktbolt:
I have added a `Parameter<std::string> include_xml` object to `EquationParameters` that stores data for an `Include_xml` parameter under `Add_equation`
```
<Add_equation type="fluid" >
   <Coupled> 1 </Coupled>
   <Min_iterations> 3 </Min_iterations>
   <Max_iterations> 10 </Max_iterations>
   <Tolerance> 1e-3 </Tolerance>
   <Backflow_stabilization_coefficient> 0.2 </Backflow_stabilization_coefficient>

   <Include_xml> linear.xml </Include_xml>

.
.
.
</Add_equation>
```

The `linear.xml` contains `LS` parametes
```
<?xml version="1.0" encoding="UTF-8" ?>
<svMultiPhysicsFile version="0.1">

   <LS type="NS" >
      <Linear_algebra type="fsils" >
         <Preconditioner> fsils </Preconditioner>
      </Linear_algebra>
      <Max_iterations> 10 </Max_iterations>
      <NS_GM_max_iterations> 3 </NS_GM_max_iterations>
      <NS_CG_max_iterations> 500 </NS_CG_max_iterations>
      <Tolerance> 1e-3 </Tolerance>
      <NS_GM_tolerance> 1e-3 </NS_GM_tolerance>
      <NS_CG_tolerance> 1e-3 </NS_CG_tolerance>
      <Krylov_space_dimension> 50 </Krylov_space_dimension>
   </LS>

</svMultiPhysicsFile>
```

I then added the `EquationParameters::process_include_xml()` method that reads in an XML file and adds parameters to the `EquationParameters` object. 

This all works just like the `LS` parameter was defined in the solver XML file and it will work for any parameters in the `Add_equation` section (e.g. `Add_BC`).

ktbolt:
[...]

PR #460, opened 2025-10-27
Add reading in an XML file within the solver XML file

ktbolt:
I've added reading in an XML file within the solver XML file (https://github.com/SimVascular/svMultiPhysics/issues/387).

This required minimal code changes since each parameter class calls its `set_values` method recursively with the root element create from the external XML file .


dcodoni (commented):
just one comment about the .sh file

dcodoni (approved):

Issue #470, opened 2025-11-10
segmentation fault when using FSILS assembly type in Trilinos implementation


dcodoni:
### Description

The FSILS type of assembly when using Trilinos library for linear algebra throw a segmentation fault. This is caused by a variable 'nnzPerRow' which was not defined before. The variable 'nnzPerRow' contains how many non-zero elements there are per row and it is used to assemble the global matrix in Trilinos data structure.

### Reproduction

Just set the following:
<Linear_algebra type="trilinos" >
          <Assembly> fsils </Assembly>
          ...
</Linear_algebra>
In the input file. A segmentation fault for invalid memory location will be thrown by the solver.

### Expected behavior

The expected result is to use either assembly type ('trilinos' or 'fsils') when selecting Trilinos library for the linear algebra.

### Additional context

Even though the default and suggested assembly type is 'trilinos', the 'fsils' type is the only option available whenever 'ustruct' type of equation is selected.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct and Contributing Guidelines

PR #471, opened 2025-11-10
Invalid Memory Access for 'fsils' Assembly Type in Trilinos Linear Algebra (addresses #470)

dcodoni:
Bug Fix: Invalid Memory Access for 'fsils' Assembly Type in Trilinos Linear Algebra

## Current situation
A segmentation fault occurred when selecting 'fsils' as the global assembly method in the Trilinos linear algebra backend. This issue made it impossible to use 'fsils' for problems requiring it, particularly when working with the 'ustruct' equation type, where 'fsils' is the only valid assembly option. The segmentation fault was caused by an undefined variable nnzPerRow during the matrix assembly process in the Trilinos implementation. This variable is necessary for correctly allocating and populating matrix entries, and its absence led to invalid memory accesses.

## Release Notes 
The variable nnzPerRow is now properly defined and initialized within the Trilinos implementation. This ensures that the matrix assembly process proceeds correctly when 'fsils' is used as the global assembly method.

## Testing
A new integration test has been added to verify the fix:
- Test name: /ustruct/LV_HolzapfelOgdenModifiedAnisotropy_passive_trilinos
- Purpose: Ensures that 'fsils' assembly type works correctly with Trilinos linear algebra.

### DIFF
--- a/Code/Source/liner_solver/ls.cpp
+++ b/Code/Source/liner_solver/ls.cpp
@@ -34,18 +34,20 @@ void fsils_ls_create(FSILS_lsType& ls, LinearSolverType LS_type, double relTol, 
 
     case LinearSolverType::LS_TYPE_GMRES:
       ls.RI.relTol = 0.1;
-      ls.RI.mItr   = 4;
+      ls.RI.mItr   = 1000;
       ls.RI.sD     = 250;
     break;
 
     case LinearSolverType::LS_TYPE_CG:
       ls.RI.relTol = 1.E-2;
       ls.RI.mItr   = 1000;
+      ls.RI.sD     = 250;
     break;
 
     case LinearSolverType::LS_TYPE_BICGS:
       ls.RI.relTol = 1.E-2;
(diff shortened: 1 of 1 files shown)


Code/Source/liner_solver/ls.cpp
```diff
@@ -34,18 +34,20 @@ void fsils_ls_create(FSILS_lsType& ls, LinearSolverType LS_type, double relTol,
...
       ls.RI.relTol = 1.E-2;
       ls.RI.mItr   = 1000;
+      ls.RI.sD     = 250;
     break;
 
     case LinearSolverType::LS_TYPE_BICGS:
       ls.RI.relTol = 1.E-2;
-      ls.RI.mItr   = 500;
+      ls.RI.mItr   = 1000;
+      ls.RI.sD     = 250;
```
ktbolt: @dcodoni Note that setting `ls.RI.mItr` here overrides the default value set in `LinearSolverParameters`.  I guess that is ok.

PR #502, opened 2026-02-09
Initial conditions for electrophysiology TTP ionic model #407

### LINKED ISSUE #407: Hard coded initial conditions for EP simulations (TTP ionic model)
### Problem

The ten-Tusscher-Panfilov (TTP) ionic model uses 18 variables which all require initialization in our simulations. However, in our implementation, we hard code these state variables (shown in the screenshot below). This can be found in the CepModTtp.cpp file. 

Image

I think there is a need to give users the option to specify different sets of initial state variables. This is because one common method to observe re-entry in electrophysiology simulations is to assign different sets of initial states to different parts of the domain. Here is an example that I've been working on: 

Image

In panel A, we assign different initial conditions to each of the 4 domains. Each domain is at a different phase of the action potential. Panel B show the simulation results which show the formation of a spiral wave. I was able to create this simulation in svMultiphysics by using a 'hack' solution; I added more hard-coded initial state variables to the initialization function. The big disadvantage to this is that every time I want to choose a different set of initial conditions, I would need to recompile the code. 

### Solution

One solution to this is to pull out all of these initial state variables and put them in input files that the user can modify. This has been done in other open source electrophysiology packages, such as openCARP [openCARP example](https://git.opencarp.org/openCARP/experiments/-/tree/master/tutorials/02_EP_tissue/03E_study_resolution). 

In our solver, we have (by default) 3 sets of initial states that have been hard-coded, each corresponding to different type of cardiac tissue: epicardium, endocardium, and myocardium. In my case that I want to run, I need 4 sets of initial states. I'm not too sure what the best approach would be in creating an input file. (For example, would it be a single file that contains multiple sets of initial conditions?) Open to hearing other suggestions. 

### Additional context

_No response_

### Code of Conduct

kko27:
Resolves #407


ktbolt:
@kko27 This PR is eight months old, that will not do !

kko27:
> @kko27 This PR is eight months old, that will not do !

@ktbolt, sorry about that. What should I do in this situation? Should I create a new branch in my fork and then include these changes

Issue #448, opened 2025-09-30
Slow nonlinear convergence when using URIS valve in fluid equation


aabrown100-git:
### Description

Nonlinear convergence in the [`uris` valve test case](https://github.com/SimVascular/svMultiPhysics/tree/main/tests/cases/uris/pipe_uris_3d) is very bad. Here is the convergence history for the 5th timestep when running on 4 processors

```
FS 5-1  1.049e+02  [0 1.000e+00 2.828e+01 9.680e-05]  [175 -2 4]
 MS 5-1  1.054e+02  [0 1.000e+00 7.349e-02 8.745e-05]  [48 -2 3]
 FS 5-2  1.067e+02  [-19 1.077e-01 3.045e+00 9.532e-05]  [94 -9 9]
 MS 5-2  1.072e+02  [-81 8.745e-05 6.426e-06 9.295e-05]  [53 -2 4]
 FS 5-3  1.084e+02  [-28 3.918e-02 1.108e+00 9.985e-05]  [44 -92 4]
 MS 5-3  1.089e+02  [-161 8.128e-09 5.973e-10 6.585e-03]  [28 0 2]
 FS 5-4  1.100e+02  [-30 2.888e-02 8.168e-01 9.879e-05]  [34 -92 3]
 MS 5-4  1.106e+02  [-205 5.353e-11 3.934e-12 1.000e+00]  [0 0 0]
 FS 5-5  1.118e+02  [-32 2.469e-02 6.985e-01 9.812e-05]  [34 -92 3]
 MS 5-5  1.123e+02  [-205 5.353e-11 3.934e-12 1.000e+00]  [0 0 0]
 FS 5-6  1.137e+02  [-33 2.180e-02 6.165e-01 9.941e-05]  [49 -92 5]
 MS 5-6  1.142e+02  [-205 5.353e-11 3.934e-12 1.000e+00]  [0 0 0]
 FS 5-7  1.155e+02  [-34 1.949e-02 5.512e-01 9.813e-05]  [88 -6 9]
 MS 5-7  1.160e+02  [-205 5.353e-11 3.934e-12 1.000e+00]  [0 0 0]
 FS 5-8  1.174e+02  [-35 1.756e-02 4.967e-01 9.833e-05]  [82 -4 7]
 MS 5-8  1.178e+02  [-205 5.353e-11 3.934e-12 1.000e+00]  [0 0 0]
[...]

PR #505, opened 2026-02-11
Updates for URIS valve with nonzero valve velocity and improved nonlinear convergence

hanzhao2020:
## Current situation
* The URIS code has stagnated nonlinear convergence for FSI problems [#448](https://github.com/SimVascular/svMultiPhysics/issues/448).
* The current implementation only considers zero valve velocity.
* The scaffold surface in the aortic valve model is not considered.

## Release Notes 
* Resolves [#448](https://github.com/SimVascular/svMultiPhysics/issues/448) by implementing explicit geometric coupling for FSI problems. At each time step, the fluid-structure equation is first solved until convergence, and then the mesh motion equation is solved.
* The valve velocity is computed from the input mesh motion and added to the resistive penalty term.
* The code is updated to consider the fixed aortic valve scaffold surface.

## Testing
* The new code is covered by tests.

### DIFF
--- a/Code/Source/solver/uris.cpp
+++ b/Code/Source/solver/uris.cpp
@@ -15,7 +15,6 @@
 #include "read_msh.h"
 #include "VtkData.h"
 
-
 namespace uris { 
 
 /// @brief This subroutine computes the mean pressure and flux on the 
@@ -61,7 +60,6 @@ void uris_meanp(ComMod& com_mod, CmMod& cm_mod, const int iUris) {
   for (int iM = 0; iM < com_mod.nMsh; iM++) {
     volU += all_fun::integ(com_mod, cm_mod, iM, sUPS);
   }
-
 
   // Let's compute right side
   Array<double> sDST(1,com_mod.tnNo);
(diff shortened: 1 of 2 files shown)


ktbolt:
@hanzhao2020 The  GitHub Issue https://github.com/SimVascular/svMultiPhysics/issues/448 has not been updated in three months and now all of a sudden there is a PR with changes to like 14 files. 

How can other Team members understand the details of what the problem was and the fix ? Merging the PR and closing the Issue makes it look like the problem was never solved. I've seen developers spend two weeks trying to understand what changes were made to the code and never really understanding why the changes were made.

hanzhao2020:
@ktbolt Thanks for this concern. The issue was not updated while I was investigating it. If it helps, I can break this PR into multiple PRs to make them easier to understand, and each PR addresses a separate issue. Or any other suggestions are appreciated.

ktbolt:
@hanzhao2020  Issues are important because they provide a history of what code was changed and why. Each Issue has a link to the PR which provides the details of code changes but a high-level description provides a context in which to understand those changes. 

For example: why was quadratic transition between RIS valve open and close thickness needed ? 

The reason we do all of this is to be able to maintain and extend the code after people leave. If a year from now there are problems with the URIS code then we need to understand how to fix it.

ktbolt:
@hanzhao2020 Sounds good, thank you !

Issue #363, opened 2025-04-29
Read mesh domains from a vtu/vtp file


msbazzi:
### Problem

The ability to set DOMAIN_ID as a cell array in VTU/VTP files was available in the FORTRAN implementation but has not been translated into the current c++ version. This functionality is necessary for workflows that involve multiple domains within a single mesh, such as in TEVG modeling, where domain-specific properties are required.

### Solution

Translate the SUBROUTINE SETDMNIDVTK(lM, fName, kwrd) from FORTRAN to c++ (we have the function call in there already, void set_dmn_id_vtk(Simulation* simulation, mshType& mesh, const std::string& file_name, const std::string& kwrd)

### Additional context

I stared the implementation in my branch -> Read_domains

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct and Contributing Guidelines

ktbolt:
@msbazzi Using VTK files for domain IDs was never tested in svFSI which is why I did not implement it in the C++ solver.

Is there a good use case for having two formats to specify domain IDs ? The text format seems to be ok although that format could be updated to make it more clear and less error prone to use, at the very least add an element ID but better to add a element range.

msbazzi:
@mrp089 @ktbolt I implemented the void set_dmn_id_vtk that was missing to read vtu data type, and for that I had to write a function that reads cell data, we only had to point data. (try to make a cleaner version in my branch https://github.com/msbazzi/svMultiPhysics/tree/Domains_vtu# ). 
However it is still not working properly. The solution obtained when reading a vtu file is completely off when compared to .dat file.[Test_domains.zip](https://github.com/user-attachments/files/20494445/Test_domains.zip) 
My initial assessment showed that there was some mismatches associated to the btest (int64 vs int32), but it seems to be reading correctly now, so I dont know what is causing the issue.

ktbolt:
@msbazzi I know that you are not using `mesh/lumen/domainIDs.vtu`, my main point was that the domain IDs text and VTU files for the solid don't contain the same information.

msbazzi:
@ktbolt I would say:
(1) consistency, we save other mesh properties in the vtu such as fiber direction, prestress, GlobaElementId, etc ... 
(2) reduce the number of file, I have a separate domainIDs.vtu, but my next step would be just add it to the mesh-complete.mesh.vtu, which I think it makes more sense.
(3) If it is in vtu file format, it can be visualized in Paraview before the simulations.

[...]

PR #513, opened 2026-03-03
Read mesh domains from vtk file 363

ktbolt:
Finish https://github.com/SimVascular/svMultiPhysics/issues/363. 

I also cleaned up the interface for VtkData.h.

I don't think there are any tests for this.


javijv4:
@ktbolt, I was looking into this code to reuse some of these functions for the 3D-0D cap implementation. However, I noticed that the functions `VtkVtpData::has_cell_data` and `VtkVtuData::has_cell_data` (lines 801 and 1060 in `VtkData.cpp`) define the number of arrays using `GetPointData()` instead of `GetCellData()`. For example:
```
int num_arrays = impl->vtk_polydata->GetPointData()->GetNumberOfArrays();
```
when it should be,
```
int num_arrays = impl->vtk_polydata->GetCellData()->GetNumberOfArrays();
```
I am not sure if I should open a new issue or something else.

ktbolt:
@javijv4 Good catch ! 

This is a bug; open an Issue for it.

PR #521, opened 2026-03-27
Fix bug when reading domain IDs from .vtus

### LINKED ISSUE #520: VtkData has_cell_data function uses wrong function to get the number of arrays
### Description

The function `VtkVtpData::has_cell_data` and `VtkVtuData::has_cell_data` (lines 801 and 1060 in `VtkData.cpp`)  define the number of arrays as:
```
int num_arrays = impl->vtk_ugrid->GetPointData()->GetNumberOfArrays();
```
when it should be
```
int num_arrays = impl->vtk_polydata->GetCellData()->GetNumberOfArrays();
```

### Reproduction

There is currently no test that checks the implementation. These functions are called when a domain is specified using a `.vtu` or `.vtp` file.

### Expected behavior

Change the code in `VtkData.cpp` to use `GetCellData()`. 
Should we add a test? Maybe the one @msbazzi sent in #363. I can modify the one I added in #504 to use .vtu. That way, we would have a solid and EP test of this.

### Additional context

_No response_

### Code of Conduct

javijv4:
## Current situation
Resolves #520 

## Release Notes 
* Fixed bug
* Added an FSI and an CEP test that checks the functionality to read domain IDs from a `.vtu` file.

## Testing
New tests are:
* `tests/cases/fsi/pipe_3d_domains`
* `tests/cases/cep/slab_domains`


ktbolt (approved):
Looks good !

ktbolt:
@javijv4 Testing reading domain IDs from a file should really be a unit test; does not make sense to have two `pipe_3d` tests that differ only by how they read domain IDs. 

Adding a unit test is probably too much to do for this right now so just remove the `pipe_3d_domains` test.

ktbolt:
@javijv4 There are two solver XML files in svMultiPhysics/tests/cases/cep/slab_domains. Can we run both of them ?

javijv4:
@ktbolt, yep. I removed the pipe test and updated the cep slab_domains test to run both .xml.

kko27 (approved):
Looks good to me!

PR #531, opened 2026-04-18
Refactor ZeroD boundary coupling / adding cap functionality for struct.

### LINKED ISSUE #73: Couple struct/ustruct to LPN (for cardiac mechanics)
Recreating the issue found here https://github.com/SimVascular/svFSI/issues/72. 

The implementation of this feature for svFSI is complete and waiting for approval before merging. I will start translating this feature into C++ code for svFSIplus and perform the work here: https://github.com/aabrown100-git/svFSIplus/tree/0D_coupling.

javijv4:
## Current situation
Resolves #73 

## Release Notes 
* New CoupledBoundaryCondition class to handle all the operations related to a coupled BC between a 3D solver and svZeroDSolver.
* A cap can be added to correctly estimate flowrates in struct / ustruct simulations.


aabrown100-git:
@javijv4 Looks like there's a Mac compiler error. Could you resolve it?

javijv4:
@aabrown100-git, not yet. I'm going to try fixing it now, and if it gets too complicated, I'll ask @ktbolt for help. I'll go through the Copilot reviews as well. Thanks!

javijv4:
Ok. Fixed the Mac compilation issues and the Copilot comments. @ktbolt, @aabrown100-git should be good for review.

Code/Source/solver/CoupledBoundaryCondition.cpp
```diff
@@ -409,8 +427,17 @@ void CoupledBoundaryCondition::distribute(const ComMod& com_mod, const CmMod& cm
     // Distribute cap flag so all ranks agree (master loaded cap_face_; slaves have has_cap_ set here)
     cm.bcast(cm_mod, &has_cap_);
 
-    // Distribute cap surface node count (same on all ranks; owning rank set it at load).
-    cm.bcast(cm_mod, &cap_n_no_);
+    // Distribute cap surface global node IDs (same count on every rank; master filled at load).
+    int cap_nn = static_cast<int>(cap_mesh_global_node_ids_.size());
+    cm.bcast(cm_mod, &cap_nn);
```
ktbolt: @javijv4 No need to use `static_cast<int>` here.

Code/Source/solver/CoupledBoundaryCondition.cpp
```diff
@@ -425,6 +452,8 @@ void CoupledBoundaryCondition::load_cap_face_vtp(const std::string& vtp_file_pat
     has_cap_ = false;
     owns_cap_ = false;
     cap_n_no_ = 0;
+    cap_mesh_global_node_ids_.resize(0);
```
ktbolt: @javijv4 Why `resize(0)` ?

Code/Source/solver/CoupledBoundaryCondition.cpp
```diff
@@ -469,15 +513,15 @@ void CoupledBoundaryCondition::initialize_cap(ComMod& com_mod)
...
+    // Reset the cap global mesh state buffers.
     const int nsd = com_mod.nsd;
-    const int gtnNo = com_mod.gtnNo;
-    cap_global_mesh_state_.gtnNo = gtnNo;
-    cap_global_mesh_state_.x.resize(nsd, gtnNo);
-    cap_global_mesh_state_.Do.resize(nsd, gtnNo);
-    cap_global_mesh_state_.Dn.resize(nsd, gtnNo);
-    cap_global_mesh_state_.Yo.resize(nsd, gtnNo);
-    cap_global_mesh_state_.Yn.resize(nsd, gtnNo);
+    const int n_cap = cap_n_no_;
```
ktbolt: @javijv4 Why use `const` here ?

Code/Source/solver/CoupledBoundaryCondition.cpp
```diff
@@ -487,63 +531,87 @@ void CoupledBoundaryCondition::initialize_cap(ComMod& com_mod)
...
+    }
+
+    // Count the number of values to pack for each node.
     const int per_node = 1 + 3 * nsd + (gather_Y ? 2 * nsd : 0);
-    const int nPack = tnNo > 0 ? tnNo : 0;
-    Vector<double> send_buf(static_cast<size_t>(nPack * per_node));
+    int n_pack = 0;
+    for (int Ac = 0; Ac < tnNo; Ac++) {
+        const int g = com_mod.ltg(Ac);
+        if (cap_gn_set.find(g) != cap_gn_set.end()) {
```
ktbolt, line 591: @javijv4 You can just use `cap_gn_set.count()` here because you are not using the iterator.

Code/Source/solver/CoupledBoundaryCondition.cpp
```diff
@@ -596,56 +663,60 @@ void gather_global_mesh_state_parallel(ComMod& com_mod, const CmMod& cm_mod, cmT
         out.Yn = 0.0;
     }
 
-    // Unpack the global mesh state from the receive buffer.
+    // Unpack the cap global mesh state.
     int pos = 0;
     while (pos + per_node <= static_cast<int>(recv_buf.size())) {
-        int g = static_cast<int>(recv_buf(pos++));
+        const int g = static_cast<int>(recv_buf(pos++));
+        const auto it = g_to_cap_col.find(g);
```
ktbolt: @javijv4 Why `const` here ?

Code/Source/solver/CoupledBoundaryCondition.cpp
javijv4: Hmm.. because it's a read-only value? It would work either way, so I can remove it if you prefer that.

Code/Source/solver/CoupledBoundaryCondition.cpp
javijv4: Yeah, I don't really need that. I'll remove it and check if there are other places in the code where I am doing this.

ktbolt:
@javijv4 I think most of the `static_cast<int>` and `static_cast<size_t>` can be removed unless it is needed to convert int to double and such.

javijv4:
@ktbolt, okay. I'll go through the code and remove those. Should I also remove the unnecessary `const` for the read-only variables?

ktbolt:
@javijv4 Maybe keep the `const` for the read-only variables for large functions where `nsd` and such are set at the beginning of the function; I see that this is done in other code.

ktbolt (approved):
Looks good !

Code/Source/linear_solver/add_bc_mul.cpp
```diff
@@ -58,7 +58,27 @@ void add_bc_mul(FSILS_lhsType& lhs, const BcopType op_Type, const int dof, const
         }
         // Computing S = coef * v^T * X
         double S = coef(faIn) * dot::fsils_dot_v(dof, lhs.mynNo, lhs.commu, v, X);
-
+        
+        // Add cap surface contribution to S
```
aabrown100-git, line 62: This contribution should greatly improve nonlinear convergence when using a cap surface. @javijv4 can you confirm that in your struct with cap test case, nonlinear convergence is good?

Code/Source/solver/baf_ini.cpp
```diff
@@ -743,7 +764,7 @@ void fsi_ls_ini(ComMod& com_mod, const CmMod& cm_mod, bcType& lBc, const faceTyp
       fsils_bc_create(com_mod.lhs, lsPtr, lFa.nNo, nsd, BcType::BC_TYPE_Dir, gNodes, sVl); 
     }
 
-  } else if (btest(lBc.bType, iBC_Neu)) {
+  } else if (btest(lBc.bType, iBC_Neu) || btest(lBc.bType, iBC_Coupled)) {
```
aabrown100-git: We only need to compute this integral of normal vector for coupled Neumann BCs, right? Is `iBC_coupled` for all coupled BCs (Neumann and Dirichlet), or just Neumann?

Code/Source/solver/CoupledBoundaryCondition.h
```diff
@@ -0,0 +1,419 @@
...
+                                                     const std::unordered_map<int, int>& gn_to_cap_local) const;
+    
+        /// @brief Compute the Jacobian and normal vector for a given element and Gauss point.
+        std::pair<double, Vector<double>> compute_jacobian_and_normal(const Array<double>& xl, int e, int g) const;
+
+        /// @brief Cap contribution to the linear solver face; \c mutable so it can be refreshed under \c const *this.
+        mutable Array<double> valM_;
+    };
+
+/// @brief Object-oriented Coupled boundary condition on a cap face
```
aabrown100-git: Is this class only for Coupled BC with cap? If not, can you edit this to be more general and descriptive?

Code/Source/solver/trilinos_impl.cpp
```diff
@@ -1144,13 +1144,25 @@ void TrilinosLinearAlgebra::TrilinosImpl::init_dir_and_coup_neu(ComMod& com_mod,
...
-      // Compute the coupled Neumann BC vector and store it in v
+      // Compute the coupled Neumann BC vector and store it in v (main face contribution)
       for (int a = 0; a < face.nNo; a++) {
         int Ac = face.glob(a);
         for (int i = 0; i < faDof; i++) {
           v(i,Ac) = v(i,Ac) + sqrt(fabs(res(faIn))) * face.val(i,a);
         }
       }
+      
+      // Add cap contribution (if cap exists for this face)
```
aabrown100-git, line 1176: Did you check that this actually works with Trilinos?

tests/cases/struct/LV_NeoHookean_passive_sv0D_cap/README.md
```diff
@@ -0,0 +1,82 @@
...
+  <Follower_pressure_load> true </Follower_pressure_load>
+  <Coupling_interface>
+    <svZeroDSolver_block> LV_IN </svZeroDSolver_block>
+    <Chamber_cap_surface> mesh/mesh-surfaces/endo_cap.vtp </Chamber_cap_surface>
+  </Coupling_interface>
+</Add_BC>
+```
+where 
+- `Time_dependence` set to `Coupled` with `Coupling_interface` selects the CoupledBoundaryCondition path for svZeroDSolver
+- `svZeroDSolver_block` names the svZeroDSolver block; optional `Chamber_cap_surface` gives the cap VTP for flow integration
```
aabrown100-git, line 80: Can you add another section about the cap surface since it might not be intuitive why we need it? Maybe also add a link to my paper.

tests/cases/struct/LV_NeoHookean_passive_sv0D_cap/README.md
```diff
@@ -0,0 +1,82 @@
...
+
```
aabrown100-git, line 85: Also, can you add the script you used to generate the cap mesh to the `utilities` folder?

aabrown100-git (changes requested):

Code/Source/linear_solver/add_bc_mul.cpp
aabrown100-git: Actually, I checked the Tests for the most recent commit, and nonlinear convergence seems to deteriorate for the cap test case with 3 and 4 procs. Can you check this? This might indicate a bug in this part of the code.

javijv4: @aabrown100-git, yes. This was already implemented in the GenBC cap implementation from @dseyler. I am getting comparable convergence in the cap test case.

**New sv0D implementation**
```
 ST 96-1  2.116e+01  [0 1.000e+00 9.776e-01 9.844e-11]  [53 -63 3]
 ST 96-2  2.121e+01  [-78 1.171e-04 1.145e-04 7.939e-10]  [110 -28 17]
 ST 96-3  2.124e+01  [-92 2.351e-05 2.298e-05 8.514e-10]  [28 -209 5]
 ST 96-4  2.128e+01  [-166 4.493e-09 4.392e-09 1.232e-08]  [58 -93 8]
 ST 96-5s 2.132e+01  [-186 4.480e-10 4.379e-10 1.404e-07]  [23 -158 2]
 ```
 
 **GenBC implementation**
 ```
 ST 96-1  1.980e+01  [0 1.000e+00 9.776e-01 8.778e-10]  [27 -209 1]
 ST 96-2  1.984e+01  [-78 1.171e-04 1.145e-04 8.933e-10]  [51 -208 11]
 ST 96-3  1.987e+01  [-92 2.351e-05 2.299e-05 8.582e-10]  [36 -209 6]
 ST 96-4  1.991e+01  [-164 5.766e-09 5.637e-09 1.306e-08]  [58 -83 11]
 ST 96-5s 1.994e+01  [-187 4.442e-10 4.343e-10 1.378e-07]  [23 -158 3]
 ```

In a few timesteps (see below), the convergence stalls for a bit (it also happens for GenBC), but I think it is still good. 
```
 ST 77-1  1.643e+01  [0 1.000e+00 9.855e-01 7.552e-10]  [58 -31 4]
 ST 77-2  1.647e+01  [-81 8.252e-05 8.132e-05 8.150e-10]  [108 -16 16]
 ST 77-3  1.651e+01  [-97 1.259e-05 1.241e-05 1.456e-10]  [53 -70 10]
 ST 77-4  1.655e+01  [-174 1.810e-09 1.783e-09 3.851e-08]  [54 -78 8]
 ST 77-5s 1.660e+01  [-197 1.336e-10 1.316e-10 4.772e-07]  [55 -3 9]
```

Comment (aabrown100-git) [reply Code/Source/linear_solver/add_bc_mul.cpp:62]: What test is this? Also, is this run with 1 proc or multiple procs?

Comment (javijv4) [reply Code/Source/linear_solver/add_bc_mul.cpp:62]: `tests/cases/struct/LV_NeoHookean_passive_sv0D_cap`, but running it for 200 timesteps. The GenBC implementation is using @dseyler's branch. 
I thought I was running in parallel, but no. It was serial... and the convergence is not the same when I ran in parallel. I probably broke something when I cleaned up the code last week (it was working before).  I'll work on it.

File: Code/Source/solver/baf_ini.cpp (hunk 7 above)
Comment (javijv4) [reply Code/Source/solver/baf_ini.cpp]: yep, you are right. `iBC_coupled` should work for Neumann or Dirichlet (although Dirichlet is not implemented yet). I'll add an additional check.

File: Code/Source/solver/trilinos_impl.cpp (hunk 9 above)
Comment (javijv4) [reply Code/Source/solver/trilinos_impl.cpp:1176]: Nope. I'll check this.

File: tests/cases/struct/LV_NeoHookean_passive_sv0D_cap/README.md (hunk 11 above)
Comment (javijv4) [reply tests/cases/struct/LV_NeoHookean_passive_sv0D_cap/README.md:85]: I did not create it. I just used the one available in @dseyler test. @dseyler do you think we can put that code in the `utilities` as part of this PR?

File: Code/Source/linear_solver/add_bc_mul.cpp (hunk 6 above)
Comment (aabrown100-git) [reply Code/Source/linear_solver/add_bc_mul.cpp:62]: Okay let me know if I can help!

File: Code/Source/solver/baf_ini.cpp (hunk 7 above)
Comment (aabrown100-git) [reply Code/Source/solver/baf_ini.cpp]: If I remember correctly, `lBc.bType` can satisfy multiple BC types. So for example a Neumann Coupled BC can be both `iBC_Neu` and `iBC_Coupled`. @ktbolt is that correct? If so, then here I think we only need to check`iBC_Coupled`?

Comment (ktbolt) [reply Code/Source/solver/baf_ini.cpp]: @aabrown100-git Correct, the individual bits of lBc.bType are set for each BC type so it can store multiple BC types.

Comment (javijv4) [reply Code/Source/solver/baf_ini.cpp]: Ah, I did not know that, and I was treating it like a single value. That makes things easier. I will update the BC type checks.

File: Code/Source/linear_solver/add_bc_mul.cpp (hunk 6 above)
Comment (javijv4) [reply Code/Source/linear_solver/add_bc_mul.cpp:62]: Okay. Fixed. The convergence in parallel follows the serial behavior. 
```
 ST 96-1  1.086e+01  [0 1.000e+00 9.776e-01 7.894e-10]  [28 -210 0]
 ST 96-2  1.088e+01  [-78 1.171e-04 1.145e-04 8.923e-10]  [66 -111 10]
 ST 96-3  1.090e+01  [-92 2.351e-05 2.298e-05 8.243e-11]  [53 -61 10]
 ST 96-4  1.092e+01  [-166 4.492e-09 4.391e-09 1.333e-08]  [61 -78 11]
 ST 96-5s 1.094e+01  [-187 4.465e-10 4.364e-10 1.413e-07]  [22 -158 4]
 ```

Code/Source/solver/baf_ini.cpp
javijv4: I revised all checks for `lBc.bType` to account for the fact that it contains multiple BC types. 
@aabrown100-git, the code you highlighted now only checks if the BC is Neumann. There's a second check after that `if` that verifies whether the BC type is `iBC_res`, which is set only for coupled BCs (including GenBC and RCR). I am not sure why we need that `iBC_res`, but setting it was the original path, so I kept it.

Code/Source/linear_solver/add_bc_mul.cpp
aabrown100-git: Thanks for fixing! What was the bug? Just curious.

Code/Source/solver/baf_ini.cpp
aabrown100-git: Okay sounds good. I think `iBC_res` is set for all "resistance" like boundary conditions, which includes resistance, RCR, and 0D-coupled BCs. The code I highlighted computes the integral of the normal vector, which is indeed necessary for "resistance" like BCs.

We might want to set `iBC_res` for `CoupledBoundaryCondition`?

javijv4: Yeah, it is being set!

Code/Source/linear_solver/add_bc_mul.cpp
javijv4: So the valM for the cap is computed only in the master and then broadcast to all ranks (because all ranks go into the linear solver and use valM in `add_bc_mul.cpp`). 
The problem was that I was passing the contribution to all nodes in the rank, including the ghost nodes. So a shared node had a value in all the ranks. Then their contribution was doubled in the calculations in `add_bc_mul.cpp`. 
I just needed to restrict the broadcast to the nodes owned by each rank, so that shared nodes aren't counted twice.

Code/Source/solver/trilinos_impl.cpp
javijv4: Ok, there were some issues. I fixed them and it is working (tested in sherlock)

Code/Source/linear_solver/add_bc_mul.cpp
aabrown100-git: Very interesting, nice find!

tests/cases/struct/LV_NeoHookean_passive_sv0D_cap/README.md
aabrown100-git: [generate_cap.py](https://github.com/user-attachments/files/27251461/generate_cap.py)
Actually I have it right here. Can you double check this produces the cap surface you want and add it to `utilities`?

Code/Source/solver/trilinos_impl.cpp
aabrown100-git: Perfect, do you want to add a Trilinos test case too then?

tests/cases/struct/LV_NeoHookean_passive_sv0D_cap/README.md
javijv4: Okay, I checked, and it generates the same surface as in the tests. I just added the script to a new folder in utilities.

Code/Source/solver/trilinos_impl.cpp
javijv4: Yeah, I just added it.

aabrown100-git (approved):
Everything looks good to me, let's merge!

PR #532, opened 2026-04-19
Allow zero fiber/sheet vectors in compute_sheet_normal

dseyler:
## Current situation
compute_sheet_normal in mat_models.cpp threw an error when the cross product of fiber and sheet vectors was below some tolerance (i.e. they are parallel.) This prevented assigning zero vector fibers on regions of the mesh where active/passive fibers should not contribute. This PR skips the runtime error message when the vectors have zero magnitude but raises an error when nonzero vectors are parallel.

## Release Notes 
Fixed compute_sheet_normal to allow zero fiber/sheet direction vectors. Regions with zero vectors now contribute no anisotropic stress, rather than causing a runtime error.

```
// Zero fiber/sheet vector means no fibers in this region: return zero so Hnn = n*n^T = 0.
    static constexpr double fib_zero_tol = 1.0e-10;
    if (fl.col(0).norm() < fib_zero_tol || fl.col(1).norm() < fib_zero_tol) {
      return Eigen::Matrix<double, nsd, 1>::Zero();
    }
```

## Documentation
 
*Please ensure that you properly document any additions.*

## Testing
 
*In addition, please ensure all modified and new code is covered by tests.*

### DIFF
--- a/Code/Source/solver/mat_models.cpp
+++ b/Code/Source/solver/mat_models.cpp
@@ -295,12 +295,18 @@ Eigen::Matrix<double, nsd, 1> compute_sheet_normal(const Eigen::Matrix<double, n
     throw std::runtime_error("Sheet-normal active stress (eta_n > 0) is not defined in 2D.");
     
   } else {  // nsd == 3
+    // Zero fiber/sheet vector means no fibers in this region: return zero so Hnn = n*n^T = 0.
+    static constexpr double fib_zero_tol = 1.0e-10;
+    if (fl.col(0).norm() < fib_zero_tol || fl.col(1).norm() < fib_zero_tol) {
+      return Eigen::Matrix<double, nsd, 1>::Zero();
+    }
+
     auto n_normal = cross_product<nsd>(fl.col(0), fl.col(1));
     double norm_n = sqrt(n_normal.dot(n_normal));
-    
-    static constexpr double sheet_normal_tol = 1.0e-10; 
+
+    static constexpr double sheet_normal_tol = 1.0e-10;
     if (norm_n < sheet_normal_tol) {
-      throw std::runtime_error("Fiber and sheet directions are parallel; sheet-normal direction is undefined.");
+      throw std::runtime_error("Fiber and sheet directions are non-zero but parallel; sheet-normal direction is undefined.");
     }
 
     return n_normal / norm_n;


aabrown100-git:
@dseyler Can you explain the use case a little more? Why doesn't defining a new domain for the isotropic material work?

Technically speaking, I think $[0,0,0]$ is parallel to $[0,0,0]$, so if we're making this change, we should update the function description and maybe the function name.

ktbolt:
@dseyler A developer previously decided that `compute_sheet_norma()` should throw an error when the cross product of fiber and sheet vectors are approximately parallel. You have now determined that this is not correct. How do you justify this decision ? Do you guarantee that allowing parallel vectors will not impact other code or cause simulation problems ?

Is there an Issue describing these changes ? 

This is exactly the kind of code change I mentioned in the last svMultiPhysics meeting that remains a mystery without a an Issue describing the history / justification of the code change.

dseyler:
@ktbolt I believe @aabrown100-git added this check for the fiber stress distribution PR. The nonzero cross product check asserts that the fiber normal direction can be defined relative to the fiber and sheet directions. This means that every point on the mesh has to have nonparallel fiber directions even if they don't contribute to passive or active stress. Adding a second domain does not avoid this error since fibers are assigned across the whole mesh, not by domain.

I had been previously assigning 0 vectors to regions of the mesh that I do not want to contract (i.e. slabs closing off ventricles). As of the fiber stress distribution PR, this method no longer works, and I have to assign arbitrary fiber directions to non-contractile domains. This makes generating fiber plots in Paraview difficult, because it will trace streamlines across the "fake" fibers, and is also unintuitive for users who will have to write a script to generate "fake" non-parallel fibers. 

Perhaps a better fix than skipping the error message when fibers are zero vectors would be to only apply the parallel fibers check on elements that lie in domains with active contraction or to have fibers assigned on the domain level, rather than across the whole mesh.

Let me know your thoughts or if it would be better to open an issue first. Just thought I'd open a PR directly since it's only a few lines of code and I needed this feature for my upcoming paper deadline.

The changes here still correctly require nonzero fibers to be nonparallel, and zero-magnitude fibers will behave as they did prior to the distributed fiber stress PR.

dseyler:
Will close this and open an issue

PR #534, opened 2026-04-22
Compute signed distance function for RIS valve only on fluid domain

### LINKED ISSUE #533: Signed distance function computed on unnecessary domains for RIS valve
### Description

The `uris_calc_sdf` function in the RIS valve code `uris.cpp` evaluates the signed distance function at all mesh nodes, including nodes not assigned to fluid equations, which was unnecessary for the RIS valve.

### Reproduction

Run the `pipe_uris_fsi` test case. The computed SDF fields `URIS_SDF_AV` and `URIS_SDF_AV` include the solid wall domain.

### Expected behavior

The SDF field should be computed only on the fluid domain.

### Additional context

_No response_

### Code of Conduct

hanzhao2020:
## Current situation
Resolves https://github.com/SimVascular/svMultiPhysics/issues/533

## Release Notes 
* Added an if statement in `uris_calc_sdf` function to restrict SDF evaluation to nodes that belong to fluid related equation domains (fluid / CMM / Stokes), instead of computing across all global nodes.

### DIFF
--- a/Code/Source/solver/ComMod.h
+++ b/Code/Source/solver/ComMod.h
@@ -1608,6 +1608,10 @@ class ComMod {
     /// @brief URIS resistance when the valve is closed
     double urisResClose;
 
+    /// @brief Fluid-related node mask for URIS SDF. Built once when
+    /// consistent with tnNo; rebuilt automatically if tnNo changes.
+    std::vector<char> urisFluidNodeMask;
+
     /// @brief Whether to use precomputed state-variable solutions
     bool usePrecomp = false;
     //----- int members -----//
--- a/Code/Source/solver/ComMod.h
+++ b/Code/Source/solver/ComMod.h
@@ -1610,7 +1610,7 @@ class ComMod {
 
     /// @brief Fluid-related node mask for URIS SDF. Built once when
     /// consistent with tnNo; rebuilt automatically if tnNo changes.
-    std::vector<char> urisFluidNodeMask;
+    std::vector<bool> urisFluidNodeMask;
 
     /// @brief Whether to use precomputed state-variable solutions
     bool usePrecomp = false;
--- a/Code/Source/solver/uris.cpp
+++ b/Code/Source/solver/uris.cpp
@@ -1163,12 +1163,12 @@ void uris_find_closest_face_centroid(const urisType& uris_obj, const Vector<doub
     for (int e = 0; e < mesh.nEl; e++) {
       xb = 0.0;
       for (int a = 0; a < mesh.eNoN; a++) {
-        int Ac = mesh.IEN(a,e);
+        const int Ac = mesh.IEN(a,e);
         xb = xb + uris_obj.x.rcol(Ac);
       }
       xb = xb / mesh.eNoN;
 
-      double dS = std::sqrt((xp - xb) * (xp - xb));
+      const double dS = std::sqrt((xp - xb) * (xp - xb));
 
       if (dS < minS) {
         minS = dS;
--- a/Code/Source/solver/uris.cpp
+++ b/Code/Source/solver/uris.cpp
@@ -459,23 +459,21 @@ void inside_tet(ComMod& com_mod, int& eNoN, Vector<double>& xp,
 }
 
 /// @brief Precompute whether each node belongs to a fluid-related domain.
-std::vector<char> uris_build_fluid_node_mask(ComMod& com_mod)
-{
+void uris_build_fluid_node_mask(ComMod& com_mod) {
   using namespace consts;
 
-  std::vector<char> in_fluid_domain(com_mod.tnNo, 0);
+  com_mod.urisFluidNodeMask.resize(com_mod.tnNo, 0);
   for (int a = 0; a < com_mod.tnNo; ++a) {
     for (int iEq = 0; iEq < com_mod.nEq; ++iEq) {
       const auto& eq = com_mod.eq[iEq];
       if (all_fun::is_domain(com_mod, eq, a, Equation_fluid) ||
           all_fun::is_domain(com_mod, eq, a, Equation_CMM) ||
           all_fun::is_domain(com_mod, eq, a, Equation_stokes)) {
-        in_fluid_domain[a] = 1;
+        com_mod.urisFluidNodeMask[a] = 1;
         break;
       }
     }
   }
-  return in_fluid_domain;
 }
 
 /// @brief Read the URIS mesh separately 
@@ -878,7 +876,7 @@ void uris_write_vtus(ComMod& com_mod) {
 void uris_calc_sdf(ComMod& com_mod) {
   #define n_debug_uris_calc_sdf 
   #ifdef debug_uris_calc_sdf
-  DebugMsg dmsg(__func__, simulation->com_mod.cm.idcm());
+  DebugMsg dmsg(__func__, com_mod.cm.idcm());
   dmsg.banner();
   #endif
 
@@ -889,8 +887,9 @@ void uris_calc_sdf(ComMod& com_mod) {
   const int nsd = com_mod.nsd;
   const int nUris = com_mod.nUris;
 
-  // Build once per SDF recompute call.
-  const auto fluid_node_mask = uris_build_fluid_node_mask(com_mod);
+  if (com_mod.urisFluidNodeMask.size() != static_cast<size_t>(com_mod.tnNo)) {
+    uris_build_fluid_node_mask(com_mod);
+  }
 
   Array<double> xXi(nsd, nsd-1);
 
@@ -914,7 +913,6 @@ void uris_calc_sdf(ComMod& com_mod) {
       }
     }
     
-    // if (uris_obj.sdf.allocated() && cnt < uris_obj.cnt) {continue;}
     if (uris_obj.sdf.size() > 0 && cnt < uris_obj.cnt) {continue;}
 
     int max_eNoN = 0;
@@ -926,7 +924,6 @@ void uris_calc_sdf(ComMod& com_mod) {
     }
 
     Array<double> lX(nsd, max_eNoN);
-    // if (!uris_obj.sdf.allocated()) {
     if (uris_obj.sdf.size() <= 0) {
       uris_obj.sdf.resize(com_mod.tnNo);
       uris_obj.sdf = 0.0;
@@ -985,7 +982,7 @@ void uris_calc_sdf(ComMod& com_mod) {
         continue;
       }
 
-      if (!fluid_node_mask[ca]) {
+      if (!com_mod.urisFluidNodeMask[ca]) {
         continue;
       }
 
@@ -1113,9 +1110,9 @@ int in_poly(Vector<double>& P, Array<double>& P1, bool ext) {
 /// @brief Chech if a point is on the same side of anotehr point wrt a triangle in 3D
 int same_side(Vector<double>& v1, Vector<double>& v2, Vector<double>& v3,
               Vector<double>& v4, Vector<double>& p, bool ext) {
-  #define n_dbg_read_sv
-  #ifdef dbg_read_sv
-    DebugMsg dmsg(__func__, simulation->com_mod.cm.idcm());
+  #define n_dbg_same_side
+  #ifdef dbg_same_side
+    DebugMsg dmsg(__func__, 0);
     dmsg.banner();
     dmsg << "checking same side";
   #endif
@@ -1155,7 +1152,7 @@ void uris_find_closest_face_centroid(const urisType& uris_obj, const Vector<doub
                                      const int nsd, double& minS, int& Ec, int& jM) {
   #define n_dbg_uris_find_closest_face_centroid
   #ifdef dbg_uris_find_closest_face_centroid
-    DebugMsg dmsg(__func__, simulation->com_mod.cm.idcm());
+    DebugMsg dmsg(__func__, 0);
     dmsg.banner();
     dmsg << "finding closest face centroid";
   #endif
@@ -1188,7 +1185,7 @@ double uris_compute_face_dotp(const urisType& uris_obj, const int nsd, const int
                               Array<double>& lX, Vector<double>& xb) {
   #define n_dbg_uris_compute_face_dotp
   #ifdef dbg_uris_compute_face_dotp
-    DebugMsg dmsg(__func__, simulation->com_mod.cm.idcm());
+    DebugMsg dmsg(__func__, 0);
     dmsg.banner();
     dmsg << "computing face dot product";
   #endif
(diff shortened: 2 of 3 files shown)


Code/Source/solver/uris.cpp
```diff
@@ -978,101 +980,118 @@ void uris_calc_sdf(ComMod& com_mod) {
...
+          }
+          dS = std::sqrt(dS);
+
+          if (dS < minS) {
+            minS = dS;
+            Ec = e;
+            jM = iM;
           }
         }
+      }
```
aabrown100-git, line 1042: Can you put this in a function for readability and encapsulation?

Code/Source/solver/uris.cpp
```diff
@@ -978,101 +980,118 @@ void uris_calc_sdf(ComMod& com_mod) {
...
-            for (int j = 0; j < nsd; j++)
-                xXi(j,i) += lX(j,a) * factor;
-          }
+      for (int a = 0; a < mesh.eNoN; a++) {
+        for (int i = 0; i < nsd - 1; i++) {
+          double factor = mesh.Nx(i,a,0);
+          for (int j = 0; j < nsd; j++)
+              xXi(j,i) += lX(j,a) * factor;
         }
+      }
```
aabrown100-git, line 1067: Same for this. Put in function if possible

Code/Source/solver/uris.cpp
```diff
@@ -978,101 +980,118 @@ void uris_calc_sdf(ComMod& com_mod) {
...
-            dotP = -1.0;
-          } else {
-            dotP = 1.0;
-          }
+      auto nV = utils::cross(xXi);
+      auto Jac = sqrt(utils::norm(nV));
+      nV = nV / Jac;
+      auto dotP = utils::norm(xp-xb, nV);
+
+      // Improved implementation for SDF sign. For closed state, sign is less
```
aabrown100-git, line 1074: I think this can be in a function too

aabrown100-git (changes requested):
Looks good! Just thought it would be better to organize this code into a few functions.

Code/Source/solver/uris.cpp
```diff
@@ -978,101 +980,118 @@ void uris_calc_sdf(ComMod& com_mod) {
...
+        for (int e = 0; e < mesh.nEl; e++) {
+          xb = 0.0;
+          for (int a = 0; a < mesh.eNoN; a++) {
+            int Ac = mesh.IEN(a,e);
+            for (int i = 0; i < nsd; i++) {
+              xb(i) += uris_obj.x(i,Ac);
+            }
+          }
+          for (int i = 0; i < nsd; i++) {
+            xb(i) /= static_cast<double>(mesh.eNoN);
```
ktbolt, line 1028: @hanzhao2020 No need to cast; compiler will convert the int to a double.

ktbolt:
@hanzhao2020 You are doing a lot of vector computations over `nsd` that could be cleaner. 

For example
```
for (int i = 0; i < nsd; i++) {
  xb(i) += uris_obj.x(i,Ac);
  lX(i,a) = uris_obj.x(i,Ac);
 }

for (int i = 0; i < nsd; i++) {
  dS += (xp[i] - xb[i]) * (xp[i] - xb[i]);
}
```
could be replaced with
```
xb = xb + uris_obj.x.rcol(Ac);
lX.rcol(a) = uris_obj.x.rcol(Ac);

ds = (xb - xp) * (xb - xp);
```
Note that `Vector` will make copies for some operations so if that is a problem then you could use `Eigen` for local vectors.

Code/Source/solver/uris.cpp
hanzhao2020: @aabrown100-git Thanks for the suggestion! Just updated the SDF function into smaller functions.

Code/Source/solver/uris.cpp
hanzhao2020: Done!

Code/Source/solver/uris.cpp
hanzhao2020: It was put in a function.

Code/Source/solver/uris.cpp
hanzhao2020: @ktbolt Thanks for reviewing. I removed the `static_cast`

hanzhao2020:
> You are doing a lot of vector computations over nsd that could be cleaner.

@ktbolt Thanks for the suggestion! I changed the `nsd` related vector computation to `rcol`

Code/Source/solver/uris.cpp
```diff
@@ -978,101 +980,118 @@ void uris_calc_sdf(ComMod& com_mod) {
...
+      bool is_fluid_domain = false;
+      for (int iEq = 0; iEq < com_mod.nEq; ++iEq) {
+        const auto& eq = com_mod.eq[iEq];
+        if (all_fun::is_domain(com_mod, eq, ca, Equation_fluid) ||
+            all_fun::is_domain(com_mod, eq, ca, Equation_CMM) ||
+            all_fun::is_domain(com_mod, eq, ca, Equation_stokes)) {
+          is_fluid_domain = true;
+          break;
         }
+      }
```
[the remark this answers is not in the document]
hanzhao2020: The fluid domain check is moved to the beginning of the SDF calculation function. It's performed when the SDF field is recomputed.

Code/Source/solver/uris.cpp
```diff
@@ -466,6 +458,25 @@ void inside_tet(ComMod& com_mod, int& eNoN, Vector<double>& xp,
...
+  using namespace consts;
+
+  std::vector<char> in_fluid_doamin(com_mod.tnNo, 0);
+  for (int a = 0; a < com_mod.tnNo; ++a) {
+    for (int iEq = 0; iEq < com_mod.nEq; ++iEq) {
+      const auto& eq = com_mod.eq[iEq];
+      if (all_fun::is_domain(com_mod, eq, a, Equation_fluid) ||
+          all_fun::is_domain(com_mod, eq, a, Equation_CMM) ||
+          all_fun::is_domain(com_mod, eq, a, Equation_stokes)) {
+        in_fluid_doamin[a] = 1;
```
aabrown100-git: Misspelled "domain"

aabrown100-git (changes requested):

Code/Source/solver/uris.cpp
hanzhao2020: Thanks for catching! Fixed

Code/Source/solver/uris.cpp
```diff
@@ -466,6 +458,25 @@ void inside_tet(ComMod& com_mod, int& eNoN, Vector<double>& xp,
...
+
+  std::vector<char> in_fluid_domain(com_mod.tnNo, 0);
+  for (int a = 0; a < com_mod.tnNo; ++a) {
+    for (int iEq = 0; iEq < com_mod.nEq; ++iEq) {
+      const auto& eq = com_mod.eq[iEq];
+      if (all_fun::is_domain(com_mod, eq, a, Equation_fluid) ||
+          all_fun::is_domain(com_mod, eq, a, Equation_CMM) ||
+          all_fun::is_domain(com_mod, eq, a, Equation_stokes)) {
+        in_fluid_domain[a] = 1;
+        break;
```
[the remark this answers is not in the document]
hanzhao2020: Fixed

Code/Source/solver/uris.cpp
```diff
@@ -892,11 +882,16 @@ void uris_calc_sdf(ComMod& com_mod) {
...
+  using namespace consts;
+
   auto& cm = com_mod.cm;
   auto& uris = com_mod.uris;
   const int nsd = com_mod.nsd;
   const int nUris = com_mod.nUris;
 
+  // Build once per SDF recompute call.
+  const auto fluid_node_mask = uris_build_fluid_node_mask(com_mod);
+
```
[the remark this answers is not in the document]
hanzhao2020: The new commit caches the fluid node mask across iterations and only recomputes if the number of nodes is changed.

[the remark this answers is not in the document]
aabrown100-git: Nice, so for most cases, `com_mod.urisFluidNodeMask` is only computed once per simulation?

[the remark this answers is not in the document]
hanzhao2020: Yes, the associated domain for each node should stay the same for most cases and will not change during simulation.

Code/Source/solver/ComMod.h
```diff
@@ -1608,6 +1608,10 @@ class ComMod {
     /// @brief URIS resistance when the valve is closed
     double urisResClose;
 
+    /// @brief Fluid-related node mask for URIS SDF. Built once when
+    /// consistent with tnNo; rebuilt automatically if tnNo changes.
+    std::vector<char> urisFluidNodeMask;
```
aabrown100-git: Sorry, just one more thing. Does this have to be in `ComMod`? I think we're trying to move things out of `ComMod`, so better to not add something to it if possible.

Also, why is it a vector of `char`?

hanzhao2020: I think we need to put it in `com_mod` if we want to use it across different time steps. I'm not sure if there is another way to do this.

Using a vector of `char` is suggested by AI for the reason: `vector<char>` has a small memory footprint: 1 byte per node, and `vector<bool>` is bit-packed and uses proxy references, which can be awkward/slower in some patterns. 

I can change it back to `vector<bool>` if it's more consistent.

aabrown100-git: Ah okay, fine with me then. I think 'bool' would be better. @ktbolt what do you think about these points?

Code/Source/solver/uris.cpp
```diff
@@ -1220,4 +1147,90 @@ int same_side(Vector<double>& v1, Vector<double>& v2, Vector<double>& v3,
...
+      double dS = std::sqrt((xp - xb) * (xp - xb));
+
+      if (dS < minS) {
+        minS = dS;
+        Ec = e;
+        jM = iM;
+      }
+    }
+  }
+}
```
michelebucelli, line 1180: Not sure if this is the case here, or if you have profiled this, but in my previous implementation of RIIS the signed distance evaluation would take up a considerable chunk of the simulation time, since it was being evaluated on every (fluid) mesh node at every time step.

If I'm reading it right, the complexity of this currently scales like $O(N_{surf})$, with $N_{surf}$ the total number of surface centroids (which then needs to be multiplied by the total number of fluid nodes inside the bounding box $N_{fluid}$).

One possible improvement would be to use [R-Trees](https://en.wikipedia.org/wiki/R-tree) to speed up this closest point search. The R-tree would need to be built only once (or once every time the surface is displaced), with a complexity of $O(N_{surf}\log N_{surf})$. Each closest point query would then have an average complexity of $O(\log N_{surf})$, and you'd get the bounding box logic for free.

There's [an implementation of R-trees available through Boost](https://www.boost.org/doc/libs/latest/libs/geometry/doc/html/geometry/reference/spatial_indexes/boost__geometry__index__rtree.html).

Code/Source/solver/uris.cpp
```diff
@@ -1220,4 +1147,90 @@ int same_side(Vector<double>& v1, Vector<double>& v2, Vector<double>& v3,
...
+    const auto& mesh = uris_obj.msh[iM];
+    for (int e = 0; e < mesh.nEl; e++) {
+      xb = 0.0;
+      for (int a = 0; a < mesh.eNoN; a++) {
+        int Ac = mesh.IEN(a,e);
+        xb = xb + uris_obj.x.rcol(Ac);
+      }
+      xb = xb / mesh.eNoN;
+
+      double dS = std::sqrt((xp - xb) * (xp - xb));
```
michelebucelli: You may want to declare these local variables as `const`, as it aids compiler optimization:
```suggestion
        const int Ac = mesh.IEN(a,e);
        xb = xb + uris_obj.x.rcol(Ac);
      }
      xb = xb / mesh.eNoN;

      const double dS = std::sqrt((xp - xb) * (xp - xb));
```
This can lead to measurable performance improvements in some cases (not necessarily here though it usually is a good idea).

Code/Source/solver/uris.cpp
hanzhao2020: @michelebucelli Thanks for the suggestion! Yes, this closest point search takes quite some time for fine meshes. An R-tree (or similar spatial index) could definitely reduce the query cost. One thing to consider is that using Boost’s R-tree would make Boost a required dependency for svMultiPhysics. @ktbolt do you have any thoughts on whether that’s ok, or if we prefer to avoid adding Boost?

Code/Source/solver/uris.cpp
ktbolt: @michelebucelli Any decent compiler will be able to identify and optimize loop-level constants without declaring them `const`.

Code/Source/solver/uris.cpp
michelebucelli: Sorry, I thought that Boost was already in the list of dependencies, but upon double-checking I noticed that this is only true if building with the Trilinos backend.

ktbolt: @hanzhao2020 Like @michelebucelli suggested it would be good to understand how much this computation adds to the total simulation time. svMultiPhysics has a `Timer` class that you can use to measure the time spent in this code if you don't want to profile it.

There are several spatial search algorithms implemented in [Zach's branch](https://github.com/zasexton/svMultiPhysics/tree/issue-449-modern-mesh-core/Code/Source/solver/Mesh/Search).

And there may be a way to write something that takes advantage of the spatial coherence of the RIS geometry.

hanzhao2020: @michelebucelli Yes, Boost is only required when building with Trilinos so far. @ktbolt Sounds good! I can look them up and try using the tree-type algorithms to accelerate the closest-point search. I think this can be in a separate issue and PR.

Code/Source/solver/ComMod.h
hanzhao2020: @aabrown100-git Yes, using `bool` is more consistent. I changed it to `vector<bool>` for this fluid domain mask.

alisonmarsden:
Tagging @fkong7 in case she has input on this issue!

aabrown100-git (approved):
Looks good to me!

ktbolt (approved):
Looks good !

Code/Source/solver/uris.cpp
hanzhao2020: Mark this comment as resolved for now.

aabrown100-git:
@hanzhao2020 Good to merge?

hanzhao2020:
@aabrown100-git Yes, I think it's ready to merge. Thanks!

Issue #526, opened 2026-04-08
Implementing Core Exception Infrastructure


zasexton:
### Current State

The solver has some exception usage, but it lacks a solver-wide infrastructure that would be useful for a modular FE plan. 

  - Most production code throws generic standard exceptions, predominantly `std::runtime_error`, spread across many modules. Examples include `Code/Source/solver/Parameters.h:166, Code/Source/solver/Vector.h`
  - There is one localized custom exception family for boundary-condition handling in `Code/Source/solver/BoundaryCondition.h`, but it is subsystem-specific. It stores only a message string and is not tied to shared status codes, source locations, MPI context, or common solver macros.
  - Some failure paths bypass exceptions entirely and terminate the process directly, for example `Code/Source/solver/trilinos_impl.cpp`. 
  - The solver execution path also does not currently have a consistent top-level exception boundary. The `read_files()` catch is commented out in `Code/Source/solver/main.cpp`, while the simulation loop proceeds without a surrounding catch in `Code/Source/solver/main.cpp`.

### Solution

We want to implement a core exception system that can define failure handling into a set of shared contracts instead of a pile of unrelated strings and exits. This system will need to be able to accommodate the following:

  - A types list that provides a common vocabulary: semantic index types like `LocalIndex`, `GlobalIndex`, an `DofIndex` etc,, plus a status taxonomy. The current solver mixes raw int/double, backend-native codes like `PetscErrorCode`, and subsystem-local conventions, so there is no single way to say “this is invalid input” versus “this is a backend failure” versus “this is a convergence failure.”
  - `FEException` class to add a base exception carrying structured context: message, status, file/line/function, MPI rank, and optional debug stack trace. It also will define typed subclasses like `InvalidArgumentException` in  and standardized throw/check macros. 
  - Against the current code, that would give three immediate benefits: failures become classifiable instead of string-only, tests can assert on error categories instead of exact text, and MPI/backend failures can be surfaced consistently instead of sometimes throwing, sometimes aborting, and sometimes just printing. This will be crucial for testing modular subsystems in the future.

zasexton:
Branch: [`526-core-exceptions`](https://github.com/zasexton/svMultiPhysics/tree/526-core-exceptions)

[...]

PR #537, opened 2026-04-28
526 core exceptions

zasexton:
## Current situation
Related to #526.

adds a shared solver exception framework and starts migrating top-level application-driver error handling to it. The branch introduces `Core/Exception.h`, `Core/PlatformSupport.h`, and `FE/Common/FEException.h`, wires the new headers into the solver build, and updates XML parameter parsing plus `main.cpp` error handling.

This is important because it centralizes solver errors around consistent status codes, source locations, MPI rank awareness, and optional stack traces. It also replaces several XML parsing paths that could previously fall through as raw `std::out_of_range`, null text/attribute handling, or generic runtime errors with structured `svmp::ParseException` failures.

## Release Notes 

  - Added shared core exception infrastructure for solver-wide error reporting.
  - Added FE exception types built on the shared exception base.
  - Improved XML parameter parsing diagnostics by reporting missing or invalid attributes/text through
    `ParseException`.
  - Updated command-line solver error handling to report structured XML parse errors and clean up MPI
    consistently.
  - No user-facing input format changes needed

## Documentation

The new exception headers include inline documentation and named exception/status types. No separate user documentation is required because these comprise internal error handling without changing solver XML syntax or runtime options.

## Testing

  Built the solver target successfully:

 ```bash
 cmake --build cmake-build --target svMultiPhysics -j2
```
Remaining verification will come from GitHub action CI and current test infrastructure. 

### CODE REVIEW

File: Code/Source/solver/main.cpp [hunk 1]
```diff
@@ -691,4 +693,15 @@ int main(int argc, char *argv[])
     }
 
   MPI_Finalize();
+  return 0;
+
+  } catch (const svmp::ParseException& exception) {
+    if (mpi_rank == 0) {
+      std::cerr << "[svMultiPhysics] XML parse error:" << std::endl;
```
aabrown100-git: We would also catch FEException here, right? Do you want to add that here for completeness, even though the code currently doesn't throw any FEExceptions?

Code/Source/solver/main.cpp
```diff
@@ -691,4 +693,15 @@ int main(int argc, char *argv[])
...
+      std::cerr << exception.what() << std::endl;
+    }
+    svmp::ExceptionRuntime::abort_mpi_if_needed(EXIT_FAILURE);
+    svmp::ExceptionRuntime::finalize_mpi_if_needed();
+    return EXIT_FAILURE;
```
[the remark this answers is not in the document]
aabrown100-git: I like this suggestion. Could we have one general catch in main?

Code/Source/solver/main.cpp
aabrown100-git: Better to have one general catch for all Exceptions

zasexton: We can catch it as the `ParseException` since we already are conditionally checking on this exception type

aabrown100-git: Sorry I don't totally follow, could you explain that more?

Code/Source/solver/main.cpp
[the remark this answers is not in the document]
zasexton: yes, I agree. we should have a base exception for these runtime errors. I'll add this so that we don't need two levels of catching for one runtime exception so that we can be more concise in the code

Code/Source/solver/main.cpp
zasexton: So right now I've included the FEException class but in reality we're not encountering any FE specific exceptions in the main or parameters files (as we shouldn't) so i don't think we can catch an FEException here but perhaps a general base exception.

Code/Source/solver/FE/Common/FEException.h
```diff
@@ -0,0 +1,347 @@
...
+
+/**
+ * @file FEException.h
+ * @brief Exception hierarchy for error handling in the FE library
+ *
+ * This header defines FE-specific exception types that derive from the shared
+ * solver exception infrastructure in Core/Exception.h.
+ */
+
+#include "Core/Exception.h"
```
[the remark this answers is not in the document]
zasexton: resolved in additional commit

Code/Source/solver/Core/PlatformSupport.inl
```diff
@@ -0,0 +1,163 @@
...
+#elif defined(__GNUC__) || defined(__clang__)
+#  include <cxxabi.h>
+#  include <execinfo.h>
+#endif
+
+namespace svmp {
+
+inline PlatformCapabilities PlatformSupport::capabilities() noexcept
+{
+#if defined(_WIN32)
```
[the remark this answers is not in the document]
zasexton: I like the idea here of making the platform support private so that we do not accidentally include it in a bunch of other files.

Code/Source/solver/main.cpp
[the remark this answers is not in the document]
zasexton: I believe this should work:

```c++
  } catch (const svmp::ExceptionBase& exception) {
    if (mpi_rank == 0) {
      std::cerr << "[svMultiPhysics] ERROR: The svMultiPhysics program has failed due to unhandled exception." << std::endl;
      std::cerr << exception.what() << std::endl;
    }
    svmp::ExceptionRuntime::abort_mpi_if_needed(EXIT_FAILURE);
    svmp::ExceptionRuntime::finalize_mpi_if_needed();
    return EXIT_FAILURE;
  } catch (const std::exception& exception) {
    if (mpi_rank == 0) {
      std::cerr << "[svMultiPhysics] ERROR: The svMultiPhysics program has failed due to unhandled exception." << std::endl;
      std::cerr << exception.what() << std::endl;
    }
    svmp::ExceptionRuntime::abort_mpi_if_needed(EXIT_FAILURE);
    svmp::ExceptionRuntime::finalize_mpi_if_needed();
    return EXIT_FAILURE;
  }
```

I'm not the biggest fan of having multiple catches here but because we will still have other parts of the code catching `std::exception` type cases we must include it since it is derived from `std::exception`. Eventually once we have migrated internal code over to the unified exception system we can clean this up to one catch level

Code/Source/solver/main.cpp
zasexton: Apologies, thought this was a different line from the above comments. Yes, we will need general catching for all exception types. so we will have to have a level for `std::exception` right now. Because both `svmp::ExceptionBase` is derived and then `FEException` is a further derivation from that base, these alone will be insufficient as the code will still have many other standard library exceptions that are thrown at this current point in the refactor. `FEException` will be important once we start to include further classes for basis functions, quadrature, elements, etc .... its just that right now we don't have any specific infrastructure that we need to throw these for yet since we have not yet added such additional finite element infrastructure

zasexton:
Cancelling all of the CI runs except the last two so that we do not have to wait forever.

aabrown100-git (approved):
Looks good to me!

aabrown100-git:
@copilot resolve the merge conflicts in this pull request

mrp089 (approved):

Code/Source/solver/Core/Exception.h
```diff
@@ -0,0 +1,402 @@
...
+    std::string function_;
+    int mpi_rank_ = -1;
+    StackTrace stack_trace_;
+};
+
+class ExceptionFormatter final {
+public:
+    static std::string format(const ExceptionContext& context,
+                              const std::string& message,
+                              const char* subsystem_label)
```
michelebucelli: If I get this correctly, the point of using `const char*` instead of `std::string` for the `subsystem_label` argument is that it is optional - if set to `nullptr` the function falls back to the default value `"Exception"`.

What would you think about using an `std::string` with a default value, or an `std::optional`, to make this intent more clear?

Code/Source/solver/Core/Exception.h
```diff
@@ -0,0 +1,402 @@
...
+private:
+    StatusCode status_code_ = StatusCode::Unknown;
+    std::string file_;
+    int line_ = 0;
+    std::string function_;
+    int mpi_rank_ = -1;
+    StackTrace stack_trace_;
+};
+
+class ExceptionFormatter final {
```
michelebucelli: I see here and elsewhere that there's a few classes with only static methods (this one and `ExceptionRuntime` below). Is there any reason why these are not `namespace`s instead?

(The first thing that comes to mind is that classes can be used as template arguments while namespaces can't, and this can be a way of selecting policies, so I guess my question is also whether this is happening/planned somewhere).

Code/Source/solver/Core/Exception.h
```diff
@@ -0,0 +1,402 @@
...
+    }
+}
+
+template <class ExceptionT, class... Args>
+void check_arg(bool condition, SourceLocation location, Args&&... args)
+{
+    if (!condition) {
+        raise<ExceptionT>(location, std::forward<Args>(args)...);
+    }
+}
```
michelebucelli, line 390: Do you have any thoughts on having macros/functions for debug-mode-only exceptions?

These would apply to situations where the condition to be checked is computationally expensive, so that you may not want to test it in production code but only in development code.

(In the past I have used the different convention of using debug-only assertions for everything that is independent of user-specified parameters, and production assertions only for those checks that may fail due to incorrect user input, but that's not necessarily the best choice).

I think that would require a `#define` macro instead of a function, because if you pass the condition to a function then it already gets evaluated at the call site, so you want something that skips the evaluation of the condition if the debug flag is disabled.

Code/Source/solver/Core/Exception.h
```diff
@@ -0,0 +1,402 @@
...
+    {
+        context_.set_status_code(status);
+        context_.set_source_location(file, line, function);
+        context_.set_mpi_rank(ExceptionRuntime::query_mpi_rank());
+#if SVMP_EXCEPTION_DEBUG_MODE
+        context_.set_stack_trace(ExceptionRuntime::capture_stack_trace());
+#endif
+    }
+
+    virtual const char* subsystem_label() const noexcept = 0;
```
michelebucelli: What would you think about defining a `subsystem_label` member instead, to be passed as a constructor argument by derived classes? I think this would lead to a (very slightly) leaner implementation in the derived expression classes (although it would constrain you to know the subsystem label at construction time).

Code/Source/solver/FE/Common/FEException.h
```diff
@@ -0,0 +1,347 @@
...
+#include "Exception.h"
+
+#include <sstream>
+#include <string>
+#include <utility>
+
+namespace svmp {
+namespace FE {
+
+class FEException : public ExceptionBase {
```
michelebucelli, line 25: I think I don't understand very clearly what is the intended difference between `CoreException` and `FEException`. It seems to me that they have similar functionality, so I guess the difference is meant to be in their semantics? Could you clarify this a little bit?

Code/Source/solver/Parameters.cpp
```diff
@@ -93,13 +131,21 @@ std::string IncludeParametersFile::NAME = "Include_xml";
 
 IncludeParametersFile::IncludeParametersFile(const char* cfile_name)
 {
+  if (cfile_name == nullptr) {
+    svmp::raise<svmp::ParseException>(SVMP_HERE, "Include_xml requires a file name.");
+  }
```
michelebucelli: Would it make sense to use the `check_*` functions defined in `Exception.h` here (and in a few similar places in this file)?

Code/Source/solver/Parameters.cpp
```diff
@@ -201,8 +247,8 @@ void Parameters::set_equation_values(tinyxml2::XMLElement* root_element)
   auto add_eq_item = root_element->FirstChildElement(EquationParameters::xml_element_name_.c_str());
 
   while (add_eq_item) {
-    const char* eq_type;
-    auto result = add_eq_item->QueryStringAttribute("type", &eq_type);
+    const char* eq_type = require_xml_attribute(add_eq_item, "type", SVMP_HERE,
+        "No TYPE given in the XML <Add_equation type=TYPE> element.");
```
michelebucelli: Would it be possible to reduce some of the information that the caller is required to provide here? Seeing as `require_xml_attribute` is called in a relatively large number of places, I think this would help conciseness and readability of this file a bit.

It seems to me that the error messages passed to this function so far are all in the format `No [uppercase attribute name] given in the XML <[XML element name] [attribute name]=[uppercase attribute name]> element`. The attribute name is already passed to the function. The XML element name is not, but I think it can be retrieved from `add_eq_item->Name()`. Then, in principle the error message can be reconstructed inside ther `require_xml_attribute` call itself, without the need for it to be passed from outside.

I admit this sacrifices a bit of flexibility, but I think that it is reasonable for a function that returns a given XML attribute to always raise an error saying "attribute not found".

As for SVMP_HERE, I think there's no full replacement for passing it here. However, it may be acceptable to put it inside the `require_xml_attribute`, provided that the exception message always prints the stack trace so that the error can be traced back to where the `require_xml_attribute` function is called.

michelebucelli (approved):
Thank you @zasexton!

I have left a few comments, none of which is critical. This looks good to me!

Code/Source/solver/Core/Exception.h
zasexton: I like this idea! And since we do not need ownership, allocation, or copying because the subsystem labels are static I will use `std::string_view`. The messages will be maintained as `std::string` since they are dynamically built.

Code/Source/solver/Core/Exception.h
zasexton: Although we haven't produced any real use cases for these yet I do agree that they would probably be better as namespaces since they are more like utilities instead of policies. I'll also propagate this forward into other modules.

Code/Source/solver/Core/Exception.h
zasexton: As long as we centralize the most of the macro definitions into the core folder then I'm perfectly fine with adding more macros :)

Code/Source/solver/Core/Exception.h
zasexton: For the most part we should know the subsystem labels at construction time so I'll remove the virtual `subsystem_label()` and move it into `ExceptionBase` to pass it from derived exception constructors.

Code/Source/solver/FE/Common/FEException.h
zasexton: Will make the comments more clear for purpose per general versus finite element specific exceptions.

Code/Source/solver/Parameters.cpp
zasexton: I can add it in a few more places but it does seem to make some of the XML parsing branches less readable. I guess this is mostly has been a stylistic preference of mine

Code/Source/solver/Parameters.cpp
zasexton: I can replace some more of the `require_xml_attribute` information so that we get standard missing attribute messaging alongside the `SVMP_HERE` macro which i rather like.

Issue #530, opened 2026-04-15
Object oriented refactoring of ionic models and plotting calcium


michelebucelli:
### Problem

1. In electrophysiology simulations, it is currently not possible to save the calcium transient to the VTU file. Besides being useful in and of itself, this is a useful step towards implementing coupled electromechanics simulations (see also #523).
2. As far as I can see, the currently available ionic models have a very similar interface, with substantial code repetition between the different models, especially for the ODE solution part (the `integ_*` methods of classes `CepModFn`, `CepModAp`, `CepModBo` and `CepModTtp`).

### Solution

I propose introducing a little class hierarchy to centralize the common parts of the ionic models into a generic ionic model interface, and add the calcium extraction (and/or output) directly to that interface.

It might look something like this (function arguments omitted in the interest of brevity):
```cpp
// Abstract ionic model.
class IonicModel {
public:
  // Purely virtual methods to be defined in the concrete ionic model.
  virtual void getf(...) = 0;
  virtual void getj(...) = 0;

  // Methods as in existing ionic model classes:
  void init(...); // this with multiple overloads
  void integ_cn2(...);
  void integ_fe(...);
  void integ_rk(...);
  
  // New method to extract the calcium concentration variable (or its 
  // proxy is for phenomenological models). This can be implemented in
  // a few different ways (e.g. return the index of the calcium variable
  // in the state vector, or extract the corresponding state value out of
  // a local or global state vector).
  virtual void get_calcium(...) = 0;
};

// Concrete ionic model (TTP as an example).
class CepModTtp: public IonicModel {
public:
  // Implement the virtual methods from the base class.
  virtual void getf(...);
  virtual void getj(...);
  virtual void get_calcium(...);
};
```
Concrete ionic models would only have to implement the purely virtual methods from the above class (`getf`, `getj`, `get_calcium`), which are the ones that are truly specific to that ionic model, and correspond to its mathematical definition.

Wherever ionic models are used, they would need to be accessed through a reference to the base class, with the concrete model dynamically instantiated (ideally through an `std::unique_ptr`) depending on the concrete model of choice.

### Additional context

This should make the future coupling to contraction models more modular, allowing to switch between different ionic models with relative ease.

I imagine part of this might be superseded in the future by the work @zasexton is doing on auxiliary ODE/DAE models, but I think it might still be a useful intermediate step towards addressing #523.

Some slightly more advanced features that could be built on top of this are:
[...]

Issue #523, opened 2026-03-30
Implementing Electromechanics Coupling


kko27:
### Problem

svMultiphysics is currently missing coupling between the electrophysiology and mechanics solvers. There is some legacy code (if you look at the cemModelType class) that hints that a coupling scheme was attempted previously. This was re-implemented in C++ in the current codebase without running any tests. 

When we couple electrophysiology to mechanics, we typically need an active tension model. This model represents how the actin and myosin filaments in cardiomyocytes form cross-bridges. This dynamic interaction produces the active tension that drives muscle contraction. The cross-bridge interactions are also modulated by the intracellular calcium concentration, which we obtain from the electrophysiology solver as an output. Schematically, we formalize the coupling approach below: 
<img width="6600" height="5100" alt="Image" src="https://github.com/user-attachments/assets/171de605-2913-49e6-9116-33524e5fee09" />
In the current code base, there are legacy active stress and strain routines which have been defined for each EP ionic model. However, these routines are a little behind literature, as the latest models are more biophysically informed (some examples include Land-Niederer [1] models and Regazzoni-Quarteroni [2] models). This motivates the need for an electromechanics coupled approach. 

### Solution

Several subtasks need to be completed before we can begin to incorporate electromechanics. These are outlined below: 

- Allow users to specify electromechanics coupling physics in the solver.xml input file. This feature is missing and was likely not ported from svFSI. 
- Plot calcium concentrations from EP ionic model to results.vtu files. This ties in with issue [#519](#519) as calcium concentration can be used as a heuristic to determine when to temporally delay the active tension activation. Also, note that calcium concentration will be important for validation studies of EP.  
- Computation of the fiber stretch and stretch rates. Fiber stretch calculations currently exist in the post.cpp routine, but this has not been tested. These are critical parameters that are inputs of the active tension model. 
- Implement cellular ODE models of active tension generation. As an initial example, the Land-Niederer active tension model [1] will be used. 
- Active tension generation coupling and integration with solver. An important note to make here is that two coupling approaches exist for electromechanics: (1) Strong Implicit Coupling and (2) Staggered (Segregated) Approaches. Staggered approaches require stabilization while implicit coupling approaches are stable but are intrusive (requires direct modification of tangent matrix). I plan to experiment with both approaches before deciding on how to implement it for other users. 
[...]

PR #522, opened 2026-03-30
User initialization of EP TTP ionic model states

kko27:
## Current situation
Resolves #407 

## Release Notes 
- Can initialize ionic state variables of TTP EP model from solver.xml file 
- Outputs warning message to user that default EP initial conditions have been selected 

## Testing
New test case in ```tests/cases/cep/spiral_TTP_2d``` that applies 4 different sets of initial conditions to the simulation domain. 

## Code of Conduct & Contributing Guidelines 
- [ ] I agree to follow the [Code of Conduct](https://github.com/SimVascular/.github/blob/main/CODE_OF_CONDUCT.md) and [Contributing Guidelines](https://github.com/SimVascular/.github/blob/main/CONTRIBUTING.md).

PR #540, opened 2026-05-06
Object oriented refactoring of ionic models and plotting calcium

### LINKED ISSUE #535: Sign of stimulus current with Aliev-Panfilov model
### Description

As part of working on #530, I tried running the test `tests/cases/cep/niederer_benchmark_ECGs_quadrature/solver_GMRES_FE_pfib_AP.xml` over multiple timesteps. The results seem to be incorrect.

In the initial timesteps, the excited region's potential becomes lower than the resting value (instead of higher):
Image

As the simulation proceeds, the (negative) excitation does not propagate, and at some point it blows up, likely due to the negative values falling in some unfeasible range for the model (next images are the same frame but with different color scales):
Image
Image

The simulation crashes in time step 42, throwing the error `std::runtime_error: [cep_integ_l] A NaN has been computed during time integration of electrophysiology variables.`

### Reproduction

On the `main` branch, run the test case `tests/cases/cep/niederer_benchmark_ECGs_quadrature/solver_GMRES_FE_pfib_AP.xml` setting `Number_of_time_steps` to 1000 (or any value above 40).

### Expected behavior

The excited region should have higher potential then the resting value, as in this picture (same timestep as first picture above):
Image
The stimulus should propagate over time to the whole domain, and should not blow up.

The above was achieved by changing the sign of the applied current stimulus from negative to positive. Accordingly, I suggest that that sign be changed in the parameter file for this test and possibly other tests involving the AP model.

### Additional context

I don't know if the same issue happens with models other than AP.

I noticed that the sign of the applied current is interpreted differently for different models (at least for TTP and AP, not sure about the others). I think it might be beneficial to make this setting consistent, so that the applied current parameter section can become independent of the ionic model being used. If people agree with this, I can take care of it as part of the refactoring proposed in #530.

### Code of Conduct

michelebucelli:
## Current situation

Addresses the issues discussed in #530 (and fixes #535 as a byproduct).

## Release Notes 

1. Introduced a new abstract class `IonicModel`;
2. Refactored existing ionic models (Aliev-Panfilov, FitzHugh-Nagumo, Bueno-Orovio, ten Tusscher-Panfilov) as classes derived from the abstract class;
4. Decoupled as much as possible concrete ionic models from the rest of the code:
      1. implemented a singleton, self-registering factory to instantiate ionic models;
      2. ionic models manage their own parameters in the respective classes;
      3. the `switch` statements selecting behavior based on the concrete ionic model have been replaced by calls to methods of the abstract class, with behavior selection achieved through polymorphism;
5. Implemented exporting of calcium to VTU from the abstract class.

The user interface is almost identical to before, with the following exceptions:
1. the ionic-model-related parameters in the XML file (e.g. `G_Na` for TTP) are now under the XML section for that ionic model. This is a consequence of (4ii) above. I think it makes sense, because it highlights that those parameters are specific to one ionic model.
2. For similar reasons, the section wrapping the TTP model initial conditions is no longer named `TTP_Initial_conditions` but `TTP` (since it also contains other parameters such as `G_Na`).

The XML files for the test cases have been updated accordingly.

## Documentation

All the new classes and methods have Doxygen documentation. For the most significant ones, I also included some mathematical documentation as part of the Doxygen comments (which is why I modified the `Doxyfile` to support rendering math expressions).

Let me know if further documentation is needed elsewhere.

## Testing

I have run the automatic tests related to electrophysiology, and they pass, except for those involving AP that fail due to the change in applied current sign discussed in #535. Let me know if and how I should update the reference solution for that.

Additionally, I have manually verified with the Niederer benchmark that the new code produces the exact same result as the old code over relatively long time intervals (more than the single timestep of the tests). The mismatch in transmembrane potential is exactly zero when visualized in Paraview.


Code/Source/solver/cep_ion.cpp
```diff
@@ -364,284 +346,41 @@ void cep_integ_l(CepMod& cep_mod, cepModelType& cep, int nX, int nG, Vector<doub
...
+  case TimeIntegratioType::CN2:
+    for (int i = 0; i < nt; i++) {
+      const double t = t1 + i * cep.dt;
+      const double Istim = (t >= Ts - eps && t <= Te + eps) ? cep.Istim.A : 0.0;
+      cep.ionic_model->integ_cn2(cep.imyo, X, Xg, t, cep.dt, Istim, Ksac,
+                                 cep.odes.maxItr, cep.odes.relTol,
+                                 cep.odes.absTol);
+    }
+    break;
+  }
```
michelebucelli: In principle, this step could also be moved to `IonicModel`, while also letting that class manage the parameter that switches between different integration methods.

Code/Source/solver/CepMod.cpp
```diff
@@ -4,6 +4,14 @@
 #include "CepMod.h"
 #include <math.h>
 
+const std::map<ElectrophysiologyModelType, std::string> cep_model_type_to_name{
+    {ElectrophysiologyModelType::NA, "NA"},
+    {ElectrophysiologyModelType::AP, "AP"},
+    {ElectrophysiologyModelType::BO, "BO"},
+    {ElectrophysiologyModelType::FN, "FN"},
+    {ElectrophysiologyModelType::TTP, "TTP"}
+};
```
michelebucelli, line 13: I am not particularly happy about this solution.

Currently, there's two ways of identifying ionic models, the `enum class ElectrophysiologyModelType` and the `std::string` label that is used to register and instantiate them through the factory. I think it would be best to switch to only use one of the two.

The main pro of using `std::string` and dropping `ElectrophysiologyModelType` is that it further decouples ionic models from the rest of the code - there's no need to keep an `enum class` up to date when adding new models.

The main advantage of the `enum class` is the increased type safety and error correction (the compiler spots typos without the need to run a test). However, ultimately there should be no need for the developers to enter the ionic model labels more than once (i.e. in places other than where the model is defined), so I do not think there will be a lot of chances for typos in model names.

As a secondary difference, evaluating conditionals on the `enum class` is faster than with `std::string`, but the difference should not be relevant if those conditionals are not in performance-critical portions of the code (and as far as I can see they currently aren't). The use of polymorphism basically removes the need for this, so this should not be a very relevant point.

In summary, I lean towards `std::string` in the interest of increased implementation locality and decoupling, but I'd be interested in hearing other people's opinion on this.

Code/Source/solver/CepMod.h
```diff
@@ -166,17 +164,8 @@ class cepModelType
...
-    /// @brief Interface for Bueno-Orovio cellular activation model
-    CepModBo bo;
-
-    /// @brief Interface for Fitzhugh-Nagumo cellular activation model
-    CepModFn fn;
-
-    /// @brief Interface for Tusscher-Panfilov cellular activation model
-    CepModTtp ttp;
+    /// @brief Ionic model instance.
+    std::shared_ptr<IonicModel> ionic_model;
```
michelebucelli, line 128: I would be happier for this to be a `unique_ptr`, but the class `cepModelType` needs to be copy-constructible, due to it being a member of `dmnType`, which in turn is stored in an `std::vector` inside `eqType`.

I guess this is acceptable, I don't think the overhead of the `shared_ptr` will be significant.

Code/Source/solver/CepMod.h
```diff
@@ -217,18 +206,6 @@ class CepMod
...
-
-    /// @brief Interface for ABueno-Orovio cellular activation model.
-    CepModBo bo;
-
-    /// @brief Interface for Fitzhugh-Nagumo cellular activation model.
-    CepModFn fn;
-
-    /// @brief Interface for Tusscher-Panfilov cellular activation model.
-    CepModTtp ttp;
-
```
michelebucelli, line 231: I am not sure I fully understand the difference between `CepModelType` and `CepMod`, so I have no strong argument for placing the pointer to the `IonicModel` instance in one class or the other.

I'd be happy to move it if anyone has an opinion on this (and I'd be grateful to anyone who might clarify the difference between the two classes).

Code/Source/solver/consts.h
michelebucelli, line 377: This is only tangentially related to this PR, but: it seems to me that this `enum class` is actually representing two disjoint sets of non-interchangeable options. In other words, it seems to me that a value of the group `outGrp_*` would not be valid in a place where a value in the group `out_*` is expected, and vice versa. Is that correct?

If it is, do you think it would make sense to split this into two distinct enumerations (say, `OutputGroup` and `OutputVariable`)?

Related to this, do the numeric values attributed to the members of this enumeration matter?

Code/Source/solver/ionic_bueno_orovio.cpp
```diff
@@ -0,0 +1,194 @@
...
+                       const Vector<double> &Xg, Vector<double> &f,
+                       const double I_stim, const double I_sac) const {
+  // Create local copies of the state variables
+  const double u = X(0);
+  const double v = X(1);
+  const double w = X(2);
+  const double s = X(3);
+
+  // @todo[michelebucelli] Why minus 1? In getj this seems not to be the case.
+  const int i = zone_id - 1;
```
michelebucelli: I imported this statement verbatim from `CepModBo::getf`. However, it seems to conflict with the evaluation of `zone_id` that is done in `getj` (which is taken from `CepModBo::getj`). I think that only one of the two is the correct one.

Based on the definition of [`read_cep_domain`](https://github.com/michelebucelli/svMultiPhysics/blob/fc85f1df3a785e91b3a0a75d8a575ef267e76951/Code/Source/solver/read_files.cpp#L1058-L1070), this index is 1-based, so I'm guessing this implementation is correct, and the one in `getj` is incorrect. Am I getting this right?

If so, is there any specific reason for using a 1-based index instead of the more standard 0-base index?

Code/Source/solver/ionic_model.h
```diff
@@ -0,0 +1,465 @@
...
+   *   endocardium, myocardium).
+   * @param[in] X Vector of state variables.
+   * @param[in] Xg Vector of gating variables.
+   * @param[out] f Right-hand side of the model equations.
+   * @param[in] I_stim Applied current.
+   * @param[in] I_sac Stretch-activated current.
+   */
+  virtual void getf(const unsigned int zone_id, const Vector<double> &X,
+                    const Vector<double> &Xg, Vector<double> &f,
+                    const double I_stim, const double I_sac) const = 0;
```
michelebucelli: What would you think of changing this function so that it returns the `Vector<double>`, instead of taking it as an output argument?

The argument in favor is that it enhances readability and it is more aligned with modern C++ idioms (see also [here](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#rf-out)).

The argument against is that it might entail an extra allocation, although I'm pretty sure that copy elision/return value optimization would work smoothly in this setting (or maybe we'd need a move constructor for `Vector`, which would not be a bad thing in and of itself).

Similar considerations apply to `getj` below.

Code/Source/solver/ionic_model.h
```diff
@@ -0,0 +1,465 @@
...
+   *
+   * @{
+   */
+
+  /// Voltage scaling [mV].
+  const double Vscale;
+
+  /// Time scaling.
+  /// @todo[michelebucelli] Document unit of measure.
+  const double Tscale;
```
michelebucelli: I wasn't able to figure out with certainty whether the standard temporal unit of measure is seconds or milliseconds. Could someone clarify this (and/or point me to the place where this kind of information is documented)?

Code/Source/solver/ionic_model.h
```diff
@@ -0,0 +1,465 @@
...
+ * [singleton](https://en.wikipedia.org/wiki/Singleton_pattern) patterns. There
+ * should always exist only one instance of this class, which cannot be accessed
+ * directly but only manipulated through the static methods of this class.
+ *
+ * To register a new ionic model into the factory, you can call the
+ * register_model static method, passing a class derived from Ionic as template
+ * argument and a label for the model as argument. A shortcut for this is to
+ * use the macro REGISTER_IONIC_MODEL.
+ */
+class IonicModelFactory {
```
michelebucelli, line 468: @kko27 This singleton/factory with self registration pattern can definitely come in handy for modularizing contraction models as well (and it might have a few other use cases). If we want to go down that path, it might be worth to turn this into a templated class, so that it is no longer ionic model-specific but can be directly used for different class hierarchies.

Code/Source/solver/ionic_model.h
michelebucelli, line 1: What do you think about grouping ionic model source files into a dedicated folder under `solver` (or elsewhere)?

Also, what do you think about having a `namespace` for ionic models?

Code/Source/solver/Parameters.h
```diff
@@ -1181,79 +1182,121 @@ class FiberReinforcementStressParameters : public ParameterLists
...
 };
 
-/// @brief Stores parameters for the 'Gating_variables' XML element
-/// under TTP_initial_conditions.
-class TTPGatingVariablesParameters : public ParameterLists
-{
-  public:
-    TTPGatingVariablesParameters();
+/// @brief Generic ionic model initial conditions parameters.
+class IonicInitialStateParameters : public ParameterLists {
```
michelebucelli, line 1200: I am not particularly happy about this class and `IonicModelParameters` being still in this file, as opposed to being defined in `ionic_model.h` and `ionic_model.cpp`.

However, I could not find an easy way around this, due to circular dependency problems. In particular:
1. `DomainParameters`, defined here, must come after the definitions of `IonicInitialStateParameters` and `IonicModelParameters`;
2. these in turn must come after the definitions of `ParameterLists` and `Parameter`, ...
3. ...so that the new classes must be defined between the general-purpose parameter management classes and the concrete class `DomainParameters`;
4. but the file `Parameters.h` must be included into `ionic_model.h`, to have access to some of the general purpose parameter management classes.

In short, I tried for a while to avoid this, also adding some forward declarations, but I gave up after a while. I think it that one possible solution, quite invasive, would be to split the general purpose parameter classes to their own header file, but honestly I haven't given a lot of thought to this, so I'm not sure.

If anyone has any advice about this point I'd be happy to take it.

michelebucelli, line 1200: Also related to these classes, I think I might have re-implemented some of the base functions of `ParameterLists` (e.g. `print_parameters` seems to repeat the logic of `print_parameter_list`, and the `xml_element_name` member appears to be duplicated). However, I saw that this is happening elsewhere too (e.g. in `BodyForceParameters` and others), so I repeated that pretty much blindly.

I'd be grateful if someone could clarify this, and I'd be interested in discussing if and how to remove this possible overlap.

Code/Source/solver/Parameters.h
```diff
@@ -1285,7 +1328,9 @@ class DomainParameters : public ParameterLists
     StimulusParameters stimulus;
     FluidViscosityParameters fluid_viscosity;
     SolidViscosityParameters solid_viscosity;
-    TTPInitialConditionsParameters ttp_initial_conditions;
+
+    // Ionic model parameters.
+    std::map<std::string, std::unique_ptr<IonicModelParameters>> ionic_models;
```
michelebucelli, line 1352: Related to the initial point about `std::string` vs `enum class`, this should be made uniform to the rest (i.e. either use the `enum class` or the `std::string`, depending on what we settle on).

Code/Source/solver/read_files.cpp
```diff
@@ -1501,6 +1474,33 @@ void read_eq(Simulation* simulation, EquationParameters* eq_params, eqType& lEq)
...
+            out.name = var.first;
+            out.options.spatial = true;
+            lEq.output.push_back(out);
+            lEq.nOutput++;
+            registered_vars.insert(var.first);
+          }
+        }
+      }
+    }
+  }
```
michelebucelli: What would you think about moving some of this logic inside `IonicModel`?

Code/Source/solver/ionic_model.h
kko27: The standard unit of measure is milliseconds for electrophysiology ionic models. You can find it under this documentation page: https://simvascular.github.io/documentation/multi_physics.html (under the Cellular Activation Models section)

Code/Source/solver/cep_ion.cpp
```diff
@@ -364,284 +346,41 @@ void cep_integ_l(CepMod& cep_mod, cepModelType& cep, int nX, int nG, Vector<doub
...
-      }
-    } break; 
-  } 
+  if (cep.ionic_model == nullptr) {
+    svmp::raise<svmp::FE::NotInitializedException>(
+        SVMP_HERE, "ionic model was not constructed.");
+  }
+
+  switch (cep.odes.tIntType) {
+  case TimeIntegratioType::FE:
```
kko27: Minor comment:  I noticed that in the old code we called this `TimeIntegratioType`. Maybe we should correct this class to `TimeIntegrationType`.

Code/Source/solver/cep_ion.cpp
```diff
@@ -364,284 +346,41 @@ void cep_integ_l(CepMod& cep_mod, cepModelType& cep, int nX, int nG, Vector<doub
...
-            if (t >= Ts-eps &&  t <= Te+eps) {
-              Istim = cep.Istim.A;
-            } else {
-              Istim = 0.0;
-            }
-
-            cep.ttp.integ_cn2(cep.imyo, nX, nG, X, Xg, t, cep.dt, Istim, Ksac, IPAR, RPAR);
-
-            // Electromechanics excitation-activation
-            if (cem.aStress) {
```
kko27, line 630: Just curious about what you did with these electromechanics calls in your new changes

Code/Source/solver/CepMod.h
kko27: The way I understand this is that `cepModelType` comes from the `dmnType` class and this is used to define parameters on a per-domain basis. `CepMod` derives from the `Simulation` class and stores data that is global across all domains.

But, this looks like a redundancy to me as the `ap`, `bo`, `fn`, and `ttp` members of `CepMod` are never used in the current codebase. 

In terms of practicality, I'm imagining there will be situations where different simulation domains will have different ionic models so I'm in favor of placing the pointer to the `cepModelType` class.

Code/Source/solver/ionic_model.h
michelebucelli: Thanks @kko27! I have edited the documentation comment. Thank you also for pointing out the right section of the docs!

Code/Source/solver/ionic_bueno_orovio.cpp
kko27: Yes, I think you are correct here. We are using 1-index because of how we have defined the myocardium zone ids in `read_cep_domain`. `getj` then contains a bug as it is assuming 0-index. 

I don't mind us moving to the standard 0-base index because users don't know about the zone id indices. They just specify the strings in the `.xml` file.

Code/Source/solver/ionic_model.h
kko27: I like the idea of having a folder for the ionic model files. @zasexton might have some thoughts on this. 

What would the motivation be to have a `namespace`?

Code/Source/solver/ionic_model.h
kko27: I'm not too familiar with this, but happy to look into it. I do see the value in making this as generalizable as possible

Code/Source/solver/ionic_model.h
```diff
@@ -0,0 +1,464 @@
...
+   * different units than a molar concentration).
+   *
+   * @return A vector of pairs {variable_name, state_vector_index}. The vector
+   * need not include the transmembrane potential V, which is exported by other
+   * means.
+   */
+  virtual std::vector<std::pair<std::string, int>>
+  get_output_variables() const {
+    return {{"Calcium", get_calcium_index()}};
+  }
```
kko27: Does this mean that users won't need to specify that they want Calcium outputs from the `.xml` file?

kko27 (commented):
I've gone through most of the code but I have left some of your comments unanswered as I wasn't too sure. 

**Comment about tests:** 
For the AP ionic model test, it probably makes sense to replace the result_001.vtu file with your latest simulation result. 

I notice that the niederer_benchmark_ECGs_quadrature test is also failing. Would you know the reason behind that?

Code/Source/solver/CepMod.h
michelebucelli: I see, thank you for the explanation! I will leave the model pointer only in `cepModelType` then.

Code/Source/solver/cep_ion.cpp
michelebucelli: For now they are removed.

I figured we would bring them back as part of #523, but most likely in their own class hierarchy, so we would need to reimplement them elsewhere, and there was no need to keep them here.

Before removing them, I also double-checked that they are not currently accessible to the user.

Let me know if you think they should remain in this version of the code.

Code/Source/solver/ionic_model.h
michelebucelli: True, in the new version calcium will always be exported. Now that you mention it, it might be worth it to add an option in the XML file for that. The easy thing would be to have an option for all ionic variables (say, `Ionic_states`), rather than an individual option for each variable that individual models may have.

Code/Source/solver/ionic_bueno_orovio.cpp
michelebucelli: Alright, then for the time being I'll use the correct index in `getj`, we can discuss whether to use 1- or 0-based indices at a later time I think (especially since, as you say, this is transparent to the user).

Code/Source/solver/ionic_model.h
michelebucelli: The generic factory class might end up being a bit abstract, but the neat thing is that it allows you to add a new model to the list of registered ones as simply as writing `REGISTER_IONIC_MODEL("ModelName", ModelClass)` in the same `.cpp` file where the model is implemented. We could have a similar `REGISTER_CONTRACTION_MODEL` macro, or something general-purpose like `REGISTER_INTO_FACTORY(IonicFactory, "ModelName", ModelClass)`.

Then, whenever you need to do something for all registered models, you can use the `visit` method (and/or implement something more specific if needed).

I took inspiration from [what we implemented in `lifex`](https://gitlab.com/lifex/lifex-public/-/blob/main/source/core/generic_factory.hpp?ref_type=heads), in turn inspired by [this tutorial](https://www.cppstories.com/2018/02/factory-selfregister/) which might be a good starting point to understand how this works.

Code/Source/solver/ionic_model.h
michelebucelli: In my view, having a `namespace` would have the same purpose as having a dedicated folder, except for class rather than file organization.

The implication would be that classes would be grouped under a common name, say `ionic`, and would be accessed by scoping with that name (e.g. `ionic::AlievPanfilov`, `ionic::IonicModel` - this might actually be turned into `ionic::Model` for brevity - and so on).

This would make the meaning of the class names a bit more explicit (e.g. to a new user it would be more evident that `BuenoOrovio` is a ionic model, even if they don't know the specific model). This information is already conveyed in a few other ways (the fact that the class inherits from `IonicModel`, and possibly that the corresponding file is inside a ionic model folder), but "self-documentation redundancy" is not necessarily bad.

Still, I don't think namespacing is strictly necessary to be honest (I think grouping the files in a folder might be more useful).

Code/Source/solver/cep_ion.cpp
michelebucelli: Good catch! I actually meant to fix that, but then I forgot 😄 Will fix.

michelebucelli:
Thanks for the review @kko27!

About the tests: also the failing tests from the `niederer_benchmark_ECGs_quadrature` suite are using the AP model, and they fail because of the changed sign in the stimulus current (#535). Just to make sure that that was the cause, I have tried putting that sign back like it was and running the tests again, and all of them pass.

I'll go ahead and update the reference solutions for all those tests then.

Code/Source/solver/ionic_bueno_orovio.cpp
zasexton: For now I would continue to use the 1-based index as the code already assumes. We will need to have a discussion about this in the future about the code being 1 or 0 indexed and clearly stated everywhere.

Code/Source/solver/Parameters.cpp
```diff
@@ -1588,6 +1588,141 @@ void SolidViscosityParameters::set_values(tinyxml2::XMLElement* xml_elem)
...
+    throw std::runtime_error(xml_element_name + " requires an '" +
+                             initial_X_parameters.xml_element_name +
+                             "' XML section.");
+  }
+
+  if (!initial_Xg_parameters.defined()) {
+    throw std::runtime_error(xml_element_name + " requires an '" +
+                             initial_Xg_parameters.xml_element_name +
+                             "' XML section.");
+  }
```
zasexton: Are initial values and gating variables necessary for every ionic model? Also should we intended for behavior that model parameter sections can partially override defaults? 

If so, models with no gating variables treat that section as unsatisfied I believe and might result in throws for otherwise valid parameter overrides. Also should scalar-only overrides be allowed without requiring restatement of all initial conditions? For example, overriding a gating variable without repopulating the initial conditions:

```xml
<TTP>
  <G_Na>...</G_Na>
</TTP>
```

michelebucelli: As I understand it, the current version of this is that the ionic model section may be omitted, but if it is included then the sections `Initial_states` and `Gating_variables` must be present, or an exception is raised.

However, I tested this just now and noticed that omitting individual parameters within the sections (or conductance parameters such as `G_Na`) does not raise an exception, despite the `required` flag for those parameters having been set to true. Perhaps I am missing some call to actually check the required parameters?

Having said all this, my personal preference would be to have everything fully optional, so that the user is free to only modify those parameters or initial states that are relevant to them, with the omitted parameter implicitly taking their default value. However, based on [this comment on a previous PR](https://github.com/SimVascular/svMultiPhysics/pull/522#discussion_r3012612002), it seemed to me that there is a preference towards requiring the user to always declare parameters within a section, which I tried to follow here as well.

Code/Source/solver/Parameters.cpp
```diff
@@ -1588,6 +1588,141 @@ void SolidViscosityParameters::set_values(tinyxml2::XMLElement* xml_elem)
...
+//                 IonicModelParameters                 //
+//////////////////////////////////////////////////////////
+
+IonicModelParameters::IonicModelParameters(
+    const std::string &xml_element_name_,
+    const std::vector<std::pair<std::string, double>> &initial_X,
+    const std::vector<std::pair<std::string, double>> &initial_Xg)
+    : xml_element_name(xml_element_name_),
+      initial_X_parameters("Initial_states", initial_X),
+      initial_Xg_parameters("Gating_variables", initial_Xg) {
```
michelebucelli: @kko27 I took this name from your previous PR (#522). However, it seems to me that `Initial_gating_variables` would be a bit more clear. Would you agree with changing it?

In a similar vein, I would suggest renaming the output field "Action potential" to "Transmembrane potential" (or the slightly shorter "Membrane potential"), which seems to me a bit more accurate.

Code/Source/solver/Parameters.cpp
kko27: Regarding the previous PR comment, @ktbolt and I were thinking that if a user wants to specify a parameter, they need to specify all the relevant parameters. We wanted to avoid situations where users would use and share `.xml` files without realizing that there were implicit overrides. 

Ultimately, all of the ionic model parameters (not just the initial conditions) should be specified in some input file. This is because some of the model parameters have specific units associated with them, and we want the solver to be dimensionless. That being said, I recently noticed that we do have scaling factors the V and time, which I hadn't seen before...

Code/Source/solver/cep_ion.cpp
kko27: I think removing them is fine. I'll just keep a copy of the current codebase as a reference when implementing #523

Code/Source/solver/Parameters.cpp
kko27: Yes, let's change it to `Initial_gating_variables` so that it is more clear. That was my mistake. 

I'm okay with renaming the output field to "Membrane potential" as action potential refers to the depolarization of the membrane potential (i.e., it is a specific instance of membrane potential). This will probably affect people with post-processing scripts that use the old terminology, but this change should be quick to fix. Tagging @javijv4 for additional thoughts

Code/Source/solver/Parameters.cpp
michelebucelli: > We wanted to avoid situations where users would use and share .xml files without realizing that there were implicit overrides.

Yeah, I can see this being a potentially annoying issue.

The way we avoided this in lifex was to have every simulation dump the parameters that were used (regardless of whether they were explicitly set or their default value was implicitly assumed), in the same format that the solver accepted in input. This way, users wanting to replicate someone else's simulation would be able to use this dump file and be sure to have the exact same setting.

But we did encounter some problems even with this, e.g. due to changes between versions or some of the dumped parameters actually not being relevant to that particular setting (or, more simply, users sharing the file with the default values omitted instead of the parameter dump). Also, I imagine this kind of solution would require a quite substantial overhaul of the current parameter management.

> Ultimately, all of the ionic model parameters (not just the initial conditions) should be specified in some input file.

I'm not sure I understand this part of your comment. With this PR, all the ionic model parameters can be specified through the XML file (either in the main file or imported through an `<Include>` tag). Do you mean to say that the whole set of ionic model parameters should be made strictly required, i.e. that the user not only can but must provide all the coefficients?

> ...and we want the solver to be dimensionless. That being said, I recently noticed that we do have scaling factors the V and time, which I hadn't seen before...

Yeah I think that the output of the electrophysiology solver is dimensional, with the potential expressed in mV, and the ionic state variables expressed in whatever the ionic model assumes. It also seems to me that the monodomain solver itself uses a dimensional formulation, but I'm not 100% sure.

kko27: Yes, I meant to say that the whole set of the ionic model parameters should be strictly required.

Code/Source/solver/Parameters.cpp
javijv4: Yes, I agree with the name change (both initial_gating_variables and membrane potential). I think the clearer the names, the better!

Code/Source/solver/read_files.cpp
javijv4: Yes, I think this code can be encapsulated. My first thought was to move it inside the `read_outputs` function to avoid having an additional output-related code after the `read_outputs` call. Maybe within `read_outputs` the `IonicModel` logic can be called?

Code/Source/solver/read_files.cpp
```diff
@@ -1501,6 +1475,43 @@ void read_eq(Simulation* simulation, EquationParameters* eq_params, eqType& lEq)
   // Read VTK files or boundaries. [TODO:DaveP] this is not a correct comment.
```
javijv4: This isn't part of the PR purpose, but maybe it is a good opportunity to fix this comment?

Code/Source/solver/Parameters.cpp
michelebucelli: Alright, thanks! I'll go ahead and change `Gating_variables` to `Initial_gating_variables`.

Do you know if there is someone else who could be affected by the `Action potential` name change, and should be asked for comment about this?

Code/Source/solver/cep_ion.cpp
```diff
@@ -57,12 +58,11 @@ void cep_init(Simulation* simulation)
           }
           int nX = eq.dmn[iDmn].cep.nX;
           int nG = eq.dmn[iDmn].cep.nG;
-          int imyo = eq.dmn[iDmn].cep.imyo;
 
           Vector<double> Xl(nX); 
           Vector<double> Xgl(nG);
 
-          cep_init_l(eq.dmn[iDmn].cep, nX, nG, Xl, Xgl);
+          cep_init_l(eq.dmn[iDmn].cep, Xl, Xgl);
```
javijv4: This function is now only one line; maybe it is worth removing the function and calling `cep.ionic_model->init` directly for clarity?

Code/Source/solver/Parameters.cpp
javijv4: I think the only people constantly using EP right now have been tagged. Maybe @zinanhu0810? @kko27 can tell if I am missing someone. 
I am not sure if this is a good idea, but maybe it is worth adding a specific throw when the user calls `Action potential` to let them know about the name change?

Code/Source/solver/Parameters.cpp
javijv4: @kko27, that is the case for TTP but not for the other models, right? In the Aliev-Panfilov tests, no parameters are defined. Also, even for TTP, only a subset of parameters is currently specified in the `ttp_*_parameters.xml` file, while the others have a default, right? 

I think we should agree on a certain way of defining the parameters, either all optional or all required, and be consistent throughout all the ionic models. 

@michelebucelli, I was looking at the issue mentioned by @zasexton, and [this](https://github.com/user-attachments/files/27718289/solver_missing_gating_variables.xml) solver file crashes. It is the same as in `tests/cases/cep/square_AP_2d/solver.xml`, but I added the `<AP>` section to modify specific parameters. Since the AP model does not have gating variables, I am unable to really use the `<AP>` because the `<Gating_variables>` is required. I think this case shouldn't fail.

michelebucelli: > I think this case shouldn't fail.

I agree, and I'll push a fix shortly. I have changed it so that the existence of the `Initial_states` or `Initial_gating_variables` section is only required if the corresponding state vector, passed in when constructing `IonicModelParameters`, is not empty. This way, models that have no gating variables will not require the corresponding section (and hypothetical models without states will not require the `Initial_states` section, although I'm not sure if that's ever going to happen).

I'll also point out that these are only required if the general ionic model section is included. That is: if there exists a `<TTP>` element, than that element must include `<Initial_states>` and `<Initial_gating_variables>`. If however the `<TTP>` element is entirely omitted, then default values are used and no exception is thrown. This is why, even though all models now can be configured in the XML files, there's no need to set the parameters for AP in the AP tests, for example.

Do you think that we should also check that the general section is defined for the concrete ionic model that is being used?

alisonmarsden:
Perhaps also check with @vvedula22 to get his input on all of this.

Code/Source/solver/cep_ion.cpp
michelebucelli: Good point! Done [here](https://github.com/michelebucelli/svMultiPhysics/commit/835d369cbf6a9ef74d5214df901c008c3a6857f6).

Should I also update `Code/Source/solver/README.md` accordingly? I've only noticed this file now, and there's a few points (including this about `cep_init_l`) that would be made outdated by this PR.

javijv4: Yeah, that's probably a good idea.

Code/Source/solver/Parameters.h
```diff
@@ -16,6 +16,8 @@
 #include <variant>
 #include <vector>
```
zasexton, line 19: should probably add `#include <memory>` for the `std::unique_ptr` and `std::make_unique` used later.

Code/Source/solver/Parameters.cpp
```diff
@@ -1588,6 +1588,147 @@ void SolidViscosityParameters::set_values(tinyxml2::XMLElement* xml_elem)
...
+  }
+
+  if (initial_Xg_parameters.required && !initial_Xg_parameters.defined()) {
+    svmp::raise<svmp::ParseException>(
+        SVMP_HERE, xml_element_name + " requires an '" +
+                       initial_Xg_parameters.xml_element_name +
+                       "' XML section.");
+  }
+
+  check_required();
```
zasexton, line 1730: for `VectorParameter<double> values` the method `VectorParameter::set(const std::string&)`does not mark `value_set_ = true` i don't think. Furthermore, parsed XML values to the default vector look like they would be appended to the default vector. Do we have a check already that looks to change default values and interrogates if they are properly used? as I see it though this is more of a generic problem with the requirement checking and parsing more than the refactoring of the ionic models.

Code/Source/solver/Parameters.h
michelebucelli: Good catch, I'll add it!

In the past I've used `cpplint` to verify this more consistently - it might be a useful addition to the CI/CD pipelines.

Code/Source/solver/Parameters.cpp
michelebucelli: > for VectorParameter<double> values the method VectorParameter::set(const std::string&)does not mark value_set_ = true i don't think.

I think you're right! Would you like me to try and address this?

> Furthermore, parsed XML values to the default vector look like they would be appended to the default vector. Do we have a check already that looks to change default values and interrogates if they are properly used?

Here I'm not sure what you mean though, could you try to clarify?

zasexton: if it isn't too much additional work I think it would be great to make sure that the required parameters produce the predicted `value_set_` marker. If too much work, then we can make this an immediate follow-on issue/PR

For clarification: I think when non-default vector values are given they will be appended instead of replacing the default values in the parameter vector. When i was reviewing the requirement check this is what the `set` method for the `VectorParameters` class looks like

https://github.com/SimVascular/svMultiPhysics/pull/540/changes#diff-65a3923e9454f028aa0e94a0a481029ae1ffdf56767d12b3210870db7fd6f8f0L221

Code/Source/solver/read_files.cpp
michelebucelli: Ah right, my bad for not putting it into `read_outputs` in the first place. I'll move this there.

Having said that, I looked a bit more into encapsulating this into `IonicModel`. This would look something like
```cpp
lEq.dmn[iDmn].cep.ionic_model->register_outputs(lEq);
```
However, this won't work because of circular dependencies (`ionic_model.h` would need to include `ComMod.h` which includes `CepMod.h` which includes `ionic_model.h`). The same issue arises when passing `lEq.output` to `register_outputs`, instead of `lEq`, because also `outputType` is defined in `ComMod.h`.

I think this is the same kind of problem I was having with moving the ionic parameter classes away from `Parameters.h` (see [here](https://github.com/SimVascular/svMultiPhysics/pull/540#discussion_r3197568472)): some of these general purpose files both define the "abstract" functionality (like `outputType`) and contain the concrete logic that uses them for specifical applications. Abstracting that logic away would require injecting the code inbetween those two (abstract and concrete), which would not be a good use of `#include` directives. (Or splitting the general file into more files, I guess, but I feel it would be way beyond the scope of this PR).

(This prompts another, somewhat unrelated, question: isn't `lEq.nOutput` redundant  with respect to `lEq.output.size()`?)

Code/Source/solver/Parameters.cpp
zasexton: here are the lines of code:

https://github.com/SimVascular/svMultiPhysics/blob/69086cb13174152673c14c86fba49ae00f4a4a11/Code/Source/solver/Parameters.h#L221-L245

Code/Source/solver/read_files.cpp
javijv4: Hmm.. I don't love this `lEq.dmn[iDmn].cep.ionic_model->register_outputs(lEq);` because it means passing `lEq,` which handles all domains, to the ionic model of a domain. 

I do see that this block is much more involved than the rest of the code in read_outputs, but if the solution is too complicated, I think it is okay to keep it in `read_files.cpp` (maybe within its own function called from `read_outputs`?) until a larger refactoring deals with the `Parameters.h` file.

michelebucelli: Yeah I see your point. Perhaps a better solution would be to have something like
```cpp
auto ionic_outputs = lEq.dmn[iDmn].cep.ionic_model->get_outputs();
lEq.output.insert(lEq.output.end(), ionic_outputs.begin(), ionic_outputs.end());
lEq.nOutput += ionic_outputs.size();
```
but this still faces the circular dependency issue related to the definition of `outputType`, which needs to be returned by `IonicModel::get_outputs`.

Maybe I can work around that with a forward declaration though.

michelebucelli: Done, let me know what you think!

Code/Source/solver/read_files.cpp
```diff
@@ -2383,6 +2358,35 @@ void read_outputs(Simulation* simulation, EquationParameters* eq_params, eqType&
...
+      }
+
+    if (output_ionic_vars) {
+      for (const auto &dmn : lEq.dmn) {
+        if (dmn.phys != consts::EquationType::phys_CEP)
+          continue;
+
+        svmp::check_not_null<svmp::FE::NotInitializedException>(
+            dmn.cep.ionic_model.get(), SVMP_HERE,
+            "ionic model was not constructed.");
```
michelebucelli: @zasexton Do you think we could change this function template so that it also works for smart pointers out of the box?

One possible solution to this could be:
```cpp
template<class ExceptionT, class PointerT, class... Args>
void check_not_null(PointerT ptr, SourceLocation location, Args&&... args)
{
    if (ptr == nullptr)
        raise<ExceptionT>(location, std::forward<Args>(args)...);
}
```
the main differences being
- that `PointerT` is no longer necessarily a raw-pointer type, and...
- ...that the function is now `void` instead of passing the pointer through, to comply with the behavior of `std::unique_ptr` (not sure if this is a crucial feature, but based on current uses of the function it seems fine to me).

I think this should be already type-safe enough, i.e. the comparison `ptr == nullptr` will raise a compiler error for types that are not pointer (including integer types). If you want to more strictly enforce pointer-ness of `PointerT`, this should be doable through static asserts type traits.

Code/Source/solver/Parameters.cpp
michelebucelli: I renamed to `Membrane_potential`.

I didn't add the assertion though, because it seemed a bit ad-hoc, and also would not cover the post-processing scenarios but only outdated XML files (which should already raise an exception with an unknown element message).

Code/Source/solver/Parameters.cpp
michelebucelli: It should be fixed. I tested the behavior with the BO ionic model, and added explicitly the BO parameters to one of the tests so that this logic gets minimally tested as part of it.

I still find the management of required/set parameters a bit confusing though, but I think discussing/addressing that would be out of scope for this PR.

Code/Source/solver/read_files.cpp
michelebucelli: Good point, I changed the comment.

Code/Source/solver/Parameters.cpp
michelebucelli: I had inadvertently broken vector-valued parameters, but it should be fixed now. 

To the best of my understanding, the logic is more or less the following:
1. `VectorParameter` is supposed to be able to parse XML parameters in two formats:
    ```xml
    <Param1> 1, 2, 3 </Param1>

    <Param2> 1 </Param2>
    <Param2> 2 </Param2>
    <Param2> 3 </Param2>
    ```
    Both `Param1` and `Param2` should be interpreted as the vector `{1, 2, 3}`.

2. Accordingly, I made it so that the default values, provided when calling `VectorParameter::set(const std::string&, bool, const std::vector<T>)`, are cleared the first time that `VectorParameter::set(const std::string&)` is called, and subsequent calls only append to whatever previous calls had stored.

Code/Source/solver/cep_ion.cpp
michelebucelli: I opted for moving part of this to a new method `IonicModel::integ`, while retaining the loop over sub-timesteps in `cep_integ_l`. Notice that this required moving the definitions of `TimeIntegrationType` and `odeType` to `ionic_model.h`, to avoid circular dependency issues between `CepMod.h` and `ionic_model.h`.

The ODE solver parameters are still managed globally.

Code/Source/solver/read_files.cpp
zasexton: Yes, I think this change is completely fine. the `check_not_null` was very much a helper utility that could be improved.

Code/Source/solver/CepMod.cpp
zasexton: Will mark this as resolved for now since `std::string` is completely fine. There can be another polishing pass in a later PR where we clean up string vs enum based selections.

Code/Source/solver/consts.h
zasexton: Let's make a new issue for this so it can be a separate and more focused PR

Code/Source/solver/ionic_model.h
zasexton: For now let's merge the PR with the ionic models loose in the `solver/` folder. I agree that they should eventually get their own folder but for now I don't think this matters much. The file naming scheme already greatly improves a new developer/user's ability to suss out which files are critical for the ionic models.

zasexton (approved):
As the code currently stands, I believe it is fine to merge this PR. there are related sub-issues that we can log for further organization and polishing but I don't believe they preclude us from pulling in a lot of this fantastic refactoring work for the electrophysiology

Code/Source/solver/Parameters.cpp
javijv4: No, I think it's fine as it is. I do think we need to decide whether the input parameters will be optional or mandatory, and ensure consistency across models. But, for the purposes of this PR, I think this is good.

javijv4 (approved):
I think this is good to merge too! The code is much easier to follow now, so any further changes will be easier to address in future issues.

zasexton:
once the new changes from the latest PR merge are pulled into this branch (and the relevant tests pass) @michelebucelli feel free to merge this PR if you are happy with it

michelebucelli:
Thank you @zasexton! I'll open follow-up issues as you suggested for the threads that are left open, and then I'll go ahead and merge this.

PR #543, opened 2026-05-13
Updated RIS valve upstream direction and mean pressure integration domain, added an optional argument to invert RIS surface normal

### LINKED ISSUE #542: Swap RIS valve upstream/downstream, cap pressure integration domain at half thickness, optional invert surface normal
### Problem

* The current logic distinguishes the upstream and downstream sides of the RIS valve using vertical coordinates.
* When computing mean pressure upstream and downstream of the valve, the current integration domain extends from the valve surface to an upper SDF limit.
* The current code defines the valve surface unit normal from the surface mesh and assumes that the normal points outward from the valve. However, some valve meshes have the opposite normal orientation.

### Solution

* The inflow side of the valve should be treated as the upstream.
* For more accurate mean pressure integration, the domain should begin at a distance of half-valve thickness from the valve surface.
* An optional argument should be added to invert the valve surface normal when needed.

### Additional context

_No response_

### Code of Conduct

hanzhao2020:
Resolves https://github.com/SimVascular/svMultiPhysics/issues/542

## Release Notes 
* Swapped the RIS valve upstream and downstream direction
* Updated mean pressure integration domain for the RIS valve until half-thickness
* Added an optional argument to invert the RIS valve surface normal direction

### DIFF
--- a/Code/Source/solver/ComMod.h
+++ b/Code/Source/solver/ComMod.h
@@ -1482,6 +1482,9 @@ class urisType
     // if the fully closed position alone is not able to prevent backflow.
     double sdf_deps_close;
 
+    // Whether to invert the valve surface normal vector.
+    bool invert_normal;
+
     // Opening positions of the valve surfaces.
     Array3<double> DxOpen;
 
--- a/Code/Source/solver/ComMod.h
+++ b/Code/Source/solver/ComMod.h
@@ -1483,10 +1483,11 @@ class urisType
     double sdf_deps_close;
 
     // Whether to invert the valve surface normal vector. Default is false.
-    // If true, inverts the shell face normal used for SDF sign calculation so the negative
-    // and positive SDF sides remain consistent with the upstream and downstream directions.
-    // Mesh face normals are assumed to point outward from the leaflet. Set to true if the
-    // surface mesh has inward-pointing normals to preserve the SDF sign convention.
+    // 
+    // Valve normal vectors are assumed to point downstream, so that the
+    // downstream region has positive signed distance and the upstream region
+    // has negative signed distance. If the input surface does not satisfy
+    // this assumption, this flag should be set to true to flip the normals.
     bool invert_normal;
 
     // Opening positions of the valve surfaces.
--- a/Code/Source/solver/ComMod.h
+++ b/Code/Source/solver/ComMod.h
@@ -1482,7 +1482,11 @@ class urisType
     // if the fully closed position alone is not able to prevent backflow.
     double sdf_deps_close;
 
-    // Whether to invert the valve surface normal vector.
+    // Whether to invert the valve surface normal vector. Default is false.
+    // If true, inverts the shell face normal used for SDF sign calculation so the negative
+    // and positive SDF sides remain consistent with the upstream and downstream directions.
+    // Mesh face normals are assumed to point outward from the leaflet. Set to true if the
+    // surface mesh has inward-pointing normals to preserve the SDF sign convention.
     bool invert_normal;
 
     // Opening positions of the valve surfaces.
--- a/Code/Source/solver/uris.cpp
+++ b/Code/Source/solver/uris.cpp
@@ -39,32 +39,30 @@ void uris_meanp(ComMod& com_mod, CmMod& cm_mod, const int iUris, const SolutionS
   auto& uris_obj = uris[iUris];
 
   const int nsd = com_mod.nsd;
-  // const int cEq = com_mod.cEq;
 
-  // auto& An = com_mod.An;
-  // auto& Ad = com_mod.Ad;
-  // auto& Dn = com_mod.Dn;
+  // Compute the mean pressure in the upstream and downstream regions of the fluid mesh 
 
-  // Let's conpute the mean pressure in the two regions of the fluid mesh 
-  // For the moment let's define a flag IdSubDmn(size the number of elements)
-
-  // Now we can compute the pressure mean on each subdomain
-  // We need to have a sdf array for each mesh
-  double Deps = uris_obj.sdf_deps * 2.5;
+  // Set the limit of the upstream and downstream regions to 5 times the closed 
+  // valve thickness. This should give a reasonable range for the upstream and 
+  // downstream regions.
+  double Deps = uris_obj.sdf_deps_close * 5.0;
   double volU = 0.0;
   double volD = 0.0;
 
-  // Let's compute left side 
-  Array<double> sUPS(1,com_mod.tnNo);
-  // std::cout << "com_mod.tnNo: " << com_mod.tnNo << std::endl;
-  // std::cout << "uris_obj.sdf size: " << uris_obj.sdf.size() << std::endl;
+  if (cm.mas(cm_mod)) {
+    std::cout << "Computing upstream region from SDF -" << Deps << " to -" 
+              << uris_obj.sdf_deps_close << " for: " << uris_obj.name << std::endl;
+    std::cout << "Computing downstream region from SDF " << uris_obj.sdf_deps_close 
+              << " to " << Deps << " for: " << uris_obj.name << std::endl;
+  }
 
+  // Compute the upstream region: negative sdf side (opposite to valve normal), 
+  // outside resistance region
+  Array<double> sUPS(1, com_mod.tnNo);
   sUPS = 0.0;
-  for (size_t j = 0; j < sUPS.size(); j++) {
-    if (uris_obj.sdf(j) >= 0.0 && uris_obj.sdf(j) <= Deps) {
-    // Reverse the sdf distance for aortic valve
-    // [HZ] Adjust this value to be more flexible about the box
-    // if (uris_obj.sdf(j) < 0.0 && uris_obj.sdf(j) >= -Deps) { 
+  for (size_t j = 0; j < com_mod.tnNo; j++) {
+    double sdf_j = uris_obj.sdf(j);
+    if (sdf_j >= -Deps && sdf_j <= -uris_obj.sdf_deps_close) {
         sUPS(0,j) = 1.0;
     }
   }
@@ -73,15 +71,14 @@ void uris_meanp(ComMod& com_mod, CmMod& cm_mod, const int iUris, const SolutionS
     volU += all_fun::integ(com_mod, cm_mod, iM, sUPS, solutions);
   }
 
-
-  // Let's compute right side
-  Array<double> sDST(1,com_mod.tnNo);
+  // Compute the downstream region: positive sdf side (valve normal direction), 
+  // outside resistance region
+  Array<double> sDST(1, com_mod.tnNo);
   sDST = 0.0;
-  for (size_t j = 0; j < sDST.size(); j++) {
-    if (uris_obj.sdf(j) < 0.0 && uris_obj.sdf(j) >= -Deps) {
-    // Reverse the sdf distance for aortic valve
-    // if (uris_obj.sdf(j) >= 0.0 && uris_obj.sdf(j) <= Deps) {
-        sDST(0,j) = 1.0;
+  for (size_t j = 0; j < com_mod.tnNo; j++) {
+    double sdf_j = uris_obj.sdf(j);
+    if (sdf_j >= uris_obj.sdf_deps_close && sdf_j <= Deps) {
+      sDST(0,j) = 1.0;
     }
   } 
 
@@ -134,7 +131,7 @@ void uris_meanp(ComMod& com_mod, CmMod& cm_mod, const int iUris, const SolutionS
 
   //  If the uris has passed the closing state
   if (uris_obj.cnt > uris_obj.DxClose.nrows()) {
-    if (uris_obj.meanPD > uris_obj.meanPU) {
+    if (uris_obj.meanPU > uris_obj.meanPD) {
       uris_obj.cnt = 1;
       uris_obj.clsFlg = false;
       com_mod.urisActFlag = true;
@@ -173,13 +170,8 @@ void uris_meanv(ComMod& com_mod, CmMod& cm_mod, const int iUris, const SolutionS
   auto& uris_obj = uris[iUris];
 
   const int nsd = com_mod.nsd;
-  // const int cEq = com_mod.cEq;
 
-  // auto& An = com_mod.An;
-  // auto& Ad = com_mod.Ad;
-  // auto& Dn = com_mod.Dn;
-
-  // Let's compute the neighboring region below the valve normal. When
+  // Compute the neighboring region with negative sdf distance. When
   // the valve is open, this region should roughly be valve oriface.
   int iEq = 0;
 
@@ -190,8 +182,6 @@ void uris_meanv(ComMod& com_mod, CmMod& cm_mod, const int iUris, const SolutionS
 
   for (int i = 0; i < com_mod.tnNo; i++) {
     if (uris_obj.sdf(i) <= -Deps) {
-    // Reverse the sdf distance for aortic valve
-    // if (uris_obj.sdf(i) >= Deps) {
       sImm(0,i) = 1.0;
     }
   }
@@ -205,7 +195,6 @@ void uris_meanv(ComMod& com_mod, CmMod& cm_mod, const int iUris, const SolutionS
   
   int m = nsd;
   int s = eq[iEq].s;
-  // int e = s + m - 1;
 
   Array<double> tmpV(maxNSD, com_mod.tnNo);
   for (int i = 0; i < nsd; i++) {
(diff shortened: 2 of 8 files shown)


Code/Source/solver/uris.cpp
```diff
@@ -39,32 +39,30 @@ void uris_meanp(ComMod& com_mod, CmMod& cm_mod, const int iUris, const SolutionS
...
-  // Let's conpute the mean pressure in the two regions of the fluid mesh 
-  // For the moment let's define a flag IdSubDmn(size the number of elements)
-
-  // Now we can compute the pressure mean on each subdomain
-  // We need to have a sdf array for each mesh
-  double Deps = uris_obj.sdf_deps * 2.5;
+  // Set the limit of the upstream and downstream regions to 5 times the closed 
+  // valve thickness. This should give a reasonable range for the upstream and 
+  // downstream regions.
+  double Deps = uris_obj.sdf_deps_close * 5.0;
```
michelebucelli: I would suggest
1. avoiding the hardcoded constant `5.0` (by introducing a named constant, e.g. `pressure_averaging_distance_factor` or something similar, and/or setting this to be user-specified);
2. trying to find a more explicit name for this variable. I can come up with `averaging_distance`, but maybe you can think of a better alternative.

Code/Source/solver/uris.cpp
```diff
@@ -173,13 +170,8 @@ void uris_meanv(ComMod& com_mod, CmMod& cm_mod, const int iUris, const SolutionS
...
   const int nsd = com_mod.nsd;
-  // const int cEq = com_mod.cEq;
 
-  // auto& An = com_mod.An;
-  // auto& Ad = com_mod.Ad;
-  // auto& Dn = com_mod.Dn;
-
-  // Let's compute the neighboring region below the valve normal. When
+  // Compute the neighboring region with negative sdf distance. When
   // the valve is open, this region should roughly be valve oriface.
```
michelebucelli: Unrelated to the PR, but: I'm not sure I understand why this is the case. 

To the best of my understanding, when the valve is open, the region with positive distance will be the one on one side of the leaflets, while the region with negative distance will be that on the other side. Even if the surface normals are pointing outwards, so that the orifice will indeed have negative signed distance, the region of negative signed distance should (or at least can) extend to unrelated parts of the domain.

Maybe there's something I'm missing?

Code/Source/solver/ComMod.h
```diff
@@ -1482,6 +1482,9 @@ class urisType
     // if the fully closed position alone is not able to prevent backflow.
     double sdf_deps_close;
 
+    // Whether to invert the valve surface normal vector.
+    bool invert_normal;
+
```
michelebucelli: I think it would be helpful to document somewhere what assumptions are made on the normal orientation (e.g. when the valve is closed, the normals must point from upstream to downstream, or something like that), so that the user can know whether they need to invert the normal or not.

Code/Source/solver/uris.cpp
michelebucelli, line 1: Is the logic used to compute the mean pressure on the two sides of the valve the same when the valve is closed and when it is open? If so, I feel that the mean pressure values when the valve is open will be less meaningful: while the signed distance function clearly separates upstream and downstream when the valve is closed, that is not the case when the valve is open (see also my other comment about this).

Maybe when the valve is open the mean pressure is never used though.

Code/Source/solver/uris.cpp
```diff
@@ -1242,7 +1233,7 @@ void uris_find_closest_face_centroid(const urisType& uris_obj, const Vector<doub
 /// @brief Compute centroid and signed distance projection along local face normal.
 double uris_compute_face_dotp(const urisType& uris_obj, const int nsd, const int jM,
                               const int Ec, const Vector<double>& xp, Array<double>& xXi, 
-                              Array<double>& lX, Vector<double>& xb) {
+                              Array<double>& lX, Vector<double>& xb, Vector<double>& unitNormal) {
```
michelebucelli: It seems to me that this function is not only computing `dotp`, as its name would suggest, but rather it is computing a relatively large number of values, corresponding to all the arguments passed in by reference. I think that the function name and signature risk to partially miscommunicate the function's intent to the user, and in general keeping track of all the things that this function does can become difficult.

In the direction of a more self-documenting implementation, I would suggest a combination of the following (which are more or less a rehash of [guidelines from here](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#s-functions)):

1. make sure that all the output arguments are actually necessary. For example, it seems to me that the value of `xXi` is not needed outside this function (but I might be mistaken). If that is the case, then it could be a function-local variable, and you would have one less thing to return (also meaning one less thing for the caller to worry about). (Counterpoint: maybe `xXi` is expensive to allocate/deallocate. Counter-counterpoint: I wouldn't worry about this unless you have profiling information that measures this to be a significant bottleneck, and there's better ways of optimizing that while keeping the function signature simple and clean).
2. Split the function into meaningfully named minimal units that only compute one thing. For example, the computation of `xb` seems to be independent of the other outputs, and could be done in a separate function; for those minimal units, the one thing being computed should be the return value (rather than an output argument passed by reference), and its name should reflect what is being computed (such as `uris_compute_xb`, or even better `uris_compute_surface_barycenter`).
3. If some of these atomic components still need to return multiple values, I would avoid having one of those as the return value (such as `dotp` here) and the others from output arguments passed by reference (such as `xb` and the others here), because this implies an asymmetry between them that is not actually part of the function's logic. That is to say, if you were to disregard point 1 above and keep this function as a single monolithic unit, I would at least move the `dotp` value to be an output argument passed by reference, rather than using a `return` statement for it. (There are actually better ways of doing that, e.g. by combining `std::pair` or `std::tuple` and `std::tie`, but I don't think that it should be necessary here).

michelebucelli (commented):
Thanks @hanzhao2020! I left a few review comment, but none of them seem critical.

Code/Source/solver/uris.cpp
hanzhao2020: Thank you for the suggestion! Would `sdf_region_factor` be a bit clearer?

Code/Source/solver/uris.cpp
hanzhao2020: Yes, this region is computed using the negative signed distance field inside the bounding box of the valve surface. In the implementation, the bounding box is defined to be 10% larger than the valve’s geometric extent, so the region may extend slightly outside the valve orifice. But this extra portion is small compared to the volume of the actual orifice.

Code/Source/solver/uris.cpp
hanzhao2020: The mean pressure is compared only when the valve is closed. If the upstream mean pressure exceeds the downstream mean pressure, the valve begins to open, and the close flag is set to false. When the close flag is false, the mean velocity in the orifice is computed in the `uris_meanv` function. If the mean velocity becomes negative, the valve close flag is set back to true. During the valve opening and closing motions, the close flag remains unchanged.

Code/Source/solver/uris.cpp
michelebucelli: Ah I see! I was missing the part about the bounding box. It's still possible that this is a larger region than the orifice if the valve is not well aligned with the box, but that's definitely better than looking at the whole domain.

What would you think about briefly explaining this in the comment, as a reminder to inexperienced developers (like me)? Something like:
```suggestion
  // Compute the neighboring region with negative sdf within the 
  // valve's bounding box. When the valve is open, this region 
  // should roughly be valve orifice.
```
(also notice the typo in "oriface" at the end).

Code/Source/solver/uris.cpp
michelebucelli: I see, thanks for the explanation! If the pressure jump is not used for closing the valve, then this is not going to be a problem.

(I think there are situations where using the pressure jump instead of the mean velocity might be useful, but that's a different story entirely, and we can discuss it some other time if it ever comes up).

Code/Source/solver/uris.cpp
hanzhao2020: Thanks for the comments! The `uris_compute_face_dotp` function is removed and split into clearer functions: `surface_element_barycenter`, `uris_find_closest_face_centroid`, and `uris_face_unit_normal` (with `xXi` kept local inside the routine). `uris_calc_sdf` now only keeps `xb` and `unitNormal` in scope, and sets `dotp` as a local variable.  Now each function matches one job.

Code/Source/solver/ComMod.h
hanzhao2020: An additional comment is included with an explanation of when it should be set to true.

Code/Source/solver/uris.cpp
hanzhao2020: Thanks for this comment. It looks good!

Code/Source/solver/uris.cpp
michelebucelli: Yes, it looks better! I'd still suggest trying to find a name that implies it is the cutoff distance used to compute the average pressure, but this can/should be explained in the comment instead.

Code/Source/solver/ComMod.h
michelebucelli: This is better, thank you!

I still think that the phrase "consistent with the upstream and downstream directions" is more vague than necessary, because it assumes that the reader knows what this consistency should be (should the normals point upstream or downstream)?

Also, am I correct in assuming that, topologically, the surface is open and not closed surface? If so, I find the distinction between outward- and inward-pointing normals a bit confusing (because that makes sense to me in the context of a closed surface).

I propose the following rephrasing (but make sure that I haven't swapped upstream or downstream 😅 ):
```cpp
// Whether to invert the valve surface normal vector. Default is false.
//
// Valve normal vectors are assumed to point downstream, so that the
// downstream region has positive signed distance and the upstream region
// has negative signed distance. If the input surface does not satisfy
// this assumption, this flag should be set to true to flip the normals.
```

Code/Source/solver/uris.cpp
michelebucelli: This seems better, thank you!

Code/Source/solver/uris.cpp
hanzhao2020: Thanks! I added a comment to explain this factor.

Code/Source/solver/ComMod.h
hanzhao2020: This looks great! I added it to the updated code. Thanks!

michelebucelli (approved):
Thanks @hanzhao2020, looks good to me!

Code/Source/solver/uris.cpp
```diff
@@ -39,32 +39,34 @@ void uris_meanp(ComMod& com_mod, CmMod& cm_mod, const int iUris, const SolutionS
...
+  // of the upstream/downstream fluid bands used for mean pressure computation.
+  double sdf_region_factor = 5.0; 
 
-  // Now we can compute the pressure mean on each subdomain
-  // We need to have a sdf array for each mesh
-  double Deps = uris_obj.sdf_deps * 2.5;
+  // Set the limit of the upstream and downstream regions to 5 times the closed 
+  // valve thickness. This should give a reasonable range for the upstream and 
+  // downstream regions.
+  double Deps = uris_obj.sdf_deps_close * sdf_region_factor;
```
aabrown100-git: Could we use a different variable name for this? Since I think `Deps` is used elsewhere in the code to mean something else right? Maybe `sdf_max` or something?

Code/Source/solver/uris.cpp
```diff
@@ -39,32 +39,34 @@ void uris_meanp(ComMod& com_mod, CmMod& cm_mod, const int iUris, const SolutionS
...
-  // std::cout << "uris_obj.sdf size: " << uris_obj.sdf.size() << std::endl;
+  if (cm.mas(cm_mod)) {
+    std::cout << "Computing upstream region from SDF -" << Deps << " to -" 
+              << uris_obj.sdf_deps_close << " for: " << uris_obj.name << std::endl;
+    std::cout << "Computing downstream region from SDF " << uris_obj.sdf_deps_close 
+              << " to " << Deps << " for: " << uris_obj.name << std::endl;
+  }
 
+  // Compute the upstream region: negative sdf side (opposite to valve normal), 
+  // outside resistance region
```
aabrown100-git, line 64: I think we've discussed this before, but is `-uris_obj.sdf_deps_close` outside of the resistance region, precisely speaking? At this distance from the valve surface, the smoothed delta function isn't zero right?

aabrown100-git (changes requested):

Code/Source/solver/uris.cpp
```diff
@@ -39,32 +39,34 @@ void uris_meanp(ComMod& com_mod, CmMod& cm_mod, const int iUris, const SolutionS
...
-  // Let's compute left side 
-  Array<double> sUPS(1,com_mod.tnNo);
-  // std::cout << "com_mod.tnNo: " << com_mod.tnNo << std::endl;
-  // std::cout << "uris_obj.sdf size: " << uris_obj.sdf.size() << std::endl;
+  if (cm.mas(cm_mod)) {
+    std::cout << "Computing upstream region from SDF -" << Deps << " to -" 
+              << uris_obj.sdf_deps_close << " for: " << uris_obj.name << std::endl;
+    std::cout << "Computing downstream region from SDF " << uris_obj.sdf_deps_close 
+              << " to " << Deps << " for: " << uris_obj.name << std::endl;
+  }
```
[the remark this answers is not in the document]
aabrown100-git: I don't mind keeping these print statements for now, since this feature is still being actively developed and improved. Once URIS valves are fully developed, we can easily remove these print statements.

Code/Source/solver/uris.cpp
hanzhao2020: For the definition of the smoothed delta function: $1/(2\varepsilon)(1+\cos(\pi y/ \varepsilon)) for |y| <= \varepsilon$, and 0 otherwise, the smoothed delta function at the half-thickness $1/(2\varepsilon)(1+\cos\pi)$ should be 0.

Code/Source/solver/uris.cpp
hanzhao2020: Yes, how about `meanp_sdf_outer_limit`?

Code/Source/solver/uris.cpp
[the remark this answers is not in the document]
hanzhao2020: Maybe we can comment out these lines in the code, and developers can temporarily uncomment them for debugging?

Code/Source/solver/uris.cpp
aabrown100-git: I like it!

Code/Source/solver/uris.cpp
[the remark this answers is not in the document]
aabrown100-git: Okay I agree! Or put them behind the `debug_uris_meanp` macro as copilot suggested.

Code/Source/solver/uris.cpp
aabrown100-git: Ah I see, thanks for explaining. Looks good to me then!

aabrown100-git (approved):
Approved, thanks @hanzhao2020. Just make those two minor changes in the unresolved comments.

Code/Source/solver/uris.cpp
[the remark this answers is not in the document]
hanzhao2020: I commented them out for now.

Code/Source/solver/uris.cpp
hanzhao2020: Changed!

hanzhao2020:
Thanks! @aabrown100-git These changes have been made.

aabrown100-git:
Excellent feel free to merge it when the checks complete!

Code/Source/solver/uris.cpp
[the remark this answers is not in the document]
michelebucelli: General question (which doesn't necessarily require changes to this PR): is there a standard `svmp` approach to debug-only messages? I saw around some messages were wrapped inside `#if` preprocessor directives, as in:
```cpp
#define dbg_msg_for_this_section 0
#if dbg_msg_for_this_section
    std::cout << "debug message" << std::endl;
#endif
```
which is, in essence, a slightly more elaborate version of commenting them in and out, seeing as there's no way to toggle the flag other than modifying the code.

Do people have any opinions on using the standard debug-mode flag (which is globally configured by CMake), and/or exposing these flags through CMake configuration?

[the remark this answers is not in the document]
hanzhao2020: I actually haven’t used the debug mode flag yet. @aabrown100-git do you have experience with the debug mode? It’s probably worth creating a new issue and gathering more opinions on this.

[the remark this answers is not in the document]
aabrown100-git: I'm not very familiar with this, but would love to discuss this more. Maybe @zasexton has ideas for debug messages as part of the refactor?

@michelebucelli Could you make a new issue about this and we can discuss?

Issue #548, opened 2026-05-19
Meaning of the enum class OutputNameType


michelebucelli:
Follow-up from [a comment on #540](https://github.com/SimVascular/svMultiPhysics/pull/540#discussion_r3197454124):

https://github.com/SimVascular/svMultiPhysics/blob/c79af8598ee27c56291f25a86540a0645178c444/Code/Source/solver/consts.h#L319-L376

It seems to me that the `enum class OutputNameType` is actually representing two disjoint sets of non-interchangeable options. In other words, it seems to me that a value of the group `outGrp_*` would not be valid in a place where a value in the group `out_*` is expected, and vice versa. Is that correct?

If it is, do you think it would make sense to split this into two distinct enumerations (say, OutputGroup and OutputVariable)? Additionally, do the numeric values attributed to the members of this enumeration matter?

PR #550, opened 2026-05-20
Fiber Stretch and Stretch Rate Computations

### LINKED ISSUE #544: Post-processing of fiber stretch and stretch rates
### Problem

For electromechanics coupling, we need fiber stretch and stretch rate computations which are variables that will feed into the cell level active stress odes. The current solver has support for a fiber stretch routine (`fib_strech` in `post.cpp`) but this has not been tested. We also need to implement a stretch rate calculation. 

### Solution

First, a test case with known stretch and stretch rates will be generated. I was thinking we simply stretch along one direction using a Dirichlet BC (see schematic below), but I am open to other ideas. (If we are just interested in testing the post-processing routines, It would be nice if this could be converted into a unit test instead of an integration test.)

Image

Regarding stretch rate calculations, I need some additional input. There are two ways in which this can be computed. 

Option 1: Finite Difference of Stretch
Option 2: Analytical Definition (see screenshot below) 

Image

I wasn't sure which was preferred... 

### Additional context

_No response_

### Code of Conduct

kko27:
## Current situation
Resolves #544 

## Release Notes 
- Added fiber stretch rate computation routine in `post.cpp`. Stretch rate is computed as a FD of fiber stretch 
- Added integration test for stretch and stretch rates

## Testing
Integration test has been added that checks if the stretch and stretch rates computed by the code match reference solutions that were obtained analytically. 

### DIFF
--- a/Code/Source/solver/consts.h
+++ b/Code/Source/solver/consts.h
@@ -342,9 +342,11 @@ enum class OutputNameType
   outGrp_divV = 521, 
   outGrp_Visc = 522,
   outGrp_fS = 523,
-  outGrp_C = 524, 
+  outGrp_C = 524,
   outGrp_I1 = 525,
   outGrp_ionicState = 526,
+  outGrp_I4f = 527,
+  outGrp_I4fRate = 528,
 
   out_velocity = 599,
   out_pressure = 598, 
@@ -373,7 +375,9 @@ enum class OutputNameType
   out_viscosity = 575,
   out_fibStrn = 574,
   out_CGstrain = 573,
-  out_CGInv1 = 572
+  out_CGInv1 = 572,
+  out_fibStretch = 571,
+  out_fibStretchRate = 570
 };
 
 /// @brief Simulation output file types. 
--- a/Code/Source/solver/consts.h
+++ b/Code/Source/solver/consts.h
@@ -345,8 +345,8 @@ enum class OutputNameType
   outGrp_C = 524,
   outGrp_I1 = 525,
   outGrp_ionicState = 526,
-  outGrp_I4f = 527,
-  outGrp_I4fRate = 528,
+  outGrp_fibStretch = 527,
+  outGrp_fibStretchRate = 528,
 
   out_velocity = 599,
   out_pressure = 598, 
@@ -501,4 +501,3 @@ enum class LinearAlgebraType {
 };
 
 #endif
-
--- a/Code/Source/solver/post.cpp
+++ b/Code/Source/solver/post.cpp
@@ -3,6 +3,7 @@
 
 #include "post.h"
 
+#include "FE/Common/FEException.h"
 #include "all_fun.h"
 #include "fluid.h"
 #include "fs.h"
@@ -43,6 +44,9 @@ namespace post {
       }   
       return F;
     }
+
+    void compute_fib_stretch(const ComMod& com_mod, const int iEq, const mshType& lM,
+        const Array<double>& lD, Vector<double>& res);
   }
 
 void all_post(Simulation* simulation, Array<double>& res, const SolutionStates& solutions,
@@ -717,23 +721,48 @@ void fib_dir_post(Simulation* simulation, const mshType& lM, const int nFn, Arra
 
 }
 
-/// @brief Compute fiber stretch λ = sqrt(I_{4,f}) from a given displacement field.
+/// @brief Compute fiber stretch based on 4th invariant: λ = sqrt(I_{4,f})
 //
-void compute_fib_stretch(Simulation* simulation, const int iEq, const mshType& lM,
+void fib_stretch(Simulation* simulation, const int iEq, const mshType& lM, const SolutionStates& solutions, Vector<double>& res)
+{
+  compute_fib_stretch(simulation->com_mod, iEq, lM, solutions.current.get_displacement(), res);
+}
+
+/// @brief Compute fiber stretch rate dλ/dt via backward finite difference.
+//
+void fib_stretch_rate(Simulation* simulation, const int iEq, const mshType& lM, const SolutionStates& solutions, Vector<double>& res)
+{
+  auto& com_mod = simulation->com_mod;
+  const double dt = com_mod.dt;
+  int nNo = lM.nNo;
+
+  if (res.size() != nNo) {
+    svmp::raise<svmp::FE::InvalidArgumentException>(
+        SVMP_HERE,
+        "[fib_stretch_rate] Expected res size " + std::to_string(nNo) + ", but got " + std::to_string(res.size()) + ".");
+  }
+
+  Vector<double> lambda_old(nNo);
+
+  compute_fib_stretch(simulation->com_mod, iEq, lM, solutions.current.get_displacement(), res);
+  compute_fib_stretch(simulation->com_mod, iEq, lM, solutions.old.get_displacement(),     lambda_old);
+
+  res = (res - lambda_old) / dt; 
+}
+
+namespace {
+/// @brief Compute fiber stretch lambda = sqrt(I_{4,f}) from a given displacement field.
+void compute_fib_stretch(const ComMod& com_mod, const int iEq, const mshType& lM,
     const Array<double>& lD, Vector<double>& res)
 {
   using namespace consts;
 
-  auto& com_mod = simulation->com_mod;
   auto& eq = com_mod.eq[iEq];
-
   int nsd = com_mod.nsd;
   int tnNo = com_mod.tnNo;
   int tDof = com_mod.tDof;
   int eNoN = lM.eNoN;
   int i = eq.s;
-  int j = i + 1;
-  int k = j + 1;
 
   Vector<double> sA(tnNo);
   Vector<double> sF(tnNo);
@@ -762,9 +791,6 @@ void compute_fib_stretch(Simulation* simulation, const int iEq, const mshType& l
         nn::gnn(eNoN, nsd, nsd, Nxi, xl, Nx, Jac, ksix);
       }
 
-      double w = lM.w(g)*Jac;
-      auto N = lM.N.col(g);
-
       // Compute Deformation Gradient: F = I + grad(u)
       F = deformation_gradient(Nx, dl, nsd, eNoN, i);
 
@@ -773,6 +799,8 @@ void compute_fib_stretch(Simulation* simulation, const int iEq, const mshType& l
       double lambda = sqrt(utils::norm(fl));
 
       // L2 projection from integration points to nodes
+      double w = lM.w(g)*Jac;
+      auto N = lM.N.col(g);
       for (int a = 0; a < eNoN; a++) {
         int Ac = lM.IEN(a,e);
         sA(Ac) = sA(Ac) + w*N(a);
@@ -793,42 +821,7 @@ void compute_fib_stretch(Simulation* simulation, const int iEq, const mshType& l
     }
   }
 }
-
-
-/// @brief Compute fiber stretch based on 4th invariant: λ = sqrt(I_{4,f})
-//
-void fib_stretch(Simulation* simulation, const int iEq, const mshType& lM, const SolutionStates& solutions, Vector<double>& res)
-{
-  auto& com_mod = simulation->com_mod;
-  auto& eq = com_mod.eq[iEq];
-  com_mod.dof = eq.dof;
-  compute_fib_stretch(simulation, iEq, lM, solutions.current.get_displacement(), res);
 }
-
-
-/// @brief Compute fiber stretch rate dλ/dt via backward finite difference.
-//
-void fib_stretch_rate(Simulation* simulation, const int iEq, const mshType& lM, const SolutionStates& solutions, Vector<double>& res)
-{
-  auto& com_mod = simulation->com_mod;
-  auto& eq = com_mod.eq[iEq];
-  com_mod.dof = eq.dof;
-
-  const double dt = com_mod.dt;
-  int nNo = lM.nNo;
-
-  Vector<double> lambda_curr(nNo);
-  Vector<double> lambda_old(nNo);
-
-  compute_fib_stretch(simulation, iEq, lM, solutions.current.get_displacement(), lambda_curr);
-  compute_fib_stretch(simulation, iEq, lM, solutions.old.get_displacement(),     lambda_old);
-
-  res = 0.0;
-  for (int a = 0; a < nNo; a++) {
-    res(a) = (lambda_curr(a) - lambda_old(a)) / dt;
-  }
-}
-
 
 void post(Simulation* simulation, const mshType& lM, Array<double>& res, const SolutionStates& solutions,
     consts::OutputNameType outGrp, const int iEq)
--- a/Code/Source/solver/post.cpp
+++ b/Code/Source/solver/post.cpp
@@ -740,7 +740,12 @@ void fib_stretch(const ComMod& com_mod, const int iEq, const mshType& lM,
   Array<double> Nx(nsd,eNoN);
 
   for (int e = 0; e < lM.nEl; e++) {
-    all_fun::domain(com_mod, lM, iEq, e);
+    int cDmn = all_fun::domain(com_mod, lM, iEq, e);
+    auto cPhys = eq.dmn[cDmn].phys;
+    if (cPhys != EquationType::phys_struct && cPhys != EquationType::phys_ustruct) {
+      continue;
+    }
+
     if (lM.eType == ElementType::NRB) {
       //CALL NRBNNX(lM, e)
     }
@@ -754,10 +759,10 @@ void fib_stretch(const ComMod& com_mod, const int iEq, const mshType& lM,
     for (int g = 0; g < lM.nG; g++) {
       double Jac = 0.0;
       Array<double> F(nsd,nsd);
-      Array<double> ksix(nsd,nsd);
       if (g == 0 || !lM.lShpF) {
         auto Nxi = lM.Nx.slice(g);
-        nn::gnn(eNoN, nsd, nsd, Nxi, xl, Nx, Jac, ksix);
+        Array<double> dummy_ksix(nsd,nsd);
+        nn::gnn(eNoN, nsd, nsd, Nxi, xl, Nx, Jac, dummy_ksix);
       }
 
       // Compute Deformation Gradient: F = I + grad(u)
--- a/Code/Source/solver/post.cpp
+++ b/Code/Source/solver/post.cpp
@@ -742,7 +742,7 @@ void fib_dir_post(Simulation* simulation, const mshType& lM, const int nFn, Arra
 
 /// @brief L2-project nodal fiber stretch λ = sqrt(I_{4,f}) from a given displacement field.
 //
-static void compute_fib_stretch(Simulation* simulation, const int iEq, const mshType& lM,
+void compute_fib_stretch(Simulation* simulation, const int iEq, const mshType& lM,
     const Array<double>& lD, Vector<double>& res)
 {
   using namespace consts;
--- a/Code/Source/solver/post.cpp
+++ b/Code/Source/solver/post.cpp
@@ -740,7 +740,7 @@ void fib_dir_post(Simulation* simulation, const mshType& lM, const int nFn, Arra
 
 }
 
-/// @brief L2-project nodal fiber stretch λ = sqrt(I_{4,f}) from a given displacement field.
+/// @brief Compute fiber stretch λ = sqrt(I_{4,f}) from a given displacement field.
 //
 void compute_fib_stretch(Simulation* simulation, const int iEq, const mshType& lM,
     const Array<double>& lD, Vector<double>& res)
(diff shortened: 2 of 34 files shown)


Code/Source/solver/post.cpp
```diff
@@ -740,45 +740,37 @@ void fib_dir_post(Simulation* simulation, const mshType& lM, const int nFn, Arra
 
 }
 
-/// @brief Compute fiber stretch based on 4th invariant: I_{4,f}
+/// @brief L2-project nodal fiber stretch λ = sqrt(I_{4,f}) from a given displacement field.
```
aabrown100-git: L2-projection?

Code/Source/solver/post.cpp
```diff
@@ -837,9 +833,44 @@ void fib_strech(Simulation* simulation, const int iEq, const mshType& lM, const
...
+void fib_stretch(Simulation* simulation, const int iEq, const mshType& lM, const SolutionStates& solutions, Vector<double>& res)
+{
+  auto& com_mod = simulation->com_mod;
+  auto& eq = com_mod.eq[iEq];
+  com_mod.dof = eq.dof;
+  compute_fib_stretch(simulation, iEq, lM, solutions.current.get_displacement(), res);
 }
 
+
+/// @brief Compute fiber stretch rate dλ/dt via backward finite difference.
```
aabrown100-git, line 799: Did you consider using the velocity field to compute this directly, rather than using finite difference?

Code/Source/solver/vtk_xml.cpp
```diff
@@ -1293,14 +1293,34 @@ void write_vtus(Simulation* simulation, const SolutionStates& solutions, const b
...
-            tmpV.resize(l,msh.nNo); 
+            tmpV.resize(l,msh.nNo);
             post::div_post(simulation, msh, tmpV, solutions, iEq);
             for (int a = 0; a < msh.nNo; a++) {
               d[iM].x(is,a) = tmpV(0,a);
             }
             tmpV.resize(consts::maxNSD,msh.nNo);
           break;
 
+          case OutputNameType::outGrp_I4f: {
```
aabrown100-git: Should we change these names, since we're not outputting I4f anymore?

tests/cases/struct/uniaxial_block_stretch/fiberstretch.png
aabrown100-git, line 1: Can you add a plot of fiber stretch rate too?

aabrown100-git (changes requested):
Nice work, made a few suggestions

Code/Source/solver/consts.h
```diff
@@ -342,9 +342,11 @@ enum class OutputNameType
   outGrp_divV = 521, 
   outGrp_Visc = 522,
   outGrp_fS = 523,
-  outGrp_C = 524, 
+  outGrp_C = 524,
   outGrp_I1 = 525,
   outGrp_ionicState = 526,
+  outGrp_I4f = 527,
+  outGrp_I4fRate = 528,
```
michelebucelli: I would suggest to rename these to `outGrp_fibStretch` and `outGrp_fibStretchRate`, for consistency with the ones below.

Having said that: I'm not sure I fully understand the distinction between output groups and output variables, so this might not make sense, but would it make sense for the fiber stretch and stretch rate to belong to the same group?

Code/Source/solver/post.cpp
```diff
@@ -740,45 +740,37 @@ void fib_dir_post(Simulation* simulation, const mshType& lM, const int nFn, Arra
 
 }
 
-/// @brief Compute fiber stretch based on 4th invariant: I_{4,f}
+/// @brief L2-project nodal fiber stretch λ = sqrt(I_{4,f}) from a given displacement field.
 //
-void fib_strech(Simulation* simulation, const int iEq, const mshType& lM, const SolutionStates& solutions, Vector<double>& res)
+void compute_fib_stretch(Simulation* simulation, const int iEq, const mshType& lM,
+    const Array<double>& lD, Vector<double>& res)
```
michelebucelli: Is this function ever called/meant to be called except through `fib_stretch` and `fib_stretch_rate`? If not, it might make sense for this to be removed from `post.hpp` and defined in an anonymous namespace in `post.cpp`, so that it is clearly marked as internal.

Code/Source/solver/post.cpp
```diff
@@ -815,13 +809,15 @@ void fib_strech(Simulation* simulation, const int iEq, const mshType& lM, const
         }
       }
```
michelebucelli: As you suggested, I think it might be a good idea to encapsulate this into a small function, that then could be reused in the other places in this file where these same lines of code are repeated.

Perhaps that function could go into the anonymous namespace as above.

Code/Source/solver/post.cpp
```diff
@@ -815,13 +809,15 @@ void fib_strech(Simulation* simulation, const int iEq, const mshType& lM, const
...
 
+      // Compute fiber stretch based on 4th invariant: I_{4,f} = F.fN.F.fN
       auto fl = mat_fun::mat_mul(F, lM.fN.rows(0,nsd-1,e));
-      double I4f = utils::norm(fl);
+      double lambda = sqrt(utils::norm(fl));
```
michelebucelli, line 773: I'm not sure about this square root here: doesn't the `norm` already compute the square root?

Code/Source/solver/vtk_xml.cpp
michelebucelli: I agree, and in general I would suggest to choose one among `lambda`, `I4f` and `fiber_stretch` (or abbreviations thereof) and use it everywhere in comments, variable and function names.

Among these I have a slight preference for `fiber_stretch` without abbreviation, since it seems to me to be the most explicit.

Code/Source/solver/post.cpp
```diff
@@ -837,9 +833,44 @@ void fib_strech(Simulation* simulation, const int iEq, const mshType& lM, const
...
+  Vector<double> lambda_curr(nNo);
+  Vector<double> lambda_old(nNo);
+
+  compute_fib_stretch(simulation, iEq, lM, solutions.current.get_displacement(), lambda_curr);
+  compute_fib_stretch(simulation, iEq, lM, solutions.old.get_displacement(),     lambda_old);
+
+  res = 0.0;
+  for (int a = 0; a < nNo; a++) {
+    res(a) = (lambda_curr(a) - lambda_old(a)) / dt;
+  }
```
michelebucelli: 1. I would store the result of the first call to `compute_fib_stretch` directly into `res`.
2. I would replace the `for` loop with calls to the `Vector` operators.

The code would then look as follows:
```suggestion
  Vector<double> lambda_old(nNo);

  compute_fib_stretch(simulation, iEq, lM, solutions.current.get_displacement(), res);
  compute_fib_stretch(simulation, iEq, lM, solutions.old.get_displacement(),     lambda_old);

  res = (res - lambda_old) / dt;
```

In fact, this probably has as many allocations as before because of the need to allocate a temporary to store the result of the `-` operation (and maybe another for the `/` operation, not sure if the compiler will elide that copy). In this regard, I think it would probably be beneficial to add assignment operators to `Vector` so that the following could become possible:
```cpp
res -= lambda_old;
res /= dt;
```

This is probably not super important as of now, but these vectors might become large for organ-scale simulations.

Code/Source/solver/post.cpp
```diff
@@ -837,9 +833,44 @@ void fib_strech(Simulation* simulation, const int iEq, const mshType& lM, const
...
+/// @brief Compute fiber stretch rate dλ/dt via backward finite difference.
+//
+void fib_stretch_rate(Simulation* simulation, const int iEq, const mshType& lM, const SolutionStates& solutions, Vector<double>& res)
+{
+  auto& com_mod = simulation->com_mod;
+  auto& eq = com_mod.eq[iEq];
+  com_mod.dof = eq.dof;
+
+  const double dt = com_mod.dt;
+  int nNo = lM.nNo;
```
michelebucelli, line 804: To check correctness, I suggest adding an assertion that `res` has exactly `nNo` elements.

Code/Source/solver/post.cpp
```diff
@@ -740,45 +740,37 @@ void fib_dir_post(Simulation* simulation, const mshType& lM, const int nFn, Arra
 
 }
 
-/// @brief Compute fiber stretch based on 4th invariant: I_{4,f}
+/// @brief L2-project nodal fiber stretch λ = sqrt(I_{4,f}) from a given displacement field.
 //
-void fib_strech(Simulation* simulation, const int iEq, const mshType& lM, const SolutionStates& solutions, Vector<double>& res)
+void compute_fib_stretch(Simulation* simulation, const int iEq, const mshType& lM,
```
michelebucelli: 1. Is `simulation` changed by this function? If not, I would mark it as `const`.
2. Since I think there's no use case where `simulation` can be `nullptr`, I would pass it in as a reference instead of a pointer (if the caller has a pointer, then it can dereference it).
3. Am I correct that `simulation` is only used to access its `com_mod` member? If so, I think it might be better to pass in `com_mod` directly, instead of indirectly through `simulation`.

The same suggestions apply to `fib_stretch` and `fib_stretch_rate` below.

Code/Source/solver/post.cpp
```diff
@@ -787,16 +779,18 @@ void fib_strech(Simulation* simulation, const int iEq, const mshType& lM, const
...
+      Array<double> ksix(nsd,nsd);
       if (g == 0 || !lM.lShpF) {
-        auto Nx = lM.Nx.slice(g);
-        nn::gnn(eNoN, nsd, nsd, Nx, xl, Nx, Jac, F);
+        auto Nxi = lM.Nx.slice(g);
+        nn::gnn(eNoN, nsd, nsd, Nxi, xl, Nx, Jac, ksix);
       }
 
       double w = lM.w(g)*Jac;
       auto N = lM.N.col(g);
```
michelebucelli: I would suggest moving these below, just before the `for` loop in which they are used (line 816), so that their purpose is more immediately clear.

Code/Source/solver/post.cpp
```diff
@@ -787,16 +779,18 @@ void fib_strech(Simulation* simulation, const int iEq, const mshType& lM, const
     for (int g = 0; g < lM.nG; g++) {
       double Jac = 0.0;
       Array<double> F(nsd,nsd);
+      Array<double> ksix(nsd,nsd);
       if (g == 0 || !lM.lShpF) {
-        auto Nx = lM.Nx.slice(g);
-        nn::gnn(eNoN, nsd, nsd, Nx, xl, Nx, Jac, F);
+        auto Nxi = lM.Nx.slice(g);
+        nn::gnn(eNoN, nsd, nsd, Nxi, xl, Nx, Jac, ksix);
       }
```
michelebucelli, line 766: I am a bit confused by these lines of code. If I get it right, their purpose is just to evaluate `Nx`, while `ksix` and `Nxi` are placeholders that are only created to be able to call `nn::gnn`. Is this correct, or am I missing something?

If it is, I would suggest
1. to move `ksix` in the same scope as `Nxi`, to make more evident that whatever is written in it is immediately discarded, and
2. to give these two variables names that make their placeholder nature more immediately evident (such as `dummy_Nxi` and `dummy_ksix`, or something of the kind).

michelebucelli (commented):
Thank you @kko27! I have left a few suggestions and comments.

Code/Source/solver/post.cpp
kko27: Yes, we discussed this in #544 and decided to do a FD approach first. This is because the stabilization approach we want to apply to the [coupled electromechanics staggered scheme](https://www.sciencedirect.com/science/article/pii/S0045782520306915) is based on first order FD. Using the velocity field has the advantage of retaining 2nd order accuracy, but we are prioritizing the fastest approach to establish an electromechanics framework in the codebase. 

I do plan to investigate the direct computation of the stretch rate and its effect on the coupled electromechanics in the future.

Code/Source/solver/post.cpp
kko27: There is a projection operation to project the fiber stretch from the Gauss points to the nodal points. But this happens towards the end of the code and is definitely not the meat of this routine. Happy to edit this comment.

Code/Source/solver/post.cpp
kko27: Ah yes, I forgot to do this. Thanks!

Code/Source/solver/post.cpp
kko27: I don't think `norm` computes the square root (which means the name is technically misleading). See below: 

```
double norm(const Vector<double>& U)
{
  double norm = 0.0;
  for (int i = 0; i < U.size(); i++) {
    norm += U(i)*U(i);
  }
  return norm;
}
```

Code/Source/solver/post.cpp
kko27: This function is internal, so I will follow your advice

Code/Source/solver/post.cpp
```diff
@@ -740,45 +740,37 @@ void fib_dir_post(Simulation* simulation, const mshType& lM, const int nFn, Arra
...
+  Array<double> xl(nsd,eNoN);
   Array<double> dl(tDof,eNoN);
-  Array<double> Nx(nsd,eNoN); 
-  Vector<double> N(eNoN);
+  Array<double> Nx(nsd,eNoN);
 
   for (int e = 0; e < lM.nEl; e++) {
-    int cDmn  = all_fun::domain(com_mod, lM, iEq, e);
-    auto cPhys = eq.dmn[cDmn].phys;
+    all_fun::domain(com_mod, lM, iEq, e);
```
javijv4: Is this being used? The `all_fun::domain` function returns the domain ID, but that value isn't assigned to anything, so I think it is not doing anything. 
I think for purely mechanical or electromechanical problems, this calculation should be domain-agnostic, but for FSI problems, maybe it makes sense to check the domain physics and skip the fluid ones?

Code/Source/solver/post.cpp
michelebucelli: Sorry, my bad for not checking the definition of the function.

I do agree that its name is misleading though - I think it might be a good idea to rename it to `norm_squared`. I'll open an issue about this.

Code/Source/solver/post.cpp
```diff
@@ -837,9 +833,44 @@ void fib_strech(Simulation* simulation, const int iEq, const mshType& lM, const
       res(a) = res(a) + sF(Ac) / sA(Ac);
     }
   }
+}
 
+
+/// @brief Compute fiber stretch based on 4th invariant: λ = sqrt(I_{4,f})
+//
+void fib_stretch(Simulation* simulation, const int iEq, const mshType& lM, const SolutionStates& solutions, Vector<double>& res)
```
javijv4: I don't think this is an issue, but I am unsure of why we need both `fib_stretch` and `compute_fib_stretch`. Can we just call the `compute_fib_stretch` directly and bypass `fib_stretch`?

Code/Source/solver/post.cpp
```diff
@@ -837,9 +833,44 @@ void fib_strech(Simulation* simulation, const int iEq, const mshType& lM, const
...
+{
+  auto& com_mod = simulation->com_mod;
+  auto& eq = com_mod.eq[iEq];
+  com_mod.dof = eq.dof;
```
javijv4: I don't think any of the three statements above are being used.

javijv4:
Hey @kko27, thanks for working on this. Besides the comments I left, I asked Codex to run a review, and it found this issue:


> In cep_ion.cpp (line 160), post::fib_stretch(simulation, iEq, ...) is called with iEq for the CEP equation. But compute_fib_stretch() (line 745) uses com_mod.eq[iEq].s at post.cpp (line 757) to choose which rows of the global displacement state feed F at post.cpp (line 791). In a coupled CEP+struct run, the CEP equation’s s is not the structural displacement offset, so the stretch-activated current path can be driven by unrelated data.

I think it's right. You are passing the EP index equation, not the mechanics, to `fib_stretch,` so that's going to fail. I don't think this PR addresses the electromechanics coupling, so this is a future issue, but it might be worth commenting out (or deleting?) the `fib_stretch` call in `cep_ion.cpp` until we have tested it?

Code/Source/solver/post.cpp
kko27: Good catch. These are not being used

Code/Source/solver/post.cpp
kko27: @javijv4 my only justification for this is for symmetry and having a clean wrapper for output processing. Not sure if @michelebucelli had thoughts on this

michelebucelli: Yeah when I saw the code the first time I interpreted as a choice due to symmetry. However, on second thought, maybe I would have only one function (especially if `fib_stretch` just calls `compute_fib_stretch` without doing anything else).

I would call that `fib_stretch`, for symmetry with `fib_stretch_rate`.

Code/Source/solver/consts.h
kko27: I have renamed these to `outGrp_fibStretch` and `outGrp_fibStretchRate` as suggested. But I'm still not sure about the distinction between out and Outgrp. When I look at `vtk_xml`, it seems that even the group variables are only outputting one field at a time. I just followed the current style, which might not be what was intended. Maybe we can follow this up via issue #548

aabrown100-git (approved):

Code/Source/solver/post.cpp
kko27: I agree with these suggestions. I will pass `com_mod` instead of `simulation` for both `fib_stretch` and `fib_stretch_rate`

Code/Source/solver/post.cpp
kko27: @michelebucelli, I agree with point 1 of moving `ksix` to the same scope as Nxi. I also agree that `ksix` is an output from `nn::gnn` that we don't use. We should label this as `dummy_ksix`. 

Regarding `Nxi`, I think this is the input shape function gradient (wrt parametric coordinates of parent element) that undergoes a coordinate transformation to return Nx (shape function gradient wrt reference configuration). For example, lines 509-511 of `nn.cpp`: 

```
    for (int a = 0; a < eNoN; a++) {
      Nx(0,a) = Nx(0,a) + Nxi(0,a)*xiX(0,0) + Nxi(1,a)*xiX(1,0) + Nxi(2,a)*xiX(2,0);
      Nx(1,a) = Nx(1,a) + Nxi(0,a)*xiX(0,1) + Nxi(1,a)*xiX(1,1) + Nxi(2,a)*xiX(2,1);
      Nx(2,a) = Nx(2,a) + Nxi(0,a)*xiX(0,2) + Nxi(1,a)*xiX(1,2) + Nxi(2,a)*xiX(2,2);
    }
```

The previous call to `gnn` contained a bug. 
```
 auto Nx = lM.Nx.slice(g);
 nn::gnn(eNoN, nsd, nsd, Nx, xl, Nx, Jac, F);
```
Labeling both input and output as `Nx` seems unsafe. I believe this inconsistency was fixed for a very similar routine (fib_dir_post) in a previous PR: https://github.com/SimVascular/svMultiPhysics/pull/435/changes/84f2ea67517cb3ace65dfa55e521aae49b428ddf

kko27:
> Hey @kko27, thanks for working on this. Besides the comments I left, I asked Codex to run a review, and it found this issue:
> 
> > In cep_ion.cpp (line 160), post::fib_stretch(simulation, iEq, ...) is called with iEq for the CEP equation. But compute_fib_stretch() (line 745) uses com_mod.eq[iEq].s at post.cpp (line 757) to choose which rows of the global displacement state feed F at post.cpp (line 791). In a coupled CEP+struct run, the CEP equation’s s is not the structural displacement offset, so the stretch-activated current path can be driven by unrelated data.
> 
> I think it's right. You are passing the EP index equation, not the mechanics, to `fib_stretch,` so that's going to fail. I don't think this PR addresses the electromechanics coupling, so this is a future issue, but it might be worth commenting out (or deleting?) the `fib_stretch` call in `cep_ion.cpp` until we have tested it?

Oh dear, this will definitely become a problem for coupling. We don't need to address this in this issue as right now users cannot enable electromechanics coupling, but I will add this note to the main issue #523

Code/Source/solver/post.cpp
ktbolt: @kko27 Using Nx for both input and output is a bug. I don't see this problem in other calls to gnn.

Code/Source/solver/post.cpp
kko27: `all_fun::domain` was not being used. I have copied over a similar domain check (as done in `fib_algn_post`) so that fiber stretch computations are only applied for structural domains for FSI problems. 

```
    int cDmn = all_fun::domain(com_mod, lM, iEq, e);
    auto cPhys = eq.dmn[cDmn].phys;
    if (cPhys != EquationType::phys_struct && cPhys != EquationType::phys_ustruct) {
      continue; 
    } 
```

kko27:
@michelebucelli and @javijv4 I think this is ready for final review and approval

javijv4 (approved):
Looks good to me!

michelebucelli (approved):
Looks good to me!

Issue #528, opened 2026-04-14
Top-level abstraction for partitioned FSI with growth and remodeling (svFSGe)


Eleven7825:
### Use Case

@mrp089 and I want to merge [svFSGe](https://github.com/StanfordCBCL/svFSGe)  a partitioned fluid-solid-growth-and-remodeling (FSGe) solver — into svMultiPhysics. svFSGe uses a partitioned approach where fluid, solid, and mesh motion are solved as separate subdomains, coupled at a shared interface. This is distinct from the existing monolithic Equation_FSI in svMultiPhysics, where fluid and solid share one linear system, one time loop, and one Integrator.

The partitioned approach is required because:

* The fluid subdomain runs a full pulsatile time loop (N cardiac cycle steps) per coupling sub-iteration to produce time-averaged wall shear stress and pressure; or in steady fluid case, the last time step.
* The solid subdomain is quasi-static — it takes a single load step with no time marching, only Newton iterations for equilibrium.
These requirements cannot be satisfied by the current single-Simulation / single-Integrator architecture.

### Problem

The current architecture assumes one Simulation owns one Integrator which owns one SolutionStates. This maps cleanly to the existing use cases (monolithic fluid, solid, FSI). However, for partitioned FSI:

There is no place to own the coupling loop. The load-step ramping, sub-iteration loop, and convergence check (IQN-ILS  / Aitken relaxation) have no home in the current class hierarchy.

There is no place to own multiple independent time integrators. The fluid subdomain needs N time steps per coupling sub-iteration; the solid subdomain needs zero time steps. These cannot share one Integrator or one SolutionStates.

There is no abstraction for interface data exchange. The displacements, wall shear stress, and pressure that are transferred between fluid and solid at each coupling sub-iteration are not represented anywhere.

### Solution

Build a PartitionedSimulation class on top of the existing Simulation / Integrator / SolutionStates stack introduced in #450. The existing classes are left unchanged — partitioning lives above them.

New abstractions:

InterfaceData — data exchanged at the fluid-solid interface each coupling sub-iteration:

```
struct InterfaceData {
    Array<double> displacement;  // interface node displacements (solid → fluid/mesh)
    Array<double> wss;           // wall shear stress (fluid → solid)
    Array<double> pressure;      // fluid pressure on wall (fluid → solid)
    // node mapping: fluid interface ↔ solid interface
};
```
CouplingAlgorithm — abstract strategy for the fixed-point acceleration method:

```
[...]

PR #554, opened 2026-05-22
Implements partitioned Fluid-Structure Interaction (FSI) coupling in svMultiPhysics

Eleven7825:
This PR implements partitioned Fluid-Structure Interaction (FSI) coupling in svMultiPhysics, building on the work in `mrp089:feature/issue-431-partitioned-fsi`. It introduces a `PartitionedFSI` class that orchestrates three independently solved sub-simulations (fluid, solid, mesh) with a Dirichlet-Neumann coupling loop and Aitken relaxation.

This is the **first step** toward the broader partitioned FSI + Growth and Remodeling framework described in [issue #528](https://github.com/SimVascular/svMultiPhysics/issues/528), which proposes `InterfaceData`, an abstract `CouplingAlgorithm` strategy, and a `PartitionedSimulation` driver capable of supporting multi-rate time integration and IQN/JFNK acceleration. The architecture introduced here — three independent `Simulation` objects with explicit interface transfer — is designed to extend cleanly toward those goals.

## What's included

- `PartitionedFSI` class: drives the coupling loop between independently initialized fluid, solid, and mesh sub-simulations
- Dirichlet-Neumann coupling: solid displacement → mesh solve → fluid ALE solve → fluid traction → solid solve
- Interface relaxation: constant-factor and Aitken Δ² acceleration (Küttler & Wall 2008)
- MPI multi-core support: all three sub-simulations are fully distributed; interface arrays are globally replicated via `MPI_Allgatherv` so convergence checks and Aitken updates are consistent across all ranks
- `initialize_partitioned_fsi` in `Simulation`: broadcasts config from rank 0 to all ranks so slave ranks enter the coupling constructor (fixes deadlock on >1 MPI process)
- Test case: `tests/cases/fsi/pipe_3d_partitioned/` — 3D pipe flow with a compliant wall, ramp and steady-state XMLs
- Coupling convergence log written to `coupling.dat` each time step


ktbolt:
@Eleven7825 At 122 commits so far I kind of feel that this draft PR is premature, like you are still doing basic implementation. The commit history is a mess and each commit is doing a CI for all tests, kind of a waste of resources. And I don't think the code is actually being reviewed.

Maybe squash all of these commits, finish the initial development and reopen the PR.

Eleven7825:
@mrp089 @zasexton This pr is ready for review.

tests/cases/fsi/pipe_3d_partitioned/.gitignore
mrp089, line 1: Shouldn't be necessary, lots of artifacts. There's already a top-level .gitignore.

tests/cases/fsi/pipe_3d/solver_10step.xml
mrp089, line 1: clean up all files in pipe_3d

tests/cases/fsi/pipe_3d_partitioned/configuration.png
mrp089, line 1: clean up all files that were copied from the monolithic case or are artifacts of debugging

tests/conftest.py
mrp089, line 1: this file should not be changed

tests/test_fsi.py
```diff
@@ -30,4 +32,28 @@ def test_pipe_3d_trilinos_ml(n_proc):
...
+    # independently, so multi-proc runs only match the 1-proc run once the
+    # Dirichlet-Neumann coupling has converged tightly. Two settings in
+    # solver.xml make that possible at the original tolerances:
+    #   - Coupling_tolerance = 1e-10 (drives all proc counts to the same fixed point)
+    #   - Time_step_size = 1e-2: a larger step weakens the added-mass coupling
+    #     stiffness (which grows ~1/dt for partitioned FSI), so the Aitken
+    #     iteration converges in ~15 steps at 1, 3 and 4 procs instead of
+    #     stagnating at 4 procs. At convergence the partition-count round-off is
+    #     5-7 orders below the global RTOL for every field.
+    run_with_reference(base_folder, test_folder, fields=[], n_proc=n_proc, t_max=t_max,
```
mrp089, line 53: less verbose comment

tests/test_fsi.py
```diff
@@ -30,4 +32,28 @@ def test_pipe_3d_trilinos_ml(n_proc):
 def test_pipe_RCR_3d(n_proc):
     test_folder = "pipe_RCR_3d"
     t_max = 5
-    run_with_reference(base_folder, test_folder, fields, n_proc, t_max)
\ No newline at end of file
+    run_with_reference(base_folder, test_folder, fields, n_proc, t_max)
+
+def test_pipe_3d_partitioned(n_proc):
```
mrp089, line 38: `test_pipe_RCR_3d_partitioned` (use also for folder)

tests/test_fsi.py
```diff
@@ -30,4 +32,28 @@ def test_pipe_3d_trilinos_ml(n_proc):
 def test_pipe_RCR_3d(n_proc):
```
mrp089, line 33: `test_pipe_RCR_3d_monolithic` (use also for folder)

Code/Source/solver/CMakeLists.txt
```diff
@@ -357,7 +359,13 @@ if(ENABLE_UNIT_TEST)
 
   # include source files (same as what svMultiPhysics does except for main.cpp)
   add_executable(run_all_unit_tests ${CSRCS})
-  
+
+  # Define test data directory for integration tests
+  # CMAKE_SOURCE_DIR points to Code/, so go up one level to reach the repo root
+  target_compile_definitions(run_all_unit_tests PRIVATE
+    TEST_DATA_DIR="${CMAKE_SOURCE_DIR}/../tests/cases"
```
mrp089, line 391: why is this necessary?

Code/Source/solver/ComMod.h
```diff
@@ -1571,9 +1571,14 @@ class ComMod {
...
 
-    /// @brief Whether mesh is moving
+    /// @brief Whether mesh is moving (monolithic FSI: mesh velocity in DOFs nsd+1..2*nsd)
     bool mvMsh = false;
 
+    /// @brief ALE mesh velocity for partitioned FSI (nsd, tnNo).
+    /// When non-empty, the fluid assembly subtracts this from the convective
+    /// velocity, providing the ALE correction without expanding tDof.
+    Array<double> ale_mesh_velocity;
+
```
mrp089, line 1619: remove verbose comments

Code/Source/solver/eq_assem.cpp
```diff
@@ -162,6 +171,9 @@ void b_assem_neu_bc(ComMod& com_mod, const faceType& lFa, const Vector<double>&
 /// @param Dg 
 void b_neu_folw_p(ComMod& com_mod, const bcType& lBc, const faceType& lFa, const Vector<double>& hg, const SolutionStates& solutions)
 {
+  // Local alias for old displacement
+  const auto& Do = solutions.old.get_displacement();
```
mrp089, line 175: unused?

Code/Source/solver/eq_assem.cpp
```diff
@@ -296,6 +308,10 @@ void b_neu_folw_p(ComMod& com_mod, const bcType& lBc, const faceType& lFa, const
 /// an arbitrary vector.
 void fsi_ls_upd(ComMod& com_mod, const bcType& lBc, const faceType& lFa, const SolutionStates& solutions)
 {
+  // Local aliases for displacement arrays
+  const auto& Dn = solutions.current.get_displacement();
+  const auto& Do = solutions.old.get_displacement();
```
mrp089, line 313: unused?

Code/Source/solver/fluid.cpp
```diff
@@ -51,7 +51,7 @@ void b_fluid(ComMod& com_mod, const int eNoN, const double w, const Vector<doubl
 
   Vector<double> u(nsd);
 
-  if (com_mod.mvMsh) {
+  if (com_mod.mvMsh || com_mod.ale_mesh_velocity.size() > 0) {
```
mrp089, line 54: why can't we use the existing `mvMsh` flag?

Code/Source/solver/Integrator.cpp
```diff
@@ -853,10 +943,42 @@ void Integrator::corrector()
...
+  #endif
+
+  // Update solution and check convergence
+  update_solution();
+
+  // Check if all equations converged - if so, skip equation cycling
   auto& eqs = com_mod.eq;
   if (std::count_if(eqs.begin(),eqs.end(),[](eqType& eq){return eq.ok;}) == eqs.size()) {
-    #ifdef debug_corrector
+    #ifdef debug_corrector_cycling
```
mrp089, line 1099: don't rename

Code/Source/solver/Integrator.cpp
```diff
@@ -853,10 +943,42 @@ void Integrator::corrector()
...
+
+//------------------------
+// corrector
+//------------------------
+/// @brief Corrector with convergence check and equation cycling
+///
+/// Calls update_solution() to update the solution and check convergence,
+/// then handles equation switching for coupled problems (modifies cEq global).
+//
+void Integrator::corrector()
```
mrp089, line 1074: why do we need this extra function?

Code/Source/solver/Integrator.cpp
```diff
@@ -171,6 +172,94 @@ bool Integrator::step() {
   } // End of Newton iteration loop
 }
 
+//------------------------
+// step_equation
+//------------------------
+/// @brief Solve a single equation to convergence without cycling to other equations
+bool Integrator::step_equation(int iEq, std::function<void()> post_assembly) {
```
mrp089: is this function called anywhere besides partitioned fsi?

Code/Source/solver/Integrator.cpp
```diff
@@ -637,13 +728,12 @@ void Integrator::initiator(SolutionStates& solutions)
...
 ///   com_mod.pSn
-///
-///   com_mod.cEq
 ///   eq.FSILS.RI.iNorm
 ///   eq.pNorm
+///   eq.ok
 /// \endcode
 //
-void Integrator::corrector()
+void Integrator::update_solution()
```
mrp089, line 854: why the renaming?

Code/Source/solver/Integrator.cpp
```diff
@@ -621,11 +710,13 @@ void Integrator::initiator(SolutionStates& solutions)
   }
 }
 //------------------------
-// corrector
```
mrp089, line 624: undo comment changes

Code/Source/solver/main.cpp
```diff
@@ -655,6 +648,9 @@ int main(int argc, char *argv[])
       add_eq_linear_algebra(simulation->com_mod, eq);
     }
 
+    // Initialize partitioned FSI coupling if configured
+    simulation->initialize_partitioned_fsi(file_name);
```
mrp089, line 654: can we do partitioned fsi without changes to `main.cpp`?

Code/Source/solver/mesh.cpp
```diff
@@ -45,9 +45,12 @@ void construct_mesh(ComMod& com_mod, CepMod& cep_mod, const mshType& lM, const S
...
 
-  // Start and end DOF
-  int is = nsd + 1;
-  int ie = 2*nsd;
+  // Start and end DOF for mesh equation in the global solution arrays.
+  // Use eq.s (DOF offset) instead of hardcoded nsd+1, so this works both
+  // in monolithic FSI (where mesh is the 3rd equation) and in partitioned
+  // FSI (where mesh is the only equation with eq.s=0).
+  int is = eq.s;
+  int ie = eq.s + nsd - 1;
```
mrp089, line 53: what do these numbers mean and does this work for all physics?

Code/Source/solver/Parameters.cpp
```diff
@@ -2926,6 +2931,50 @@ void ProjectionParameters::set_values(tinyxml2::XMLElement* xml_elem)
...
+  bool required = true;
+
+  set_parameter("Max_coupling_iterations", 50, !required, max_coupling_iterations);
+  set_parameter("Coupling_tolerance", 1e-6, !required, coupling_tolerance);
+  set_parameter("Initial_relaxation", 1.0, !required, initial_relaxation);
+  set_parameter("Omega_max", 1.0, !required, omega_max);
+  set_parameter("Coupling_method", "aitken", !required, coupling_method);
+  set_parameter("Fluid_interface_face", "", required, fluid_interface_face);
+  set_parameter("Solid_interface_face", "", required, solid_interface_face);
+
```
mrp089, line 3083: remove `_face`

Code/Source/solver/Parameters.cpp
```diff
@@ -2926,6 +2931,50 @@ void ProjectionParameters::set_values(tinyxml2::XMLElement* xml_elem)
...
+const std::string PartitionedCouplingParameters::xml_element_name_ = "Partitioned_coupling";
+
+PartitionedCouplingParameters::PartitionedCouplingParameters()
+{
+  bool required = true;
+
+  set_parameter("Max_coupling_iterations", 50, !required, max_coupling_iterations);
+  set_parameter("Coupling_tolerance", 1e-6, !required, coupling_tolerance);
+  set_parameter("Initial_relaxation", 1.0, !required, initial_relaxation);
+  set_parameter("Omega_max", 1.0, !required, omega_max);
```
mrp089, line 3079: consistent notation: omega vs. relaxation

Code/Source/solver/Parameters.h
```diff
@@ -1508,6 +1508,7 @@ class EquationParameters : public ParameterLists
     Parameter<double> tolerance;
 
     Parameter<std::string> type;
+    Parameter<std::string> role;  // "partitioned_fluid", "partitioned_solid", or "partitioned_mesh"
```
mrp089, line 1640: better name than role?

Code/Source/solver/PartitionedFSI.cpp
```diff
@@ -0,0 +1,1104 @@
...
+
+#include <cmath>
+#include <cstdio>
+#include <iostream>
+#include <set>
+#include <stdexcept>
+#include <unordered_map>
+
+/// Check if any value in the solution arrays is NaN
+static bool has_nan(const SolutionStates& sol) {
```
mrp089, line 25: is this not done elsewhere in the code?

Code/Source/solver/PartitionedFSI.cpp
```diff
@@ -0,0 +1,1104 @@
...
+static void init_sub_sim(Simulation* sim, const std::string& xml_path)
+{
+  read_files_ns::read_files(sim, xml_path);
+  sim->logger.set_cout_write(false);
+
+  // The mesh sub-sim includes a <Partitioned_coupling> stub so that read_files
+  // accepts the mesh equation, but this sim has only the mesh equation (tDof=3).
+  // baf_ini assumes FSI DOF layout (Do(i+nsd+1,Ac) reaches row 4) when mvMsh=true,
+  // which would go out-of-bounds for the standalone mesh sub-sim.
+  if (sim->com_mod.nEq == 1 &&
```
mrp089, line 52: there should be a general documentation (e.g., in the header with doxygen) on what the fields are, what DOFs they have, and what information is being exchanged

Code/Source/solver/PartitionedFSI.cpp
```diff
@@ -0,0 +1,1104 @@
...
+  // so each sub-sim writes to distinct files (result_fluid_*, result_solid_*, result_mesh_*).
+  XMLElement* gen = root->FirstChildElement("GeneralSimulationParameters");
+  if (gen) {
+    XMLElement* gen_clone = gen->DeepClone(&sub)->ToElement();
+    // Map role → short suffix used in the prefix name
+    std::string suffix;
+    if      (role == "partitioned_fluid") suffix = "fluid";
+    else if (role == "partitioned_solid") suffix = "solid";
+    else if (role == "partitioned_mesh")  suffix = "mesh";
+    if (!suffix.empty()) {
```
mrp089, line 134: this shouldn't be done with strings

Code/Source/solver/PartitionedFSI.cpp
```diff
@@ -0,0 +1,1104 @@
...
+                               const PartitionedFSIConfig& config,
+                               const std::string& xml_file_path)
+  : main_sim_(main_simulation), config_(config),
+    xml_file_path_(xml_file_path), omega_(config.initial_relaxation)
+{
+  auto& cm = main_sim_->com_mod.cm;
+  auto& cm_mod = main_sim_->cm_mod;
+
+  // Compute deterministic temp XML paths (same formula as in build_sub_xml).
+  auto make_path = [&](const std::string& role) -> std::string {
```
mrp089: why do we need to write temp XMLs?

Code/Source/solver/PartitionedFSI.cpp
```diff
@@ -0,0 +1,1104 @@
...
+    offset += recv_counts[p];
+  }
+}
+
+//----------------------------------------------------------------------
+// build_face_node_map — match each LOCAL face_a node to its nearest
+// node in the PRE-GATHERED global face_b coordinates.
+// Returns local_src_idx → global_tgt_idx map.
+//----------------------------------------------------------------------
+void PartitionedFSI::build_face_node_map(
```
mrp089, line 361: there should already be code to do that in svMP

Code/Source/solver/PartitionedFSI.cpp
```diff
@@ -0,0 +1,1104 @@
...
+    global_nNo += all_nNo[p];
+  }
+}
+
+//----------------------------------------------------------------------
+// gather_face_data — MPI_Allgatherv local face data to all ranks.
+// local_data is (nrows, local_nNo); returns (nrows, global_nNo).
+// Rank ordering in global array matches the local_offset convention.
+//----------------------------------------------------------------------
+Array<double> PartitionedFSI::gather_face_data(
```
mrp089, line 284: there should already be code to do that in svMP

Code/Source/solver/PartitionedFSI.cpp
```diff
@@ -0,0 +1,1104 @@
...
+    node_offset += p_nNo;
+  }
+  return result;
+}
+
+//----------------------------------------------------------------------
+// gather_global_map — all-gather a local (local_src → global_tgt) map
+// into a global (global_src → global_tgt) map.
+//----------------------------------------------------------------------
+void PartitionedFSI::gather_global_map(
```
mrp089, line 327: there should already be code to do that in svMP

Code/Source/solver/PartitionedFSI.cpp
```diff
@@ -0,0 +1,1104 @@
...
+              << " dups" << std::endl;
+  }
+}
+
+//----------------------------------------------------------------------
+// build_face_dedup_map — canonical[i] = first global slot index that
+// shares the same physical node (global node ID) as slot i.
+// Slots with no duplicate have canonical[i] == i.
+//----------------------------------------------------------------------
+std::vector<int> PartitionedFSI::build_face_dedup_map(
```
mrp089, line 530: what does this do?

Code/Source/solver/PartitionedFSI.cpp
```diff
@@ -0,0 +1,1104 @@
...
+      for (int row = 0; row < arr.nrows(); row++)
+        arr(row, i) = arr(row, canonical[i]);
+}
+
+//----------------------------------------------------------------------
+// build_face_owner_flags — flags[a] = 1 if THIS rank is the lowest-ID rank
+// that contains local face node a, else 0.  A node shared across ranks
+// (a partition-ring node) is owned by exactly one rank under this rule.
+//----------------------------------------------------------------------
+std::vector<int> PartitionedFSI::build_face_owner_flags(
```
mrp089, line 573: i don't understand why we need this

Code/Source/solver/post.cpp
```diff
@@ -2088,4 +2089,148 @@ void tpost(Simulation* simulation, const mshType& lM, const int m, Array<double>
...
 
+//----------------------------------------------------------------------
+// compute_face_traction — consistent nodal forces at a fluid face
+//----------------------------------------------------------------------
+// Computes f(i,a) = integral(sigma_ij * n_j * N_a dGamma) at each face
+// node.  Returns the force that the fluid exerts ON the solid.
+// Adapts the stress computation from bpost() but integrates over the
+// face using face Gauss quadrature.
+//
+Array<double> compute_face_traction(
```
mrp089: can this be done in bpost?

Code/Source/solver/set_bc.cpp
```diff
@@ -1109,6 +1109,9 @@ void set_bc_dir_l(ComMod& com_mod, const bcType& lBc, const faceType& lFa, Array
 //
 void set_bc_dir_w(ComMod& com_mod, const SolutionStates& solutions)
 {
+  // Local alias for old displacement
+  const auto& Do = solutions.old.get_displacement();
```
mrp089, line 1189: unused?

Code/Source/solver/set_bc.cpp
```diff
@@ -1991,6 +2003,71 @@ void set_bc_undef_neu_l(ComMod& com_mod, const bcType& lBc, const faceType& lFa)
   }
 }
 
+//----------------------------------------------------------------------
+// enforce_dirichlet_on_face
+//----------------------------------------------------------------------
+void enforce_dirichlet_on_face(ComMod& com_mod, const faceType& lFa, int nsd)
```
mrp089, line 2119: what's the difference between `enforce_dirichlet_on_face` and `enforce_dirichlet_dofs_on_face`? isn't this functionality already in svMP?

Code/Source/solver/Simulation.cpp
```diff
@@ -112,3 +124,81 @@ Integrator& Simulation::get_integrator()
...
+  bcast_str(config.solid_interface_face);
+
+  config.max_coupling_iterations = max_iter;
+  config.coupling_tolerance       = tol;
+  config.initial_relaxation       = relax;
+  config.omega_max                = omax;
+  config.coupling_method          = static_cast<CouplingMethod>(method_i);
+
+  partitioned_fsi_ = std::make_unique<PartitionedFSI>(this, config, xml_file_path);
+}
```
mrp089, line 211: i don't think these `MPI_Bcast` should be in here

Code/Source/solver/Simulation.cpp
```diff
@@ -112,3 +124,81 @@ Integrator& Simulation::get_integrator()
...
+PartitionedFSI* Simulation::get_partitioned_fsi()
+{
+  return partitioned_fsi_.get();
+}
+
+/// @brief Initialize partitioned FSI if configured in parameters.
+///
+/// Parameters are only parsed on rank 0 (slaves skip read_files), so we
+/// broadcast the active flag and config to all ranks before branching.
+void Simulation::initialize_partitioned_fsi(const std::string& xml_file_path)
```
mrp089, line 145: i think until here partitioned fsi should be handled like any other physics type and not with its own function in `Simulation`

mrp089 (changes requested):
Thank you for your work on this, @Eleven7825! Happy this is working now. See my comments below. My main concerns are:
1. Parallelization strategy
2. Modification of central parts (`main`, `Simulation`, `Intergrator`)
3. Potentially duplicated code

Curious to hear what feedback @zasexton has.

mrp089:
@claude review

Code/Source/solver/PartitionedFSI.cpp
Eleven7825: We need them because current `Simulation` initialization pipeline can only handle one xml at a time(read_files.cpp). Partitioned fsi needs three sub `Simulation` instances. If we want to do it without temp xml, we need to refactor the current read_files.cpp substantially(could be done in another PR).
Morever, the unit test you created back in April needs solid specific checks so it needs to have the capacity to run solid simulation only for test purposes.

Code/Source/solver/main.cpp
Eleven7825: Not quite. But it should be an easy refactor:
`iterate_solution`  (in `main.cpp`) contains the main time stepping. Currently, `iterate_solution` assumes we only need to solve nonlinear problems once. I propose creating a polymorphic abstraction for `iterate_solution`. If we want to implement load steps in G&R problem, we have an entry point here.

Code/Source/solver/Parameters.cpp
```diff
@@ -222,14 +222,29 @@ void Parameters::print_parameters()
...
 /// @brief Set the simulation parameter values given in an XML format file.
 void Parameters::read_xml(std::string file_name)
-{ 
+{
   tinyxml2::XMLDocument doc;
-  
   auto error = doc.LoadFile(file_name.c_str());
-  
+  set_values_from_doc(doc, error, "XML file '" + file_name + "'");
+}
```
ktbolt, line 234: @Eleven7825 Why are you passing a failed file load error to a function ?

Code/Source/solver/PartitionedFSI.cpp
```diff
@@ -191,35 +190,18 @@ PartitionedFSI::PartitionedFSI(Simulation* main_simulation,
...
+  // initialize a standalone sub-simulation from it. No temp files are written:
+  // every rank parses the same main XML (read-only) and passes the resulting
+  // sub-XML string straight to read_files, so no rank-0 write / barrier is needed.
   fluid_sim_ = std::make_unique<Simulation>();
-  init_sub_sim(fluid_sim_.get(), fluid_xml);
+  init_sub_sim(fluid_sim_.get(), build_sub_xml(xml_file_path_, "partitioned_fluid"));
 
   solid_sim_ = std::make_unique<Simulation>();
-  init_sub_sim(solid_sim_.get(), solid_xml);
+  init_sub_sim(solid_sim_.get(), build_sub_xml(xml_file_path_, "partitioned_solid"));
```
ktbolt, line 202: @Eleven7825 Don't use strings to identify different cases; create an enum class for this.

Code/Source/solver/read_files.cpp
```diff
@@ -1663,12 +1663,18 @@ void read_files(Simulation* simulation, const std::string& file_name)
...
-  dmsg << "Read the solver XML file " << " ... ";
+  dmsg << "Read the solver XML " << " ... ";
   #endif
   if (!com_mod.resetSim) {
-    simulation->read_parameters(std::string(file_name));
+    if (from_string) {
+      simulation->read_parameters_from_string(file_name);
+    } else {
+      simulation->read_parameters(std::string(file_name));
+    }
```
ktbolt, line 1855: @Eleven7825 The function `read_parameters_from_string` uses a variable named `file_name` so is this reading from a file or from a string ?

Code/Source/solver/Simulation.cpp
```diff
@@ -52,6 +52,13 @@ void Simulation::read_parameters(const std::string& file_name)
   parameters.read_xml(file_name);
 }
 
+/// @brief Read solver parameters from an in-memory XML string (no file).
+//
```
ktbolt, line 56: @Eleven7825 You don't need to write `in-memory`; strings are in variables stored in memory.

Code/Source/solver/post.cpp
```diff
@@ -2088,4 +2089,148 @@ void tpost(Simulation* simulation, const mshType& lM, const int m, Array<double>
...
+
+      auto N = lFa.N.col(gf);
+      for (int a = 0; a < lFa.eNoN; a++) {
+        int Ac = lFa.IEN(a, e);
+        int a_local = global_to_face[Ac];
+        for (int i = 0; i < nsd; i++)
+          result(i, a_local) -= w * N(a) * trac(i);
+      }
+    }
+  }
```
zasexton: From reading this it seems that the approach does not evaluate the fluid stress at the interface quadrature points? Instead the stress is averaged in `sigma_avg` from the volume quadrature and applied at face Gauss points. Is this the intended method? If so, I do not believe this will give the correct Neumann load condition, generally.

Code/Source/solver/Integrator.cpp
```diff
@@ -853,10 +943,42 @@ void Integrator::corrector()
     dmsg << "com_mod.eq[1].ok: " << com_mod.eq[1].ok;
     #endif
   }
```
zasexton, line 1063: Can we double-check the success conditions here? if the equation reaches the maximum number of nonlinear iterations without reaching the desired residual tolerance shouldn't this return `false`? From what I'm reading if the equation reaches the max number of iterations then it it satisfies the success condition regardless of meeting the tolerance...

Perhaps something like this:
```c++
  bool converged = ((l2 || l3) && l4);
  bool reached_max_iterations = l1;

  if (converged) {
    eq.ok = true;
  #ifdef debug_corrector
    dmsg << "eq.ok: " << eq.ok;
    dmsg << "com_mod.eq[0].ok: " << com_mod.eq[0].ok;
    dmsg << "com_mod.eq[1].ok: " << com_mod.eq[1].ok;
  #endif
  } else if (reached_max_iterations) {
    eq.ok = false;
  #ifdef debug_corrector
    dmsg << "Maximum nonlinear iterations reached without convergence.";
    dmsg << "eq.ok: " << eq.ok;
  #endif
  }
```

Code/Source/solver/PartitionedFSI.cpp
```diff
@@ -0,0 +1,1082 @@
...
+    for (int i = 0; i < nrows; i++)
+      local_solid_traction(i, a) =
+          w * global_solid_traction(i, solid_face_local_offset_ + a);
+  }
+
+  set_bc::set_bc_dir(solid_com, solid_sol);
+  solid_int.step_equation(0, [&]() {
+    fsi_coupling::apply_traction_on_solid(
+        solid_com, solid_com.eq[0], *solid_face_, local_solid_traction);
+  });
```
zasexton, line 976: Why is this method not leveraging the `true`/`false` convergence condition that it returns? 

```c++
bool step_equation(int iEq, std::function<void()> post_assembly = nullptr);
```

Does this mean that we are only checking for `NaN` values and not solution convergence criteria?

Code/Source/solver/PartitionedFSI.cpp
```diff
@@ -0,0 +1,1082 @@
...
+  for (int a = 0; a < mesh_face_->nNo; a++)
+    for (int i = 0; i < nsd; i++)
+      local_mesh_disp(i, a) = global_mesh_disp(i, mesh_face_local_offset_ + a);
+
+  set_bc::set_bc_dir(mesh_com, mesh_sol);
+  fsi_coupling::apply_displacement_on_mesh(
+      mesh_com, mesh_com.eq[0], *mesh_face_, local_mesh_disp, mesh_sol);
+  mesh_int.step_equation(0, [&]() {
+    set_bc::enforce_dirichlet_on_face(mesh_com, *mesh_face_, nsd);
+  });
```
zasexton, line 1008: Similarly here we are not checking for equation convergence from the `step_equation()` method but instead the existence of `NaN` values...

Code/Source/solver/PartitionedFSI.cpp
```diff
@@ -0,0 +1,1082 @@
...
+  double af = fluid_com.eq[0].af;
+  fluid_com.ale_mesh_velocity.resize(nsd, fluid_com.tnNo);
+  for (int a = 0; a < fluid_com.tnNo; a++)
+    for (int i = 0; i < nsd; i++)
+      fluid_com.ale_mesh_velocity(i, a) = (1.0 - af) * mesh_vel_Yo(i, a)
+                                         + af * mesh_vel_Yn(i, a);
+
+  fluid_int.step_equation(0, [&]() {
+    set_bc::enforce_dirichlet_dofs_on_face(fluid_com, *fluid_face_, 0, nsd);
+  });
```
zasexton: use `step_equation()` success boolean argument returned to gauge success? why are we switching to just `NaN` searches?

Code/Source/solver/Integrator.cpp
```diff
@@ -171,6 +172,94 @@ bool Integrator::step() {
...
+    if (eq.ok) {
+      return true;
+    }
+
+    // Abort on NaN in residual norm (indicates divergence).
+    if (newton_count_ > 1 && std::isnan(eq.FSILS.RI.iNorm)) {
+      return false;
+    }
+
+    output::output_result(simulation_, com_mod.timeP, 2, iEq);
```
zasexton, line 274: We should double-check that if the max iterations are reached but the tolerance is not met that we appropriately return `false`. This series of return statements leads be to believe that this method will only ever return `true`?

Code/Source/solver/post.cpp
michelebucelli: My past experience with evaluating the fluid stress at the interface quadrature points is that it's very (very!) unstable (as in, energy builds up over time and the system explodes at some point).

My understanding of this is that the derivatives at the interface are elements of $H^{-\frac{1}{2}}(\partial\Omega)$, so that the expected convergence rate is very bad, and treating them as regular functions (i.e. stuff that you can evaluate at points) is problematic.

The approach that I think is correct is to evaluate the interface traction as the residual of the equation when the interface conditions are dropped (not sure if this is what is happening in this code though). See [here](https://doi.org/10.4208/cicp.OA-2021-0243) and the references that it cites for more details.

zasexton:
I ran this partitioned solve and compared it against the monolithic FSI solve for the first 20 timesteps ($\delta t = 0.01$). Result averages for comparisons are taken at a slice from the middle slice. 
mid_slice_velocity_magnitude
I see that fluid velocity has the best agreement (usually with a relative difference within 1.6%). 
mid_slice_solid_displacement_magnitude
mid_slice_pressure
However, pressure and wall deformation differ with max relative differences of about 49% and 55%, respectively. I believe this may be related to the coupling residual relying on kinematic continuity being enforced better than traction consistency. This may be related to earlier comments made about stress averaging at the interface points. With such differences emerging early in the simulation, it suggests that partitioned results (while numerically "stable") will diverge from the monolithic FSI solution. 

Full simulation files are given here:
Monolithic:
[monolithic.zip](https://github.com/user-attachments/files/29557813/monolithic.zip)

Partitioned:
[partitioned.zip](https://github.com/user-attachments/files/29557835/partitioned.zip)

zasexton (changes requested):

Eleven7825:
@zasexton, thank you for testing the partitioned FSI implementation. The discrepancy between the monolithic and partitioned results is expected. While these differences could theoretically be mitigated by using a smaller time step, we have found that utilizing a time step of $dt = 10^{-4}$ in the partitioned case triggers added-mass instabilities. In those instances, the coupling iterations fail to reach the target tolerance, and we observed significant result variations across different processor counts—to the point that the integration tests fail to pass. For these reasons, we deliberately chose $dt = 10^{-2}$ for the current test cases. While this is not physically representative, it provides the necessary numerical stability to validate the code and ensure the framework is functional for testing purposes.

Code/Source/solver/post.cpp
zasexton: I agree. Thank you for this reference. Even still, the current element-volume average approach does not seem to be supported. 

I agree with your cited approach: 

  $$
  \sigma_f(u,p)n = \sigma_s(d)n \quad \text{on } \Sigma
  $$

  With the discrete formulation it imposes this condition in residual form. In Appendix A, the dynamic coupling operator $C_d$ includes the fluid interface residual terms:

  $$
  \frac{\rho_f}{\Delta t} M^f_{\Sigma f}(V_f^{n+1}-V_f^n)
  +
  \frac{\rho_f}{\Delta t} M^f_{\Sigma\Sigma}(V_\Sigma^{n+1}-V_\Sigma^n)
  +
  C_{\Sigma f}(V^n,G^{n+1},G^n)V_f^{n+1}
  +
  C_{\Sigma\Sigma}(V^n,G^{n+1},G^n)V_\Sigma^{n+1}
  +
  K^f_{\Sigma f}U_f^{n+1}
  +
  K^f_{\Sigma\Sigma}U_\Sigma^{n+1}
  +
  B_\Sigma^T P^{n+1}.
  $$

The structural interface residual terms:

  $$
  \frac{\rho_s}{\Delta t} M^s_{\Sigma s}
  \left(D_s^{n+1} - 2D_s^n + D_s^{n-1}\right)
  +
  \frac{\rho_s}{\Delta t} M^s_{\Sigma\Sigma}
  \left(D_\Sigma^{n+1} - 2D_\Sigma^n + D_\Sigma^{n-1}\right)
  +
  K^s_\Sigma(D^{n+1}).
  $$


From my understanding this supports recovering the fluid force from the unconstrained assembled fluid residual on the interface momentum DOFs, before the Dirichlet rows are zeroed/overwritten, rather than by evaluating or averaging \($\sigma_f n\$) at face quadrature points. Is this a fair understanding @michelebucelli ? 

Ultimately if this element-average approximation is "good-enough" for the purposes of the partitioned FSI then I have no issues, but we should definitely advertise such limitations.

alisonmarsden:
Assuming you're planning to deal with the numerical instabilities related to added mass effects in the partitioned scheme via Aitkin relaxation as mentioned above - just make sure you have thought this through thoroughly as there are many known issues that crop off with these schemes.

Code/Source/solver/post.cpp
michelebucelli: > Is this a fair understanding @michelebucelli ?

Yes, I think it is.

One way you can see this is by taking the Neumann contribution to the weak form of the solid subproblem, which looks something like

$$
N(v; u, p) = \int_\Sigma \sigma_f(u, p)n \cdot v ds
$$

and observing that, if $u$ and $p$ are solutions to the fluid equation, it will satisfy the fluid weak formulation:

$$ \begin{gathered}
F(u, p; v, q) + N(v; u, p) = 0 \\
N(v; u, p) = -F(u, p; v, q)
\end{gathered} $$

where $F$ contains all the other terms (time derivative, advection, viscosity, volume forces). You use the last equation above to compute the terms to add to the structure equation.

One key aspect is that, in the weak formulation of the fluid problem, the term $N(v; u, p)$ wouldn't really appear, because you're imposing Dirichlet conditions on the interface, so that $v = 0$ on $\Sigma$. That is why you need to compute this residual **before** applying Dirichlet interface conditions, that would make that part of the residual go away.

This has better accuracy properties than evaluating the traction point-wise because you're looking at the boundary traction not as a function (i.e. something that you can evaluate on points of $\mathbb{R}^n$) but as a function**al** (i.e. an operator between function spaces) that maps $v$ into $N(v; u, p)$, which is the more mathematically correct point of view.

I think averaging might be slightly better than pointwise evaluation, purely based on intuition (meaning: I might be wrong), but I haven't thought about it too deeply.

zasexton:
Additionally, is there any timing comparison against the monolithic solver (or is this a factor that we need to consider for implementing this approach)? When I was running the 20 time step runs on 1 processor the partitioned approach was taking nearly 90x longer per time step when reviewing the log files.

Code/Source/solver/post.cpp
Eleven7825: Thanks @michelebucelli — this is really interesting. I agree that for the coupling load itself we don't need the traction at quadrature points. The reason I still want a pointwise interface traction is our planned Fluid-Solid Growth (FSG) work: there, the wall shear stress/interface traction acts as a mechanobiological growth stimulus for the solid, so we need it pointwise. 

For this PR, I'd propose keeping the current volume-averaged traction while clearly documenting its limitations, since the goal here is a validated partitioned framework as the foundation for FSG. The residual-based traction is a natural refinement we could move to in the future if we need higher coupling accuracy. Does that sound reasonable, @michelebucelli @zasexton?

zasexton:
In further reviewing this issue/PR I have checked out the branch for this pull request and made some additional changes to address differences between the partitioned and monolithic results for the CI test case. At a high level I will summarize some of the changes that have been made and their implications toward matching the monolithic solution profiles:

  1. Partitioned solid loading now uses recovered fluid residual force instead of postprocessed face traction. See `Code/Source/solver/PartitionedFSI.cpp:897`, `Code/Source/solver/PartitionedFSI.cpp:923`, and `Code/Source/solver/fsi_coupling.cpp:80`. The old `post::compute_face_traction` path was removed from `post.cpp/post.h`. Thanks to the conversation with @michelebucelli for pointing in this direction. 

  2. The fluid residual can now be reassembled without solving, specifically to recover the interface force through the normal assembly path. Specifically so we can avoid using a post processing route. See `Code/Source/solver/Integrator.cpp:283`.

  3. Fluid mesh coordinates now use generalized-alpha staged ALE geometry during residual assembly, then return to full `n+1` geometry for output/next-step state. See `Code/Source/solver/PartitionedFSI.cpp:1011`, `Code/Source/solver/PartitionedFSI.cpp:1205`, and `Code/Source/solver/fsi_coupling.cpp:159`.

  4. Interface ALE velocity is explicitly synchronized to the staged fluid interface velocity, matching the monolithic residual time integration state behavior. See `Code/Source/solver/PartitionedFSI.cpp:845`.

  5. Boundary-condition assembly can temporarily use old fluid geometry, matching the monolithic face-Neumann `gnnb()` behavior. See `Code/Source/solver/PartitionedFSI.cpp:869` and `Code/Source/solver/Integrator.cpp:180`.

  6. The partitioned solid subproblem now copies the FSI/fluid generalized-alpha parameters, avoiding a monolithic-vs-partitioned time integration mismatch that was originally occurring. See `Code/Source/solver/PartitionedFSI.cpp:203` and `Code/Source/solver/fsi_coupling.cpp:36`.

  7. Interface force components constrained by strong solid Dirichlet BCs are masked before applying the recovered force. See `Code/Source/solver/PartitionedFSI.cpp:468`.

With these changes we now observe much better agreement across averaged pressure, velocity, and solid deformation at the mid slice of the fsi test case between partitioned and monolithic solves across 20 time steps. Max relative mid-slice differences were pressure `1.283e-2`, velocity magnitude `1.500e-3`, and solid deformation magnitude `1.570e-2`

I will proceed to open a pull request with these changes against your branch `feature/issue-431-partitioned-fsi` @Eleven7825 . Then you can decide which, if any, of these you and @mrp089 think are worth keeping. 

mid_slice_comparison_all

Monolithic Result Files:
[monolithic.zip](https://github.com/user-attachments/files/29614776/monolithic.zip)


Partitioned Result Files:
[partitioned.zip](https://github.com/user-attachments/files/29614778/partitioned.zip)


I still have remaining questions about the overall compute time that these partitioned fsi simulations take, but we can discuss these at another point. I also agree with @ktbolt that we should look to squash all of these commits to make review of this work easier. This will be especially important if changes are pulled into your local branch `feature/issue-431-partitioned-fsi` prior to merging with the production code `main`

zasexton:
Branch reference: [`zasexton:pr-554-partitioned-fsi`](https://github.com/zasexton/svMultiPhysics/tree/pr-554-partitioned-fsi)

Corresponding PR: [`feature/issue-431-partitioned-fsi` <- `pr-554-partitioned-fsi`](https://github.com/Eleven7825/svMultiPhysics/pull/1/commits)

mrp089:
Thank you so much for digging into that, @zasexton! We'll have a look at your branch. Do you also have input on refactoring the time loop (#576) that's required for this PR?

Eleven7825:
@zasexton Thank you for your work on this. After discussing it with @mrp089, we agree that the residual approach is a significant improvement for the partitioned FSI and yields much more accurate results. We can reimplement the traction calculation in a later pull request, as it is needed for our Growth and Remodeling (G&R) implementation.

PR #561, opened 2026-06-02
Issue 525 FE Basis refactor implementation

### LINKED ISSUE #525: Implementing Basis Module
### Problem

In the solver, basis functions are hard-coded through a Fortran-style set of per-element dispatch tables, cached directly inside mesh and function-space data structures, and then copied, recomputed, or patched by special-case logic whenever we need mixed spaces or NURBS behavior. That makes basis evaluation tightly coupled to mesh storage, quadrature setup, and physics assembly. Adding new basis families or extending derivative support tends to spread changes across unrelated parts of the solver. A modular Basis folder therefore is necessary because it gives us a single mesh-agnostic interface for basis families, plus centralized construction, caching, and evaluation machinery, so different basis families (Lagrange, Bernstein, B-spline/NURBS, and vector bases) can be implemented once, tested once, and reused consistently across elements, geometry mappings, and assembly kernels instead of proliferating element-specific and physics-specific code paths.

  - Basis evaluation is implemented as a monolithic Fortran-port dispatch layer. `Code/Source/solver/nn.cpp` pulls in large element tables, and `Code/Source/solver/nn_elem_gnn.h` hard-codes closed-form `N`/`Nx` formulas per element type instead of exposing a reusable basis abstraction.
  - Basis state is stored directly on mesh and function-space containers rather than encapsulated in its own module. `Code/Source/solver/ComMod.h` and `Code/Source/solver/ComMod.h` embed `N`, `Nx`, `Nxx`, bounds, and quadrature data into `fsType` and `mshType`, while `Code/Source/solver/nn.cpp` and `Code/Source/solver/fs.cpp` allocate and populate those arrays procedurally.
  - Mixed spaces and extensibility are handled by special cases. `Code/Source/solver/fs.cpp` and `Code/Source/solver/fs.cpp` build Taylor-Hood spaces by copying or regenerating alternate shape tables, while NURBS support is still scattered as repeated update hooks or TODOs in places like `Code/Source/solver/fluid.cpp`, `Code/Source/solver/eq_assem.cpp`, and `Code/Source/solver/nn.cpp`.

### Solution

To address the current solver’s basis-function, we should move all reference-element basis logic into a dedicated, mesh-agnostic `Basis` module with a common interface for evaluating values, gradients, and higher derivatives, and create those bases through a factory keyed by element family, order, continuity, and field type. Instead of storing handwritten `N`, `Nx`, and `Nxx` tables directly on `mshType` and `fsType` and rebuilding them through scattered element-specific branches, the mesh and element layers should hold lightweight basis and quadrature objects, while geometry mapping handles only the transformation from reference to physical space. Mixed discretizations such as Taylor-Hood should be represented as explicit mixed/composite elements rather than ad hoc control flow, and any future implemented NURBS/IGA support should be implemented as another basis family rather than a repeated special case. With caching and batched evaluation centralized in the basis layer, the same implementations can be reused consistently by physics assembly kernels, which reduces duplication, makes new basis families easier to add, and localizes both testing and performance optimization to a single part of the codebase.

This issue will outline `Basis` module design and API/ABI infrastructure.

zasexton:
## Current situation
  Tracks #525.

  The solver currently relies on legacy table-driven shape-function paths in `nn.cpp`, which makes basis evaluation, Hessian support, node-ordering validation, and parity testing difficult to extend. This PR introduces a self-contained FE Basis layer while preserving the existing solver-facing storage contracts.

  ## Release Notes
  - Added a new FE Basis module with Lagrange and Serendipity basis support for mapped solver elements, including values, gradients, Hessians, node-ordering conventions, factory creation, and cache support.
  - Added FE math and quadrature utilities used by the new basis implementation.
  - Routed supported `nn::get_gnn` and `nn::get_gn_nxx` volume/face evaluations through the new FE Basis adapter.
  - Added `SVMP_BASIS_MODE=auto|legacy|fe` to compare legacy and FE Basis evaluation paths at runtime.
  - Added typed FE/basis exception paths for unsupported elements, invalid configurations, and backend failures.
  - Added a basis comparison harness for running pytest integration cases under legacy and FE modes and comparing outputs, timing, fields, and iteration data.
  - Build migration: solver code now requires a C++20-capable compiler.

  ## Documentation
  - Added Doxygen-style documentation to the new FE Basis, FE Math, and Quadrature APIs.
  - No solver XML migration is expected for users; basis mode selection is controlled through `SVMP_BASIS_MODE`.

  ## Testing
  - Added GoogleTest coverage under `tests/unitTests/FE/Basis` for Lagrange/Serendipity basis evaluation, Hessians, cache/factory behavior, error paths, node ordering, solver adapter parity, and supported mapped element coverage.
  - Added GoogleTest coverage under `tests/unitTests/FE/Math` for matrix/vector operations, expression helpers, math constants, and dense linear algebra.


zasexton:
From the CI test cases we got the following:

  - macOS: 12 failed, 227 passed
  - Ubuntu: 21 failed, 218 passed
  - Failures are all FSI / FSI-ustruct pipe cases.

The affected meshes are `HEX8`. Current code allocated `Nxx` but did not populate `HEX8` second derivatives; this branch now computes nonzero `HEX8` Hessians via FE Basis. FSI uses those in `Code/Source/solver/fsi.cpp:191`, so the stored legacy reference VTUs no longer match.

zasexton:
In the FSI fluid assembly, `Nxx` is passed into the element kernel and used in terms that contribute to the assembled equations, especially VMS/stabilization terms. 

  - Code/Source/solver/fluid.cpp:1841: `Nwxx` is used to build `uxx` at lines 1841-
    1860.

  - Code/Source/solver/fluid.cpp:1880: `uxx` feeds `d2u2`.
  - Code/Source/solver/fluid.cpp:1933: `uxx` also feeds `es_x`, then `mu_x` at lines
    1948-1963.

  - Code/Source/solver/fluid.cpp:2027: `d2u2` and `mu_x` enter `rS`.
  - Code/Source/solver/fluid.cpp:2037: `rS` enters `up`.
  - Code/Source/solver/fluid.cpp:2103: `up`-dependent quantities are used when
    assembling `lR`, the local residual, at lines 2103-2106.

That means:

  - in the current code, it solves a discrete problem where `HEX8` second-derivative contributions are effectively zero or absent when using the old basis functions

  - when using the new FE Basis functions, it solves a different discrete problem where `HEX8` mapped Hessian contributions are present
  
  Therefore the Hessian is not just participating in the tangent assembly and may not necessarily converge on the same solution as the current reference vtu. This is also why we do not see such CI failures for test cases that use Tet elements because `TET4` have zero reference second derivatives or for `TET10` Hessian values are provided. For such elements, introduction of the new Basis function infrastructure did not cause a change in the residual being solved for these cases.

zasexton:
Overall, we see that providing the non-zero Hessian values slightly increases the number of linear iterations required but does not impact the number of nonlinear iterations for the FSI test cases considered. More importantly, this difference in how the Hessian is used in the formulated residual suggested that the problem being solved is slightly different depending on if a user is providing brick elements versus `TET10` (which shouldn't be the case)

convergence-iteration-behavior

zasexton:
Confirmed. The difference in accuracy for the FSI test cases when employing the new FE Basis functions is due to the non-zero Hessian values for the reference elements. We can see that when we zero-out the values for the Hessian on the `HEX8` cases that we recover the same behavior as the current solver. @ktbolt do you think it would make sense to update the reference solutions for these test cases or is there an alternative strategy that we should look at?

hex8-hessian-confirmation-by-case

aabrown100-git:
@zasexton Were Hessians also zeroed out in the Fortran svFSI?

I think we should update reference solutions unless there is a good reason to use zero Hessians. @sujaldave, @hanzhao2020  Is there any reason (in VMS or something) to force $u_{xx}$ to be zero in HEX8 or higher order elements?

zasexton:
@aabrown100-git yes, the Hessian was also effectively zero for the `HEX8` in the original fortran `svFSI`. Only `TET10`, `TRI6`, `QUD8`, `QUD9`, and `LIN2` had cases in the `GETGNNxx` subroutine. All other cases return the zero-initialized Hessian.

Code/Source/solver/FE/Math/DenseLinearAlgebra.h
aabrown100-git, line 1: @zasexton Wondering if this new Math library could be replaced with Eigen, since we already include it as a dependency. Can you talk about the advantages of this library over Eigen?

Code/Source/solver/nn.cpp
```diff
@@ -37,22 +47,358 @@ using namespace consts;
...
+      BasisSelection{fe::ElementType::Quad9,     fe::BasisType::Lagrange,    2},
+      BasisSelection{fe::ElementType::Tetra4,    fe::BasisType::Lagrange,    1},
+      BasisSelection{fe::ElementType::Tetra10,   fe::BasisType::Lagrange,    2},
+      BasisSelection{fe::ElementType::Hex8,      fe::BasisType::Lagrange,    1},
+      BasisSelection{fe::ElementType::Hex20,     fe::BasisType::Serendipity, 2},
+      BasisSelection{fe::ElementType::Hex27,     fe::BasisType::Lagrange,    2},
+      BasisSelection{fe::ElementType::Wedge6,    fe::BasisType::Lagrange,    1},
+  };
+
+  const int index = static_cast<int>(eType) - static_cast<int>(consts::ElementType::LIN1);
```
aabrown100-git: (with help from Codex) This depends on the enum order matching the lookup table. Could we make this an explicit switch or map instead? That would be easier to read and less likely to break if new element types are added later.

Documentation/Doxyfile
aabrown100-git, line 1: I don't see the `Documentation/basis-comparison-plan.md` you mentioned in the PR description.

aabrown100-git (changes requested):
Left some comments!

aabrown100-git:
@zasexton Also, are you planning on adding an independent design document? Something I found helpful was asking Codex to explain the flow from mesh input to basis function generation to mapping to existing `N`, `Nx`, and `Nxx` arrays for TET4 (as an example). Perhaps something like that? It would also help to understand the existing basis function code surrounding this PR. For example, that element types/order are determined from the provided mesh, and that `N`, `Nx`, and `Nxx` are cached and correspond to the parametric element, but basis functions and gradients w.r.t to the physical element in the reference configuration are computed during assembly (I think).

zasexton:
In general, I only assume that we derive the topology from the mesh information. Although often mesh structure can imply interpolation order, this does not always have to be true; I believe the solution basis order may be a separate discretization choice. The geometry mapping basis should always be interpolated from the mesh nodes provided, but the field solution basis functions need not be assumed the same. One practical example could be the Taylor-Hood elements where mixed fields expect different solution field orders (e.g. velocity P2, pressure P1). If we only allow for isoparametric order then these types of mixed order formulations require much more delicate flagging and checking (which I think might be the case currently). So for flow diagram type of illustration, any mesh information would be directly linked to the geometric basis while the field solution basis/orders rely on information from practical formulation choices.

Code/Source/solver/FE/Math/DenseLinearAlgebra.h
zasexton: When I originally started writing this Basis submodule I did so without the assumptions of an Eigen dependency, but I agree. The small matrix/vector math here should be fully migrated to Eigen as we do not want to keep multiple linear algebra infrastructures. I'll migrate these over and remove the Math subfolder.

Documentation/Doxyfile
zasexton: Originally I wanted to include this but instead I opted to just include the summary graphs in the conversation of this draft PR since I thought this was the better place for this information. I'll remove this from the summary.

Code/Source/solver/FE/Math/DenseLinearAlgebra.h
zasexton: wrapped the Vector and Matrix implementations to Eigen implementations in commit [1289c08](https://github.com/SimVascular/svMultiPhysics/pull/561/commits/1289c086f637cdc1544aff0bfe99eb78ad3b9f1c) so now we are explicitly using Eigen only.

Code/Source/solver/nn.cpp
zasexton: added switch case for converting the `consts` element types to the element types for the `fe` library (because we cannot yet remove the element types from `consts` yet; although this is the goal). and we made these explicit and not related to the enum to make this more robust in the case that new element types are added. [1289c08](https://github.com/SimVascular/svMultiPhysics/pull/561/commits/1289c086f637cdc1544aff0bfe99eb78ad3b9f1c)

aabrown100-git:
@zasexton Thanks for explaining. I think a DESIGN.md in the `Basis` folder (or somewhere similar) explaining that point, other relevant details, and the design of the `Basis` module itself would be helpful.

zasexton:
I think I'll plan to add these notes to the `Basis` topic in the doxygen documentation which will overview the layout and purpose of this submodule. This way the design overview for this (and eventually other topics) will live in the github pages.

Code/Source/solver/FE/Basis/SerendipityBasis.cpp
```diff
@@ -19,6 +18,61 @@ namespace basis {
 namespace {
 using Vec3 = math::Vector<Real, 3>;
 
+void store_gradient(const Gradient& gradient, Real* dst) {
```
ktbolt: @zasexton What is all of this `Real` type about ? Let's get rid of it unless there is a very very very good reason to use it.

Code/Source/solver/FE/Basis/SerendipityBasis.cpp
```diff
@@ -19,6 +18,61 @@ namespace basis {
...
+    dst[1] = gradient[1];
+    dst[2] = gradient[2];
+}
+
+void evaluate_hex8_reference(Real r,
+                             Real s,
+                             Real t,
+                             Real* values,
+                             Real* gradients,
+                             Real* hessians) {
```
ktbolt: @zasexton Raw pointers ?

Code/Source/solver/FE/Basis/SerendipityBasis.cpp
```diff
@@ -19,6 +18,61 @@ namespace basis {
 namespace {
 using Vec3 = math::Vector<Real, 3>;
 
+void store_gradient(const Gradient& gradient, Real* dst) {
+    dst[0] = gradient[0];
+    dst[1] = gradient[1];
+    dst[2] = gradient[2];
+}
+
+void evaluate_hex8_reference(Real r,
```
ktbolt: @zasexton What exactly is this function doing and what are the parameters ?

Isn't the Hessian of a hex8 element 0 ?

Code/Source/solver/FE/Basis/SerendipityBasis.cpp
```diff
@@ -19,6 +18,61 @@ namespace basis {
...
+        const Real c = Real(signs[i][2]);
+        const Real ar = Real(1) + a * r;
+        const Real bs = Real(1) + b * s;
+        const Real ct = Real(1) + c * t;
+
+        if (values) {
+            values[i] = Real(0.125) * ar * bs * ct;
+        }
+        if (gradients) {
+            Real* g = gradients + i * 3u;
```
ktbolt: @zasexton Offsetting from a raw pointer ? Yuck, like C !

aabrown100-git:
@zasexton good idea! Will that be written in the comments of the code, or a separate file?

zasexton:
@aabrown100-git I'm intending for it to live in the `BasisFunction.h` file so that doxygen can build it. I believe this will follow a similar style to how deal.ii creates topic sections for different components of the code. An example from the compiled html pages is attached here along with a crude input/output data flow diagram to conceptually communicate where the Basis function objects responsibilities sit in the chain of calls. Let me know what you think. 
basis_dox_inputs-responsibilities

Code/Source/solver/FE/Basis/SerendipityBasis.cpp
zasexton: Right now `Real` is an alias for the scalar type and defaults to `double`. My thoughts were that as we refactored the FE machinery we would use `Real` which would serve as the solver-wide scalar alias policy. Specifically I implemented this in case there was ever a need to switch to `float` for a type of GPU support in the future. I can go ahead and replace these if this seems to be more confusion than it is worth.

Code/Source/solver/FE/Basis/SerendipityBasis.cpp
zasexton: I know, I know....my thoughts here: these calls allow for genuine zero-allocation evaluation paths for solver production runs and compiler restriction optimizations. The `evaluate_values` along with their gradient and hessian analogues offer the safe, sized, allocated API while the `*_to` methods offer these allocation-free methods.  I could replace the raw pointers with `std::span` and remove the `restrict` to avoid directly using raw pointers while maintaining the flat, non-owning, no-allocation buffer model. We just wouldn't be able to use the compiler `restrict` anymore which is probably fine if we are always caching these values anyway.

Code/Source/solver/FE/Basis/SerendipityBasis.cpp
zasexton: The pure second derivative terms are 0 but the mixed terms are not.

Code/Source/solver/FE/Basis/SerendipityBasis.cpp
zasexton: mainly I'm trying to avoid having to create temporary `Gradient` or `Hessian` objects within functions that are internal. since the output buffer shape is pretty simple for gradients and hessians I opted for the offsets. I can change these if to have an accessor if you think this would be better long-term.

Code/Source/solver/FE/Basis/SerendipityBasis.cpp
ktbolt: @zasexton What I meant was why did the original code set the Hessian to 0 and why is that being changed ?

zasexton: I'm not sure why the original code set the Hessian to zero...in the svFSI the Hessian is zero initialized and never populated. So the svFSI and svMultiPhysics codes are doing the same thing I'm just not sure why

zasexton: as for why it's being changed: I don't think it makes sense that a change in the element type should affect the formulated residual that is being solved. Either we chose to zero out the Hessian completely (effectively eliminating these terms in the formulation) or we consider reference solutions that appropriately include these terms (as seems to be the case in the fsi formulation intent). For linear tet elements this doesn't matter since the Hessian is completely zero, but this is a special case.

aabrown100-git: I think we should include non-zero Hessians unless there is a real numerical reason not to, but I don't think that's the case (although I haven't looked closely into this)

michelebucelli: If I remember correctly, the convergence of residual-based stabilization (SUPG, VMS-LES) rests on the assumption that the residual appearing in the stabilization terms is zero if evaluated on the exact solution (that is, the method must be consistent, or weakly consistent at least). If that residual does not include the laplacian term, then there is an inconsistency, and the method is not convergent. Optimistically, the error (numerical vs. exact solution) will stagnate as the mesh is refined. In light of this, I think that neglecting the Hessian is wrong, and including it is correct.

I do not know how this will affect the convergence properties of the non-linear and linear solvers and the system's conditioning - they might be better or they might be worse. Most likely they will vary between scenarios. But I think this is a secondary issue, and should not affect the decision of whether to include the Hessian or not.

Code/Source/solver/CMakeLists.txt
```diff
@@ -23,15 +23,18 @@
 # LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
 # NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 
-set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD 20)
```
michelebucelli, line 26: What features from C++ 20 are required now?

I don't know about recent updates, but until last year or so compiler support for C++ 20 was still a bit sketchy, and requiring developers to update to the latest compiler versions might be a big ask, especially in HPC environments.

Code/Source/solver/CMakeLists.txt
```diff
@@ -86,7 +89,7 @@ endif()
 # add trilinos flags and defines
 if(USE_TRILINOS)
   ADD_DEFINITIONS(-DWITH_TRILINOS)
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++20")
```
michelebucelli: Similar to the above, but also: would it make sense to take this version number from the CMake variable `CMAKE_CXX_STANDARD`?

Documentation/Doxyfile
```diff
@@ -191,10 +191,10 @@ TREEVIEW_WIDTH         = 250
...
 FORMULA_FONTSIZE       = 10
 USE_MATHJAX            = YES
-MATHJAX_VERSION        = MathJax_3
-MATHJAX_FORMAT         = chtml
-MATHJAX_RELPATH        = https://cdn.jsdelivr.net/npm/mathjax@3
-MATHJAX_EXTENSIONS     = ams
+MATHJAX_VERSION        = MathJax_2
+MATHJAX_FORMAT         = HTML-CSS
+MATHJAX_RELPATH        = https://cdn.jsdelivr.net/npm/mathjax@2
+MATHJAX_EXTENSIONS     = TeX/AMSmath TeX/AMSsymbols
```
michelebucelli: Just out of curiosity, what prompted this downgrading to version 2? Are there new features that are currently unsupported?

Code/ThirdParty/eigen/include/eigen3/unsupported/Eigen/CXX11/Tensor
```diff
@@ -34,7 +34,9 @@
   */
 
 #include <atomic>
+#ifdef EIGEN_USE_GPU
 #include <chrono>
+#endif
```
michelebucelli: Do I understand correctly that these lines are changing code inside the Eigen code? If so, this seems potentially dangerous to me, as we risk branching into a customized version of Eigen that might be hard to update with new Eigen versions if necessary.

What issue prompted introducing this change? Maybe we can figure out a workaround.

michelebucelli (commented):
Thanks @zasexton, I left a few comments mostly related to housekeeping - I haven't yet dug into the more logic-heavy part of the PR.

Documentation/Doxyfile
zasexton: I downgraded this to avoid the polyfilll.io login that pops up on every single doxygen page. I'm sure there is a way around it via some other setup parameter but I figured we could also use version 2 without loss of support for the math rendering that we need.

Code/Source/solver/CMakeLists.txt
zasexton: hmmm if this is the case then I would probably be adding more raw pointer handling back into some of this code. I've been using `std::span` for some of these to offer a safer non-owning view over a contiguous sequence of objects.

Code/ThirdParty/eigen/include/eigen3/unsupported/Eigen/CXX11/Tensor
zasexton: oh yes, I meant to remove this. I had guarded this previously because I was having some clang compiler mismatches that specifically were causing some issues with the standard `chrono` library

zasexton: I've reverted the Eigen chrono guard in [bd7c2ad](https://github.com/SimVascular/svMultiPhysics/pull/561/commits/bd7c2ad86d687606319d768d9abd4ab85c997d63)

Code/Source/solver/CMakeLists.txt
zasexton: For this note, I could make a helper `span` like work around to place in the FE/Common folder.

ktbolt: @zasexton I see that C++ 20 has wide support; now the default for the latest GCC compilers, and the Intel compiler uses Clang/LLVM.

Code/Source/solver/FE/Basis/SerendipityBasis.cpp
zasexton: When looking at the test cases specifically the inclusion of the Hessian for the HEX8 case lightly increases the number of linear iterations. It does not change the number of non-linear iterations. I agree that this may likely vary on a case-by-case basis.

Code/Source/solver/FE/Basis/SerendipityBasis.cpp
ktbolt: @zasexton Be careful not to fall into the premature optimization trap: spending a lot of time writing complex code that uses less memory or runs faster but is more difficult to understand.

The primary goal is to write code that is clean and easy to understand. The solver can later be profiled to identify code sections that actually might could be optimized. Even then you need to think about the tradeoff between more complex code and the overall effect on resource usage and run time.

Code/Source/solver/CMakeLists.txt
zasexton: Then if this is okay with you @ktbolt I will continue to use `std::span` and clean up some of the cmake plumbing as @michelebucelli points out. I believe that it would be great to move to c++20 support if compiler architectures now have wide support.

Code/Source/solver/FE/Basis/SerendipityBasis.cpp
zasexton: @ktbolt gotcha! I'll plan to switch out the raw pointers for the `std::span` here since it will be both cleaner and safer. I'll remove the `restrict` compiler optimizations.

zasexton:
@aabrown100-git github is telling me that you have requested changes I just can't seem to find where. would you mind letting me know where you were originally wanting changes to the code?

Code/Source/solver/CMakeLists.txt
michelebucelli: Ok, thanks for the explanation!

aabrown100-git:
@zasexton Oh I think it's just that my latest review was "requested changes". But I think you addressed all my comments in that review. I can review again and approve if you're ready to merge this.

zasexton:
@aabrown100-git yea! if you have a moment, give it another review :) I think we're getting closer to a merge here. I still have some remaining items to work out related to the raw pointer discussions above, but once these are handled I don't believe there will be any more un-addressed items...

Code/Source/solver/fs.cpp
```diff
@@ -5,10 +5,66 @@
...
+    case ElementType::HEX20:
+    case ElementType::HEX27:
+    case ElementType::WDG:
+      return true;
+    default:
+      return false;
+  }
+}
+
+void populate_reference_hessians_if_supported(fsType& fs, const int insd)
```
aabrown100-git: Going forward, will we ever have elements for which Hessians aren't supported? If not, can we remove this code and just always populate reference Hessians?

Code/Source/solver/nn.cpp
```diff
@@ -37,22 +49,361 @@ using namespace consts;
...
+  }
+  return std::nullopt;
+}
+
+bool use_basis_adapter_for(consts::ElementType eType)
+{
+  return to_fe_element_type(eType).has_value();
+}
+
+bool supports_face_basis_adapter_for(consts::ElementType eType)
```
aabrown100-git, line 116: What is this function for? Can you add a comment description?

Code/Source/solver/nn.cpp
```diff
@@ -37,22 +49,361 @@ using namespace consts;
...
+    // NURBS are outside the current FE Basis scope.
+    case consts::ElementType::NA:
+    case consts::ElementType::PNT:
+    case consts::ElementType::NRB:
+      return std::nullopt;
+  }
+  return std::nullopt;
+}
+
+bool use_basis_adapter_for(consts::ElementType eType)
```
aabrown100-git: Can this function be combined with the previous?

Code/Source/solver/nn.cpp
```diff
@@ -37,22 +49,361 @@ using namespace consts;
...
+
+  const std::lock_guard<std::mutex> lock(cache_mutex);
+  auto it = cache.find(eType);
+  if (it == cache.end()) {
+    it = cache.emplace(eType, febasis::basis_factory::create_default_for(*fe_type)).first;
+  }
+  return *it->second;
+}
+
+std::span<const std::size_t> solver_to_basis_node_map(consts::ElementType eType)
```
aabrown100-git, line 170: I think docstrings for this and following functions would help. Unless there are docstrings in the .h file already?

Code/Source/solver/fs.cpp
zasexton: yes, specifically the `point` and `NURBS` elements right now. However, we probably don't need to have this second checking function since `nn::get_gn_nxx()` already checks for element type compatibility for Hessian values. I can remove this function. This was mostly used for my own unit test checking.

Code/Source/solver/nn.cpp
zasexton: this function is called in `nn.cpp` in the `face get_gnn` to signify topologies that can actually appear as faces — points/lines/surfaces (`LIN1/2`, `TRI3/6`, `QUD4/8/9`) and excludes the volume type elements. I'll add this note.

Code/Source/solver/nn.cpp
zasexton: since this is essentially just one line I can remove this helper and swap out this return line at the call sites.

Code/Source/solver/nn.cpp
zasexton: will add a docstring here.

Code/Source/solver/FE/Basis/SerendipityBasis.cpp
zasexton: I have made these changes and removed the `restrict` compiler hint.

Code/Source/solver/FE/Basis/SerendipityBasis.cpp
zasexton: Removed the raw pointers here in favor of `std::span` for non-owning views of contiguous data so that we can maintain allocation-free methods

Code/Source/solver/FE/Basis/SerendipityBasis.cpp
zasexton: the `Real` alias has been replaced with the `double` type per commit [d442f70](https://github.com/SimVascular/svMultiPhysics/pull/561/commits/d442f70c30abe06c3aeccdd06fb558e7aa362b06)

considering this comment resolved.

Code/Source/solver/FE/Basis/SerendipityBasis.cpp
zasexton: I'm considering this conversation resolved

Code/Source/solver/CMakeLists.txt
zasexton: I've made the appropriate c++ standard variable change for the trilinos flags

ktbolt (commented):
@zasexton Excellent documentation !

Code/Source/solver/FE/Basis/BasisFactory.h
```diff
@@ -32,10 +34,27 @@ struct BasisRequest {
     std::vector<std::vector<double>> axis_weights{};
     std::vector<int> tensor_extents{};
     std::string custom_id{};
+    // Reference topology for arbitrary-order bases. This field is intentionally
+    // last so existing aggregate initializers for named elements keep their
+    // positional meaning.
```
michelebucelli: I suggest turning these comments into Doxygen comments. It would also be nice to add Doxygen comments to the other members as well.

(Maybe the comment on `topology` being intentionally last can be omitted from Doxygen, since it's more of a low-level implementation detail that shouldn't affect the users of this type).

Code/Source/solver/FE/Basis/BasisFactory.h
```diff
@@ -0,0 +1,94 @@
...
+
+#include "BasisFunction.h"
+#include <memory>
+#include <optional>
+#include <string>
+#include <vector>
+
+namespace svmp {
+namespace FE {
+namespace basis {
```
michelebucelli: Minor: I find the following more compact syntax a bit better (less brackets to keep track of):
```suggestion
namespace svmp::FE::basis {
```

Code/Source/solver/Core/Exception.h
```diff
@@ -397,6 +397,38 @@ void check_not_null(PointerT ptr, SourceLocation location, Args&&... args)
...
+[[noreturn]] void not_implemented(SourceLocation location, Args&&... args)
+{
+    raise<ExceptionT>(location, std::forward<Args>(args)...);
+}
+
+template <class ExceptionT, class FeatureT>
+[[noreturn]] void not_implemented(FeatureT&& feature, SourceLocation location)
+{
+    raise<ExceptionT>(location, std::forward<FeatureT>(feature));
+}
```
michelebucelli, line 564: I appreciate very much having shorthand functions for assertions and various types of checks!

However, in some cases with the current exception library, I find that it is not very clear what the advantage or specific use case of the shorthand is, and some of the functions seem to replicate functionality to me. In this particular instance:
1. doesn't `throw_if` have the same behavior and intended use of the already existing function `check`?
2. what's the difference between the two overloads of `not_implemented`? A bit of documentation guiding other developers on which method to use might help here.
3. Since the `not_implemented` function requires the user to specify the exception type, it is not really shorter than calling `svmp::raise<svmp::FE::NotImplementedException>(SVMP_HERE, message)`. I would think it would make more sense if the `not_implemented` function always threw a `NotImplementedException` (which should probably be moved to `Core/Exception.h`, come think of it), with a custom message.
4. With similar logic, I would define an exception type `IndexOutOfRangeException` and have `check_index` always throw that exception.

Points 3 and 4 will also facilitate a more consistent use of exception types, which is something that I've had some doubts about in recent work.

Code/Source/solver/FE/Basis/BasisExceptions.h
```diff
@@ -0,0 +1,111 @@
+// SPDX-FileCopyrightText: Copyright (c) Stanford University, The Regents of the University of California, and others.
+// SPDX-License-Identifier: BSD-3-Clause
+
+#ifndef SVMP_FE_BASIS_BASISEXCEPTIONS_H
+#define SVMP_FE_BASIS_BASISEXCEPTIONS_H
+
+#include "FEException.h"
```
michelebucelli, line 7: This is partially unrelated to this PR, but since you're starting to reorganize the code in folders as part of refactoring, I think it's worth bringing up.

Right now CMake is configured to add all subdirectories to the include search path. I think a cleaner/safer option is to only add the root directory (I guess `solver`), and always include the relative path in include statements.

This way the folders not only provide an organization to the files, but also some sort of "header-level namespacing", i.e. you could have `FE/Basis/Exceptions.h` and `FE/Exceptions.h` - which is not necessarily a good idea, but at least you don't have to come up with a unique name for every header across folders.

(Anecdotally, I have recently had issues with a header not being found because the path wasn't resolved correctly, and using the relative path was an effective solution, but I haven't figure out what was going wrong)

Code/Source/solver/FE/Basis/BasisExceptions.h
```diff
@@ -0,0 +1,111 @@
...
+ * @brief Internal basis construction or transform setup failure
+ */
+class BasisConstructionException : public BasisException {
+public:
+    BasisConstructionException(const std::string& message,
+                               const char* file,
+                               int line,
+                               const char* function)
+        : BasisException(message, file, line, function, StatusCode::InternalError) {}
+};
```
michelebucelli: In the same vein as the previous comment, I am not sure I fully understand the use cases for these classes. While I imagine that the typical developer wouldn't have to raise these exceptions, I still think it would be nice to be a bit more pedagogic for future reference.

I think that adding a more detailed explanatory sentence to the documentation of these classes would help. Additionally, it would be nice to provide one clarifying example for each (e.g. for `BasisElementCompatibilityException`, you could write that it's thrown if requesting a tetrahedral type of element on a hexahedral grid, or something in that vein).

Code/Source/solver/FE/Basis/BasisFactory.h
```diff
@@ -0,0 +1,105 @@
...
+ * BasisRequest::element_type for a named mesh-node layout, or set
+ * BasisRequest::topology for an arbitrary-order reference-topology basis.
+ * Setting neither target, or setting both, is rejected. Named element requests
+ * keep the element's fixed polynomial order contract; topology requests are the
+ * arbitrary-order path.
+ *
+ * @param req Basis family, target, and order request.
+ * @return Shared basis instance.
+ */
+[[nodiscard]] std::shared_ptr<BasisFunction> create(const BasisRequest& req);
```
michelebucelli: Is there any reason to use a `shared_ptr` instead of a `unique_ptr`? The latter is slightly cheaper, its semantics are a bit less complicated, and it can be moved into a `shared_ptr` if needed, that is you would still be able to do
```cpp
std::shared_ptr<BasisFunction> basis = create(req);
```
so that overall the `unique_ptr` is more general.

Same question applies to other uses of `shared_ptr` below.

Code/Source/solver/FE/Basis/BasisFunction.h
```diff
@@ -0,0 +1,339 @@
...
+ *   @ref FE_SerendipityBasis "SerendipityBasis": the implemented nodal
+ *   families, including analytical first and second derivatives in reference
+ *   coordinates.
+ * - basis_factory (BasisFactory.h): runtime construction from a BasisRequest.
+ *   basis_factory::default_basis_request() centralizes the family/order that
+ *   matches each supported element's public node layout.
+ * - ReferenceNodeLayout (NodeOrderingConventions.h): canonical reference-node
+ *   coordinates and the output ordering used by every basis evaluator.
+ * - BasisTraits.h and BasisExceptions.h: topology classification,
+ *   compile-time helpers, and module-specific exception types.
```
michelebucelli: Very minor: I think you can add some further `@ref` commands here to ensure that the other classes are clickable links (Doxygen might be able to do this automatically sometimes, but sometimes it doesn't, not sure why).

Code/Source/solver/FE/Basis/BasisFunction.h
```diff
@@ -0,0 +1,339 @@
...
+ *
+ * ## Object and evaluation contract
+ *
+ * A basis object is immutable after construction. It represents one reference
+ * topology, basis family, and effective polynomial order, and can be shared
```
michelebucelli: I assume that "topology" here refers to the geometry of the element (e.g. triangle/tetrahedron). If that is the case, it might be worth explaining it here. The same goes for basis family. I suggest:

```suggestion
 * topology (e.g. tetrahedron, hexahedron, ...), basis family (Lagrange or serendipity), and effective polynomial order, and can be shared
```

Code/Source/solver/FE/Basis/BasisFunction.h
```diff
@@ -0,0 +1,339 @@
...
+ * safely across evaluations. Construction may build node lattices or invert
+ * interpolation matrices, so callers should construct through basis_factory
+ * and cache one instance for each distinct basis request instead of rebuilding
+ * inside element loops.
```
michelebucelli: If I get it right, the point you're making here is that constructing a `BasisFunction` object is potentially costly. Is that right? If so, I suggest rephrasing for clarity as follows:

```suggestion
 * safely across evaluations. Construction may be computationally expensive,
 * due to building node lattices or inverting nterpolation matrices. Therefore,
 * the basis function object should not be rebuilt inside element loops, but
 * constructed only once for each distinct basis needed, using functions from
 * @ref basis_factory.
```

Code/Source/solver/FE/Basis/BasisFunction.h
```diff
@@ -0,0 +1,339 @@
...
+/**
+ * @brief Throw BasisEvaluationException when an output span is smaller than the
+ * basis size. \p label is the full "Class::method" context used in the message,
+ * so each basis family passes its own qualified name.
+ */
+void require_span_size(std::size_t actual, std::size_t expected, const char* label);
+
+/**
+ * @brief Check a requested output span unless it is empty, following the
+ * "skip this output" convention used by the combined evaluators.
```
michelebucelli: I'm not sure I understand the meaning of "following the skip this output convention used by the combined evaluators". Could you try to clarify?

Code/Source/solver/FE/Basis/BasisFunction.h
```diff
@@ -0,0 +1,339 @@
...
+    }
+}
+
+/**
+ * @brief Abstract interface for finite-element basis-function families.
+ * @ingroup FE_Basis
+ *
+ * BasisFunction defines the common query and evaluation API used by solver
+ * code that does not need to know the concrete basis implementation. Derived
+ * classes provide values at minimum and can override analytical gradients,
```
michelebucelli: Minor suggestion for clarity:
```suggestion
 * classes provide shape function values at minimum and can override analytical gradients,
```

Code/Source/solver/FE/Basis/BasisFunction.h
```diff
@@ -0,0 +1,339 @@
...
+    virtual BasisType basis_type() const noexcept = 0;
+
+    /**
+     * @brief Return the reference topology of this basis.
+     *
+     * @details Together with order(), this is the authoritative identity of a
+     * basis: a topology plus a polynomial order, with no node-count assumption.
+     * Arbitrary-order bases are constructed from a BasisTopology and an order;
+     * named ElementType layouts (Hex8, Hex27, ...) are a fixed-order shorthand
+     * that maps to the same (topology, order) pair.
```
michelebucelli: I am confused by this paragraph: wouldn't the basis family (Lagrange or serendipity or other) also contribute to defining the identity of the basis? Or am I misinterpreting some of the terminology here?

Code/Source/solver/FE/Basis/BasisFunction.h
```diff
@@ -0,0 +1,339 @@
...
+
+    /**
+     * @brief Return the reference-space dimension of the basis.
+     * @return Reference dimension, from zero for points through three for volume elements.
+     */
+    virtual int dimension() const noexcept = 0;
+
+    /**
+     * @brief Return the polynomial order represented by this basis.
+     * @return Effective polynomial order after any element-family normalization.
```
michelebucelli: I'm not sure I understand the meaning of "element-family normalization".

Code/Source/solver/FE/Basis/BasisFunction.cpp
```diff
@@ -0,0 +1,136 @@
...
+// only the vector evaluators: they evaluate into a temporary and copy into the
+// caller's span. The concrete nodal families (LagrangeBasis, SerendipityBasis)
+// override these to compute directly into the span without the temporary.
+void BasisFunction::evaluate_values_to(const math::Vector<double, 3>& xi,
+                                       std::span<double> values_out) const {
+    require_span_size(values_out.size(), size(), "BasisFunction::evaluate_values_to");
+    std::vector<double> tmp(size());
+    evaluate_values(xi, tmp);
+    std::copy_n(tmp.begin(), tmp.size(), values_out.begin());
+}
```
michelebucelli: Would it be possible to have the derived classes only override the method that fills an `std::span`, and then the base class implements the methods evaluating on a vector by allocating that vector and falling back onto the `std::span` method? Something like:
```cpp
// Derived classes need to implement this, and that's where the basis
// evaluation is defined.
virtual void
evaluate_values_to(const math::Vector<double, 3>& xi,
                   std::span<double> values_out) const = 0;

void
evaluate_values(const math::Vector<double, 3>& xi,
                std::vector<double>& values) const {
values.resize(size());
evaluate_values_to(values);
}
```

I think this should allow to have the maximum efficiency with the minimum implementation.

Code/Source/solver/FE/Basis/BasisFunction.h
```diff
@@ -0,0 +1,339 @@
...
+                                    std::vector<Gradient>& gradients) const;
+
+    /**
+     * @brief Evaluate basis Hessians at a reference coordinate.
+     * @param xi Reference coordinate. Lower-dimensional elements use the active prefix components.
+     * @param hessians Receives one 3-by-3 Hessian per basis function.
+     * @throws BasisEvaluationException If Hessians are not available for the basis.
+     */
+    virtual void evaluate_hessians(const math::Vector<double, 3>& xi,
+                                   std::vector<Hessian>& hessians) const;
```
michelebucelli: Is there any reason preventing these methods from returning their result (as opposed to passing it out as an output argument)? I think return-value optimization should allow to do that for free here (see also [here](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#rf-out)).

If not, the functions should document whether they resize the output argument or whether it is assumed to be of the correct size already.

Code/Source/solver/FE/Basis/BasisTraits.h
```diff
@@ -0,0 +1,249 @@
...
+            return 1;
+        case ElementType::Line3:
+        case ElementType::Triangle6:
+        case ElementType::Quad9:
+        case ElementType::Tetra10:
+        case ElementType::Hex27:
+        case ElementType::Wedge18:
+            return 2;
+        default:
+            return -1;
```
michelebucelli, line 138: Is this -1 value used in any way, that are there any "legal" paths that would reach this branch, or could we raise an exception here instead (e.g. unknown element type or something similar)?

The same comment applies below, and (although to a lesser extent) to functions returning `ElementType::Unknown`.

Code/Source/solver/FE/Math/Vector.h
```diff
@@ -0,0 +1,53 @@
...
+namespace math {
+
+/**
+ * @brief Fixed-size column vector for element-level computations
+ * @ingroup FE_VectorMath
+ * @tparam T Scalar type (float, double)
+ * @tparam N Vector dimension
+ */
+template<typename T, std::size_t N>
+using Vector = Eigen::Matrix<T, static_cast<int>(N), 1>;
```
michelebucelli, line 51: Is this supposed to replace the other `Vector` class from `solver/Vector.h` (now or in the future)?

Code/Source/solver/FE/Math/DenseLinearAlgebra.h
```diff
@@ -0,0 +1,126 @@
...
+
+[[nodiscard]] std::size_t dense_matrix_rank(std::vector<double> matrix,
+                                            std::size_t rows,
+                                            std::size_t cols);
+
+[[nodiscard]] DensePseudoInverseResult rank_revealing_pseudo_inverse(
+    std::span<const double> matrix,
+    std::size_t rows,
+    std::size_t cols,
+    std::string_view label = "dense matrix");
```
michelebucelli: I suggest adding some Doxygen documentation to these functions - some of them are obvious, but not all of them, and the meaning of their argument is not always obvious (for example, the purpose of `label` is not immediately clear from the signature).

Also, since the `label` argument is used to construct error message, I suggest renaming it to `error_message_label` to make this more self-documenting (in addition to the documentation comments suggested above).

Code/Source/solver/FE/Basis/NodeOrderingConventions.h
```diff
@@ -0,0 +1,142 @@
...
+
+namespace svmp {
+namespace FE {
+namespace basis {
+
+// NodeOrderingConventions is internal to the Basis module: it is the reference-node
+// generator the basis families build on, not a consumer entry point (callers obtain
+// a basis through basis_factory and read BasisFunction::nodes()), so it is excluded
+// from the public Doxygen output below.
+/** @cond INTERNAL */
```
michelebucelli: I'm not sure I would exclude this from Doxygen (especially since you wrote a pretty good and detailed documentation already).

While it should never be accessed by model developers (e.g. people who implement or maintain physical models but don't want to touch the finite element core) it might have to be accessed by core developers in the future, and they might enjoy having Doxygen-formatted documentation for this.

If you want to dissuade people from trying to use this directly, you can add a `@warning` comment somewhere at the top, along the lines of "You should never use this directly, but only through `basis_factory` and `BasisFunction`. These functions are part of internal implementation and their interface is subject to change". This should be rendered into a big yellow box that is hard (although, admittedly, not impossible) to miss.

Code/Source/solver/FE/Basis/LagrangeBasis.h
```diff
@@ -0,0 +1,358 @@
...
+ *   the distribution and its conditioning -- not an equispaced layout.
+ * - **Simplex (triangle, tetrahedron):** the equispaced barycentric lattice
+ *   (each barycentric coordinate at @f$i/p@f$). The closed-form evaluator below
+ *   is specific to this equispaced lattice.
+ * - **Wedge:** the tensor product of an equispaced triangle cross-section with a
+ *   GLL through-axis.
+ *
+ * Because GLL coincides with the equispaced layout at orders 1 and 2
+ * (line_coord_pm_one), every named production element keeps its standard
+ * coordinates and the GLL/equispaced distinction appears only for order >= 3.
```
michelebucelli: If I get it right, "every named production element keeps its standard coordinates" is a reference to the current implementation of basis functions. Is that right?

If so, this comment might not age well, as the number of people that know of the previous implementation reduces (due to turnover or forgetfulness 😄). I suggest the following rephrasing, which should be more independent of the current state of things:
> GLL nodes are equispaced for orders 1 and 2. Therefore, elements Hex8, Hex27 [...add any others here...] are built on equispaced nodes. The GLL/equispaced distinction appears only for order >= 3.

Code/Source/solver/FE/Basis/LagrangeBasis.cpp
```diff
@@ -0,0 +1,643 @@
...
+
+    svmp::raise<BasisEvaluationException>(SVMP_HERE,
+        "Unsupported element in LagrangeBasis evaluation");
+}
+
+void LagrangeBasis::evaluate_values(const Vec3& xi,
+                                    std::vector<double>& values) const {
+    values.resize(size());
+    evaluate_values_to(xi, std::span<double>(values.data(), values.size()));
+}
```
michelebucelli: Related to my previous comment on `BasisFunction::evaluate_values`, I think this could be the default implementation in `BasisFunction` (perhaps even marked as `final`).

The same suggestion applies to the other `evaluate_` methods.

Code/Source/solver/FE/Common/Types.h
```diff
@@ -0,0 +1,551 @@
...
+
+#if defined(SVMP_FE_WITH_MESH) && SVMP_FE_WITH_MESH
+#  include "Mesh/Core/MeshTypes.h"
+/** Nonzero when FE shares scalar/index types with the Mesh library. */
+#  define SVMP_FE_HAS_MESH_TYPES 1
+#else
+// Build FE without Mesh types unless explicitly enabled.
+/** Nonzero when FE shares scalar/index types with the Mesh library. */
+#  define SVMP_FE_HAS_MESH_TYPES 0
+#endif
```
michelebucelli, line 30: Is this `Mesh` library an external (optional) module of `svmp`? Or is it part of the planned refactor?

Code/Source/solver/FE/Common/Types.h
```diff
@@ -0,0 +1,551 @@
...
+    Hex20      = 16,  ///< 20-node hexahedron (serendipity)
+    Wedge15    = 17,  ///< 15-node wedge
+    Wedge18    = 18,  ///< 18-node wedge (complete quadratic)
+    Pyramid13  = 19,  ///< 13-node pyramid
+    Pyramid14  = 20,  ///< 14-node pyramid
+
+    // Special elements
+    Point1     = 30,  ///< 1-node point element
+
+    Unknown    = 255  ///< Unrecognized or uninitialized element type
```
michelebucelli: Is there any meaning attached to the numbers used here? If not, I would suggest removing them (after all, the point of `enum class` is that the underlying numbers are unimportant as long as they are different).

Code/Source/solver/FE/Common/Types.h
```diff
@@ -0,0 +1,551 @@
...
+     * @brief Convert to the underlying global index value.
+     * @return The stored global index.
+     */
+    constexpr operator GlobalIndex() const noexcept { return value; }
+    /**
+     * @brief Check whether this index refers to a valid DOF.
+     * @return True when the stored value is non-negative.
+     */
+    constexpr bool is_valid() const noexcept { return value >= 0; }
+};
```
michelebucelli: Would it make sense for this to be replaced by `StrongType<GlobalIndex, ...>`?

Code/Source/solver/nn.cpp
```diff
@@ -37,22 +49,401 @@ using namespace consts;
...
+  }
+  return "unknown (" + std::to_string(static_cast<int>(eType)) + ")";
+}
+
+/// Translate a solver element type into its FE library counterpart. This is a
+/// pure renaming between the two enum vocabularies: the FE library owns the
+/// choice of basis family and polynomial order for each element type
+/// (basis_factory::default_basis_request). The switch deliberately has no
+/// default case so that compilers building with -Wswitch flag any newly added
+/// solver element type that is missing a mapping here. Returns std::nullopt for
```
michelebucelli: I would have a mild preference towards a `default` case throwing an exception, rather than relying on the warning being turned on (although I recognize this would be caught at runtime, so the behavior wouldn't be exactly the same).

michelebucelli (commented):
Thank you @zasexton, this is lots of work! I added some other comments.

Also good job with the documentation, it's very detailed.

Code/Source/solver/Core/Exception.h
zasexton: addressed and resolved these in [7003bf3](https://github.com/SimVascular/svMultiPhysics/pull/561/commits/7003bf3d66207135348731f4c3d932a5d4558aa4) to make future exception infrastructure easier to develop and clear. removing the duplicated `check_arg`

Code/Source/solver/FE/Basis/BasisExceptions.h
zasexton: I will defer this option to a smaller PR for now. I agree we have the option which is nice.

Code/Source/solver/FE/Basis/BasisTraits.h
zasexton: The `-1` currently is used by callers `BasisFactory::default_basis_request` and `named_lagrange_order()` and raise exceptions with context for serendipity and lagrange basis functions. I agree that certain aliases may not even exist but I wire them through to the basis family selection anyway since certain aliases conflict with basis family selection e.g. `Hex20` for lagrange despite being an entirely valid element serendipity basis generally. Specifically this is a guard on erroneous Lagrange-on-Serendipity basis requests

Code/Source/solver/FE/Common/Types.h
zasexton: In my opinion, it will be part of the planned refactor as another library that will be used by the `FE` library. The trait interface will allow for topology and cell identification and mapping to the corresponding `FE` elements. Otherwise how mesh data is processed and augmented I imagine we can detangle from much of the `FE` code. These traits and cmake definitions are only the precursors. I can also remove them here to make the PR more narrow

Code/Source/solver/FE/Common/Types.h
zasexton: I agree. I'll make this change.

Code/Source/solver/FE/Common/Types.h
zasexton: perhaps only if we serialized these types for persistent files in the future but currently this is not necessary so you are correct. I'll remove these for clarity.

Code/Source/solver/FE/Math/Vector.h
zasexton: No, this will just be used for the dense vector and matrix operations right now (similar to how the material models implement `array`). eventually these should all be formalized to an Eigen system but this will have to be piecemeal.

Code/Source/solver/nn.cpp
zasexton: have made changes for this so that we throw an error and have changed the documentation appropriately.

Code/Source/solver/FE/Basis/BasisFactory.h
zasexton: have resolved these to unique pointers

Code/Source/solver/FE/Basis/BasisFunction.cpp
zasexton: changed `_to` methods to be pure virtual with the `evaluate_values` as a non-virtual wrapper

Code/Source/solver/FE/Basis/BasisFunction.h
zasexton: added documentation for resizing behavior.

Code/Source/solver/FE/Math/DenseLinearAlgebra.h
zasexton: added doxygen documentation for these methods

Code/Source/solver/FE/Basis/LagrangeBasis.cpp
zasexton: we have now made the base wrappers all non-virtual now in addressing the previous comment. I think this should be pretty reasonable to addressing the series of comments (but obviously will not be override and final are no longer applicable with this change)

Code/Source/solver/FE/Basis/LagrangeBasis.h
zasexton: the documentation is now changed to be more specific to the equispaced nodes so that newer users/developers don't have to infer the previous implementation

Code/Source/solver/FE/Basis/NodeOrderingConventions.h
zasexton: have added this documentation and clearly specified in the doxygen that `NodeOrderingConventions.h` is internal and not supposed to be consumed by other functions outside of `Basis` while there might be utility for future developers/users understanding how these node layout functions work I want to make it clear they should not be used elsewhere.

Code/Source/solver/FE/Basis/BasisFunction.h
zasexton: added specific information in the doxygen to explain that this returns the polynomial order of the basis function implied by the named type (`Quad8` and `Hex20` order 2 for example)

Code/Source/solver/FE/Basis/BasisFunction.h
zasexton: updated the docs to specifically mention that element name also implies basis family type

Code/Source/solver/FE/Basis/BasisFunction.h
zasexton: added

Code/Source/solver/FE/Basis/BasisFunction.h
zasexton: rephrased for clarity

Code/Source/solver/FE/Basis/BasisFunction.h
zasexton: changed the paragraph

Code/Source/solver/FE/Basis/BasisFunction.h
zasexton: applied doc changes to clarify

Code/Source/solver/FE/Basis/BasisFunction.h
zasexton: applied doc changes to add ref

Code/Source/solver/FE/Basis/BasisExceptions.h
zasexton: added further comments to clarify and explain exceptions

Code/Source/solver/FE/Basis/BasisFactory.h
zasexton: have made the appropriate stylistic change to the compact namespace convention. This will also set a precedent for the rest of the solver refactor.

Code/Source/solver/FE/Basis/BasisFactory.h
zasexton: made these into doxygen comments

zasexton:
I believe we should be ready to merge this PR into the repo. I believe the basis refactor is complete and the documentation is sufficient for future users and developers. the code has good coverage via unit testing and passes all current integration tests.

ktbolt (approved):
@zasexton Let's merge it.

zasexton:
@aabrown100-git @michelebucelli do either of you have remaining comments? we'll need one more review approval for this merge

michelebucelli (approved):
Looks good to me!

aabrown100-git (approved):
Approve!

PR #565, opened 2026-06-08
Include valve velocity in the URIS valve implementation

### LINKED ISSUE #558: Including valve velocity in the URIS valve implementation
### Problem

The current URIS valve implementation is a simplified version that does not include valve velocity (assumed to be zero). The background fluid velocity streamlines and flow patterns during the valve opening and closing phases are affected.

### Solution

Compute the valve velocity from the prescribed URIS valve motion and interpolate the resulting velocity field onto the background fluid mesh to incorporate valve motion into the URIS implementation.

### Additional context

_No response_

### Code of Conduct

hanzhao2020:
Resolves https://github.com/SimVascular/svMultiPhysics/issues/558

## Release Notes 
* Computed valve velocity $u_{\Gamma}$ from prescribed valve motion using finite differences
* Included valve velocity $u_{\Gamma}$ in the URIS force term $f_{RIS} = R {\delta} (u - u_{\Gamma})$
* Added an optional input argument `Include_URIS_velocity` to enable inclusion of valve velocity. The default is `false`

## Testing
* Updated the URIS CFD test case reference results with valve velocity enabled. All tests passed

### DIFF
--- a/Code/Source/solver/ComMod.h
+++ b/Code/Source/solver/ComMod.h
@@ -1469,6 +1469,12 @@ class urisType
     // Valve surface position coordinates.
     Array<double> x;
 
+    // Valve position coordinates at the previous time step.
+    Array<double> x_prev;
+
+    // Valve velocity on the valve surface nodes.
+    Array<double> valve_velocity;
+
     // Valve displacement.
     Array<double> Yd;
 
@@ -1493,6 +1499,9 @@ class urisType
     // this assumption, this flag should be set to true to flip the normals.
     bool invert_normal;
 
+    // Whether to include the valve velocity in the RIS implementation. Default is false.
+    bool include_RIS_velocity;
+
     // Opening positions of the valve surfaces.
     Array3<double> DxOpen;
 
@@ -1508,8 +1517,14 @@ class urisType
     // Iteration count.
     int cnt = 1000000;
 
-    // Signed distance function indexed by fluid/background mesh node.
+    // Flag to indicate if the SDF is computed.
+    bool sdf_computed = false;
+
+    // Signed distance function indexed by backgroundfluid mesh node.
     Vector<double> sdf;
+
+    // Valve velocity interpolated on background fluid mesh nodes.
+    Array<double> valve_velocity_fluid;
 
     // Mesh scale factor.
     double scF;
--- a/Code/Source/solver/ComMod.h
+++ b/Code/Source/solver/ComMod.h
@@ -1500,7 +1500,7 @@ class urisType
     bool invert_normal;
 
     // Whether to include the valve velocity in the RIS implementation. Default is false.
-    bool include_RIS_velocity;
+    bool include_uris_velocity;
 
     // Opening positions of the valve surfaces.
     Array3<double> DxOpen;
--- a/Code/Source/solver/ComMod.h
+++ b/Code/Source/solver/ComMod.h
@@ -1500,7 +1500,7 @@ class urisType
     bool invert_normal;
 
     // Whether to include the valve velocity in the RIS implementation. Default is false.
-    bool include_uris_velocity;
+    bool include_uris_velocity = false;
 
     // Opening positions of the valve surfaces.
     Array3<double> DxOpen;
--- a/Code/Source/solver/fluid.cpp
+++ b/Code/Source/solver/fluid.cpp
@@ -621,9 +621,10 @@ void construct_fluid(ComMod& com_mod, const mshType& lM, const SolutionStates& s
     Array<double> ksix(nsd,nsd);
     // Total resistance factor value of the RIS valves for the current element 
     // at different quadrature points
-    Vector<double> risFactorTotalEl;
+    Vector<double> urisFactorTotalEl;
+    Array<double> urisValveVelTotalEl;
     if (com_mod.urisFlag) {
-      uris::uris_compute_ris_factor(com_mod, lM, fs[0], e, risFactorTotalEl);
+      uris::eval_uris_ris_factors_quadrature(com_mod, lM, fs[0], e, urisFactorTotalEl, urisValveVelTotalEl);
     }
 
     for (int g = 0; g < fs[0].nG; g++) {
@@ -661,12 +662,15 @@ void construct_fluid(ComMod& com_mod, const mshType& lM, const SolutionStates& s
         auto N0 = fs[0].N.rcol(g); 
         auto N1 = fs[1].N.rcol(g); 
         double risFactorTotal = 0.0;
+        Vector<double> risValveVelTotal(nsd);
+        risValveVelTotal = 0.0;
         if (com_mod.urisFlag) {
-          risFactorTotal = risFactorTotalEl(g);
+          risFactorTotal = urisFactorTotalEl(g);
+          risValveVelTotal = urisValveVelTotalEl.rcol(g);
         }
         fluid_3d_m(com_mod, vmsStab, fs[0].eNoN, fs[1].eNoN, w, ksix, N0, N1, 
             Nwx, Nqx, Nwxx, al, yl, bfl, lR, lK, K_inverse_darcy_permeability, 
-            risFactorTotal);
+            risFactorTotal, risValveVelTotal);
 
       } else if (nsd == 2) {
         auto N0 = fs[0].N.rcol(g); 
@@ -694,8 +698,8 @@ void construct_fluid(ComMod& com_mod, const mshType& lM, const SolutionStates& s
     // If the number of quadrature points is different for the continuity and 
     // momentum function spaces, recompute the RIS factor
     if (com_mod.urisFlag) {
-      if (static_cast<int>(risFactorTotalEl.size()) != fs[1].nG) {
-        uris::uris_compute_ris_factor(com_mod, lM, fs[1], e, risFactorTotalEl);
+      if (static_cast<int>(urisFactorTotalEl.size()) != fs[1].nG) {
+        uris::eval_uris_ris_factors_quadrature(com_mod, lM, fs[1], e, urisFactorTotalEl, urisValveVelTotalEl);
       }
     }
 
@@ -725,12 +729,15 @@ void construct_fluid(ComMod& com_mod, const mshType& lM, const SolutionStates& s
         auto N0 = fs[0].N.rcol(g); 
         auto N1 = fs[1].N.rcol(g); 
         double risFactorTotal = 0.0;
+        Vector<double> risValveVelTotal(nsd);
+        risValveVelTotal = 0.0;
         if (com_mod.urisFlag) {
-          risFactorTotal = risFactorTotalEl(g);
+          risFactorTotal = urisFactorTotalEl(g);
+          risValveVelTotal = urisValveVelTotalEl.rcol(g);
         }
         fluid_3d_c(com_mod, vmsStab, fs[0].eNoN, fs[1].eNoN, w, ksix, N0, N1, 
               Nwx, Nqx, Nwxx, al, yl, bfl, lR, lK, K_inverse_darcy_permeability, 
-              risFactorTotal);
+              risFactorTotal, risValveVelTotal);
 
       } else if (nsd == 2) {
         auto N0 = fs[0].N.rcol(g); 
@@ -1439,7 +1446,7 @@ void fluid_3d_c(ComMod& com_mod, const int vmsFlag, const int eNoNw, const int e
     const Array<double>& Kxi, const Vector<double>& Nw, const Vector<double>& Nq, const Array<double>& Nwx, 
     const Array<double>& Nqx, const Array<double>& Nwxx, const Array<double>& al, const Array<double>& yl, 
     const Array<double>& bfl, Array<double>& lR, Array3<double>& lK, double K_inverse_darcy_permeability, 
-    double risFactorTotal)
+    const double risFactorTotal, const Vector<double>& risValveVelTotal)
 {
   #define n_debug_fluid3d_c
   #ifdef debug_fluid3d_c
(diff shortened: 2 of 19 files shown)


Code/Source/solver/ComMod.h
```diff
@@ -1493,6 +1499,9 @@ class urisType
     // this assumption, this flag should be set to true to flip the normals.
     bool invert_normal;
 
+    // Whether to include the valve velocity in the RIS implementation. Default is false.
+    bool include_RIS_velocity;
```
ktbolt: @hanzhao2020 Set a value for this.

Code/Source/solver/fluid.cpp
```diff
@@ -661,12 +662,15 @@ void construct_fluid(ComMod& com_mod, const mshType& lM, const SolutionStates& s
         auto N0 = fs[0].N.rcol(g); 
         auto N1 = fs[1].N.rcol(g); 
         double risFactorTotal = 0.0;
+        Vector<double> risValveVelTotal(nsd);
+        risValveVelTotal = 0.0;
```
ktbolt: @hanzhao2020 Arrays values are set to 0.0 when they are allocated.

Code/Source/solver/fluid.cpp
```diff
@@ -694,8 +698,8 @@ void construct_fluid(ComMod& com_mod, const mshType& lM, const SolutionStates& s
     // If the number of quadrature points is different for the continuity and 
     // momentum function spaces, recompute the RIS factor
     if (com_mod.urisFlag) {
-      if (static_cast<int>(risFactorTotalEl.size()) != fs[1].nG) {
-        uris::uris_compute_ris_factor(com_mod, lM, fs[1], e, risFactorTotalEl);
+      if (static_cast<int>(urisFactorTotalEl.size()) != fs[1].nG) {
```
ktbolt: @hanzhao2020

ktbolt: @hanzhao2020 `urisFactorTotalEl.size()` is an `int`.

Code/Source/solver/fluid.cpp
```diff
@@ -725,12 +729,15 @@ void construct_fluid(ComMod& com_mod, const mshType& lM, const SolutionStates& s
         auto N0 = fs[0].N.rcol(g); 
         auto N1 = fs[1].N.rcol(g); 
         double risFactorTotal = 0.0;
+        Vector<double> risValveVelTotal(nsd);
+        risValveVelTotal = 0.0;
```
ktbolt: @hanzhao2020 `Vector` is initialized to 0.0.

Code/Source/solver/fluid.cpp
hanzhao2020: @ktbolt Thanks, setting arrays to 0 after allocation has been removed.

Code/Source/solver/ComMod.h
hanzhao2020: @ktbolt Set the default value to `false`.

Code/Source/solver/uris.cpp
```diff
@@ -648,11 +649,11 @@ void uris_read_msh(Simulation* simulation) {
...
 
-      Array3<double> dispOpen(dispNtOpen, nsd, dispNnOpen);
-      for (int t = 0; t < dispNtOpen; t++) {
-        for (int a = 0; a < dispNnOpen; a++) {
+      Array3<double> dispOpen(nsd, dispNnOpen, dispNtOpen);
+      for (int k = 0; k < dispNtOpen; k++) {
+        for (int j = 0; j < dispNnOpen; j++) {
           for (int i = 0; i < nsd; i++) {
-            file_stream >> dispOpen(t,i,a);
+            file_stream >> dispOpen(i,j,k);
```
ktbolt, line 667: @hanzhao2020 Add error checking.

Code/Source/solver/uris.cpp
```diff
@@ -704,40 +702,70 @@ void uris_read_msh(Simulation* simulation) {
...
+          for (int j = 0; j < uris_obj.tnNo; j++) {
+            for (int k = 0; k < nsd; k++) {
+              uris_obj.DxOpen(k,j,i) = tmpDxOpen(k,j,i);
+            }
+          }
         }
-        // Move data for open
-        Array3<double> tmpDxClose(dispNtClose, nsd, uris_obj.tnNo);
+        // Move data for close
+        Array3<double> tmpDxClose(nsd, uris_obj.tnNo, dispNtClose);
```
ktbolt: @hanzhao2020 You can just do a copy 
```
auto tmpDxClose = uris_obj.DxClose;
```

Code/Source/solver/fluid.cpp
```diff
@@ -621,9 +621,10 @@ void construct_fluid(ComMod& com_mod, const mshType& lM, const SolutionStates& s
     Array<double> ksix(nsd,nsd);
     // Total resistance factor value of the RIS valves for the current element 
     // at different quadrature points
-    Vector<double> risFactorTotalEl;
+    Vector<double> urisFactorTotalEl;
+    Array<double> urisValveVelTotalEl;
```
aabrown100-git: Perhaps rename this variable? At first, I thought this was just the valve velocity, but it seems like it's the forcing term associated with valve velocity?

aabrown100-git: Maybe urisValveVelFactorTotalEl?

aabrown100-git: Actually now I'm a bit confused about `urisValveVelTotalEl`. What does it correspond to?

Code/Source/solver/fluid.cpp
```diff
@@ -1682,11 +1687,11 @@ void fluid_3d_c(ComMod& com_mod, const int vmsFlag, const int eNoNw, const int e
     // up[2] = -tauM*(rho*rV[2] + px[2] - rS[2] + mu*K_inverse_darcy_permeability*u[2]);
 
     up[0] = -tauM*(rho*rV[0] + px[0] - rS[0] + mu*K_inverse_darcy_permeability*u[0]
-                   + risFactorTotal*u[0]);
+                   + urisFactorTotal*u[0] - urisValveVelTotal[0]);
```
aabrown100-git: It would be nice if these two terms had the same structure. For example, can you replace `urisValveVelTotal[0]` with something like `urisFactorTotal * u_valve[0]`?

tests/cases/uris/pipe_uris_cfd/result_005.vtu
aabrown100-git, line 1: Can you verify that with URIS_velocity off, the code reproduces this result exactly?

tests/cases/uris/pipe_uris_cfd/solver.xml
```diff
@@ -5,7 +5,7 @@
 
   <Continue_previous_simulation> false </Continue_previous_simulation>
   <Number_of_spatial_dimensions> 3 </Number_of_spatial_dimensions> 
-  <Number_of_time_steps> 5 </Number_of_time_steps> 
+  <Number_of_time_steps> 3 </Number_of_time_steps>
```
aabrown100-git, line 8: Does the valve start opening/closing by the 3rd timestep?

tests/cases/uris/pipe_uris_cfd/result_005.vtu
aabrown100-git, line 1: I'd rather keep the old URIS test (without valve velocity) and add a new test with valve velocity.

@ktbolt What do you think? I know we're starting to worry about too many integration tests.

Code/Source/solver/uris.cpp
```diff
@@ -955,36 +983,38 @@ void uris_calc_sdf(ComMod& com_mod) {
...
+        uris_obj.x_prev = uris_obj.x;
       }
-    } else {
-      cnt = std::min(uris_obj.cnt, uris_obj.DxClose.nrows());
-      for (int i = 0; i < uris_obj.x.nrows(); i++) {
-        for (int j = 0; j < uris_obj.x.ncols(); j++) {
-          uris_obj.x(i,j) = uris_obj.DxClose(cnt-1,i,j);
-        }
+      for (int i = 0; i < uris_obj.tnNo; i++) {
+        uris_obj.valve_velocity.rcol(i) = (uris_obj.x.rcol(i) - uris_obj.x_prev.rcol(i)) / com_mod.dt;
```
aabrown100-git, line 1035: I was wondering, could we add some check that the user has provided enough frames of opening/closing motion to yield reasonable velocities? 

Also, do you have a script to interpolate opening/closing motion in time that you can add to the `utilities` directory? That might make this uris velocity feature more useful. I'm thinking the user could provide the opening/closing data they have, as well as the timestep size and desired opening and closing durations, and this script could create the necessary interpolated opening/closing data.

aabrown100-git (changes requested):

Code/Source/solver/fluid.cpp
hanzhao2020: @ktbolt Removed `static_cast`

Code/Source/solver/uris.cpp
hanzhao2020: @ktbolt An error check is added after reading from `file_stream`, does this look good?

Code/Source/solver/fluid.cpp
hanzhao2020: @ktbolt Fixed in the updated code.

Code/Source/solver/fluid.cpp
hanzhao2020: @aabrown100-git This variable corresponds to the sum of multiple valve velocity contributions $R_1\delta_1 u_{\Gamma_1} + R_2\delta_2 u_{\Gamma_2} + R_3\delta_3 u_{\Gamma_3} + \ldots$ from the resistive force term $\sum_{i=1}^n R_i \delta_i(u_i - u_{\Gamma_i})$. `urisValveVelFactorTotalEl` sounds good!

Code/Source/solver/fluid.cpp
hanzhao2020: @aabrown100-git Yes, I was thinking of using this. But the factor for each valve may be different in this term $R_1\delta_1 u_{\Gamma_1} + R_2\delta_2 u_{\Gamma_2} + R_3\delta_3 u_{\Gamma_3} + \ldots$, the factor can't be extracted for all valves. If we want to use the same structure, an additional loop over all valves is needed in the fluid code.

tests/cases/uris/pipe_uris_cfd/result_005.vtu
hanzhao2020: @aabrown100-git Yes, verified!

tests/cases/uris/pipe_uris_cfd/solver.xml
hanzhao2020: @aabrown100-git The valve starts opening at the 2nd time step in this test case.

tests/cases/uris/pipe_uris_cfd/result_005.vtu
hanzhao2020: @aabrown100-git The test for the URIS valve in the [FSI case](https://github.com/SimVascular/svMultiPhysics/tree/main/tests/cases/uris/pipe_uris_fsi) still uses the valve without velocity. That test still passed without any changes.

Code/Source/solver/uris.cpp
hanzhao2020: @ktbolt Fixed

Code/Source/solver/uris.cpp
hanzhao2020: @aabrown100-git I think implementing a robust check for whether the valve velocity is reasonable would require quite a bit of additional work. It needs to consider whether the velocity magnitude and direction are physically reasonable, and detect sudden velocity changes between consecutive time steps.

I have a script that interpolates the opening/closing motion for a given input dataset, based on selected motions and desired number of steps. It can generate fairly reasonable velocity fields with some manual inspection.

Since this seems like a useful enhancement, would it make sense to open a separate issue for the velocity-field quality checks (maybe implemented as a utility rather than in the URIS C++ code itself) and the motion interpolation utility?

Code/Source/solver/fluid.cpp
aabrown100-git: Ah okay, maybe `urisValveVelTermTotalEl` then? To distinguish it from "factor" which refers to the term without the velocity?

Code/Source/solver/fluid.cpp
aabrown100-git: Okay makes sense. Just renaming that term should be fine.

tests/cases/uris/pipe_uris_cfd/result_005.vtu
aabrown100-git: Oh excellent, I'm satisfied then

Code/Source/solver/uris.cpp
aabrown100-git: Okay sounds good! If the velocity-field quality check is too complicated/not worth it, maybe just add the motion interpolation utility now since you already have that and that seems very useful for this PR. If you decide to add it in this PR, can you also update the `pipe_uris_cfd` README to link to it?

Code/Source/solver/fluid.cpp
hanzhao2020: Yes, `Term` is less confusing!

Code/Source/solver/fluid.cpp
hanzhao2020: Updated!

Code/Source/solver/uris.cpp
hanzhao2020: @aabrown100-git Sounds good! I added a Python script to the utilities folder for valve motion interpolation and a README file. Please have a look if it makes sense.

utilities/README.md
aabrown100-git, line 1: Would you mind adding short descriptions for the other utilities in this folder? Just for completeness

Code/Source/solver/uris.cpp
aabrown100-git: Looks perfect!

aabrown100-git (approved):
Approved pending my minor comment about the utilities README

utilities/README.md
hanzhao2020: @aabrown100-git Sure! Just updated the README file in the `utilities` folder, please have a look if the descriptions are accurate.

aabrown100-git: Looks great thanks!

ktbolt (approved):
Looks good !

PR #569, opened 2026-06-19
Add scaffold surface support for URIS valves

### LINKED ISSUE #566: Add support for scaffold surfaces in URIS valves
### Problem

The current URIS valve implementation only supports valve surfaces with prescribed input motion. The valve moves when the opening or closing condition is met. However, it does not support scaffold surfaces in the heart valve geometry.

### Solution

For heart valve geometries with a scaffold, it would be good to include the scaffold surface in the simulation to provide sealing around the valve. The scaffold surface should remain fixed throughout the simulation and add a resistive force on the fluid.

### Additional context

_No response_

### Code of Conduct

hanzhao2020:
Resolves https://github.com/SimVascular/svMultiPhysics/issues/566

## Release Notes 
* Added optional scaffold surface support to URIS valve via `Scaffold_file_path`
* Computed scaffold surface unsigned distance function (UDF) on fluid nodes using existing URIS SDF routines
* Scaffold surfaces remain fixed during the simulation and add a resistive force to the fluid

## Testing
* Updated the URIS CFD test case to include a scaffold surface


Code/Source/solver/uris.cpp
```diff
@@ -607,6 +643,47 @@ void uris_read_msh(Simulation* simulation) {
...
+      scaffold_mesh.gN = 0;
+      scaffold_mesh.lN.resize(scaffold_mesh.nNo);
+      scaffold_mesh.lN = 0;
+      for (int a = 0; a < scaffold_mesh.nNo; a++) {
+        scaffold_mesh.gN(a) = b;
+        scaffold_mesh.lN(b) = a;
+        b++;
+      }
+      // Remap msh%gIEN array
+      scaffold_mesh.nEl = scaffold_mesh.gnEl;
```
ktbolt: @hanzhao2020 This remaps `scaffold_mesh.IEN`.

Code/Source/solver/uris.cpp
```diff
@@ -607,6 +643,47 @@ void uris_read_msh(Simulation* simulation) {
     uris_obj.clsFlg = param->valve_starts_as_closed();
     uris_obj.invert_normal = param->invert_normal();
     uris_obj.include_uris_velocity = param->include_uris_velocity();
+    std::string scaffold_file_path = param->scaffold_file_path();
+
+    if (scaffold_file_path != "") {
+      uris_obj.scaffold_flag = true;
+      uris_obj.scaffold_msh.lShl = true;
+      vtk_xml::read_vtu(scaffold_file_path, uris_obj.scaffold_msh);
```
ktbolt: @hanzhao2020 You should check if `vtk_xml::read_vtu` throws an exception and re-throw with a message saying the that scaffold surface can't be read.

Code/Source/solver/uris.cpp
```diff
@@ -1037,38 +1114,32 @@ void uris_calc_sdf(ComMod& com_mod) {
...
-    double extra_val = 0.1;  // [HZ] The BBox is 10% larger than the actual valve, default is 0.1
-    for (int i = 0; i < nsd; i++) {
-      for (int j = 0; j < uris_obj.x.ncols(); j++) {
-        double val = uris_obj.x(i,j);
-        if (val < minb(i))
-            minb(i) = val;
-        if (val > maxb(i))
-            maxb(i) = val;
+      if (cm.idcm() == 0) {
+        std::cout << "Recomputing SDF for " << uris_obj.name << std::endl;
```
ktbolt: @hanzhao2020 These should be `DebugMsg` statements; need to only display this information when you are debugging and disable it before the PR.

ktbolt (changes requested):

Code/Source/solver/uris.cpp
hanzhao2020: @ktbolt An exception is added to check whether the mesh has been read

Code/Source/solver/uris.cpp
hanzhao2020: @ktbolt Thanks for catching that! Fixed.

Code/Source/solver/uris.cpp
hanzhao2020: @ktbolt Thanks, I moved the message to `DebugMsg` under the existing `debug_uris_calc_sdf` guard.

ktbolt (commented):
@hanzhao2020 These changes look good !

However, there are still lots of messages printed to `std::cout` in `uris.cpp`; please replace those with `DebugMsg` calls.

hanzhao2020:
> @hanzhao2020 These changes look good !
> 
> However, there are still lots of messages printed to `std::cout` in `uris.cpp`; please replace those with `DebugMsg` calls.

@ktbolt Sounds good! Replaced all the URIS printed messages with `DebugMsg`.

ktbolt (approved):
@hanzhao2020 Good !

Note that if you would like to output useful run-time statistics for uris then you could create a log file for it, something like `SimulationLogger.h`, that could be activated from a parameter in the solver XML file.

hanzhao2020:
@ktbolt Thanks for the suggestion, I'll keep that in mind!

Code/Source/solver/ComMod.h
```diff
@@ -1553,6 +1553,15 @@ class urisType
...
     std::vector<mshType> msh;
 
+    // Scaffold mesh flag
+    bool scaffold_flag = false;
+    // Scaffold mesh data
+    mshType scaffold_msh;
+    // Unsigned distance function (UDF) for scaffold mesh
+    Vector<double> scaffold_udf;
+    // Flag to indicate if the UDF for scaffold mesh is computed
+    bool scaffold_udf_computed = false;
```
michelebucelli: I suggest to turn these comments into Doxygen-style documentation comments. I think it might be a good idea to do the same for other members of this class, even if they're otherwise untouched by this PR.

Additionally, I think it might be useful to document what the scaffold is and what its purpose is, for future reference and for other developers. I think a good way of doing that is to still use Doxygen comments, in one of the following ways:
1. in the general documentation of the `urisType` class, e.g.
    ```cpp
    /// @brief Unfitted Resistive Immersed surface data type
    ///
    /// ...some generalities about the class...
    ///
    /// ### Scaffolding
    ///
    /// ...what a scaffold is, what is its purpose, what options are 
    /// supported, and how to enable it...
    ```
2. in a Doxygen section grouping these members, e.g.
    ```cpp
    /// @name Scaffold mesh
    /// ...what a scaffold is, what is its purpose, what options are 
    /// supported, and how to enable it...
    /// @{

    /// Scaffold mesh flag
    bool scaffold_flag = false;

    // ...other members...

    /// @}
    ```

I personally slightly prefer option 1, as it gives the chance to expand a bit on the general class documentation as well.

Code/Source/solver/uris.cpp
```diff
@@ -1264,6 +1375,16 @@ void surface_element_barycenter(const urisType& uris_obj, int jM, int Ec, Vector
...
 
+/// @brief Barycenter of a fixed shell element using mesh coordinates.
+void mesh_element_barycenter(const mshType& mesh, const int Ec, Vector<double>& xb) {
+  xb = 0.0;
+  for (int a = 0; a < mesh.eNoN; a++) {
+    const int Ac = mesh.IEN(a, Ec);
+    xb = xb + mesh.x.rcol(Ac);
+  }
+  xb = xb / mesh.eNoN;
+}
```
michelebucelli, line 1403: Since this function is only supposed to return a single value, I suggest to use that as the return value instead of an output argument, by changing the signature to
```cpp
Vector<double>
mesh_element_barycenter(const mshType& msh, const int Ec)
```
Additionally, I suggest renaming `Ec` to something more explicit like `element_index`, to make its meaning more obvious.

Code/Source/solver/uris.cpp
```diff
@@ -607,6 +672,55 @@ void uris_read_msh(Simulation* simulation) {
...
+        }
+      }
+      scaffold_mesh.gIEN.clear();
+      
+      # ifdef debug_uris_read_msh
+      dmsg << "Scaffold mesh is included for: " + uris_obj.name << std::endl;
+      dmsg << "Scaffold mesh nodes: " + std::to_string(uris_obj.scaffold_msh.gnNo) << std::endl;
+      dmsg << "Scaffold mesh elements: " + std::to_string(uris_obj.scaffold_msh.gnEl) << std::endl;
+      # endif
+    }
```
michelebucelli, line 753: I feel like the readability of this piece of code would benefit from being encapsulated into a function with a meaningful name, such as `load_scaffold_from_file`. I imagine similar, if not identical, instructions are used to load the "normal" surfaces, in which case the function should be designed to be used in both cases.

In a dream world, this function would be a method of a class representing the surface itself (e.g. `scaffold.load_from_file(...)`).

Code/Source/solver/uris.cpp
```diff
@@ -1291,6 +1412,22 @@ void uris_find_closest_face_centroid(const urisType& uris_obj, const Vector<doub
   }
 }
 
+/// @brief Find the closest fixed mesh element centroid to xp.
+void uris_find_closest_mesh_centroid(const mshType& mesh, const Vector<double>& xp,
+                                     const int nsd, double& minS, int& Ec,
+                                     Vector<double>& xb) {
```
michelebucelli: Here I suggest expanding the documentation to explain the meaning of each parameter using the Doxygen syntax `@param[in]` and `@param[out]`.

I also suggest renaming the function to `find_closest_element_centroid`, since it seems more clear to me.

Code/Source/solver/uris.cpp
```diff
@@ -1291,6 +1412,22 @@ void uris_find_closest_face_centroid(const urisType& uris_obj, const Vector<doub
...
+  Vector<double> elem_centroid(nsd);
+  for (int e = 0; e < mesh.nEl; e++) {
+    mesh_element_barycenter(mesh, e, elem_centroid);
+    const double dS = std::sqrt((xp - elem_centroid) * (xp - elem_centroid));
```
michelebucelli: This should be equivalent to
```suggestion
    const double dS = utils::norm(xp - elem_centroid);
```

michelebucelli (commented):
Thanks @hanzhao2020! I left a few comments.

My one general concern is related to the comment I made to #566. What I meant by that comment is that, as far as I can see, in principle one could already have scaffolding with the current implementation (i.e. before this PR), by processing the input surface as follows:
1. define a displacement field equal to zero on the scaffold mesh;
2. append the scaffold mesh and the valve mesh (just append, no need to fix the connectivity in principle);
3. give the resulting surface in input to URIS.

Is there some part of this that doesn't actually work as I would think?

hanzhao2020:
> Thanks @hanzhao2020! I left a few comments.
> 
> My one general concern is related to the comment I made to #566. What I meant by that comment is that, as far as I can see, in principle one could already have scaffolding with the current implementation (i.e. before this PR), by processing the input surface as follows:
> 
> 1. define a displacement field equal to zero on the scaffold mesh;
> 2. append the scaffold mesh and the valve mesh (just append, no need to fix the connectivity in principle);
> 3. give the resulting surface in input to URIS.
> 
> Is there some part of this that doesn't actually work as I would think?

@michelebucelli Thank you for the review and comments!

Yes, with the current implementation on the main branch, it is already possible to include a scaffold by treating it as an additional valve leaflet and providing motion data that keeps it fixed throughout the simulation. However, this approach requires the user to provide motion data for both the opening and closing phases (even though they are identical for a stationary scaffold), and the motion files must have the same number of time steps as those of the valve leaflets. This adds unnecessary effort and could be confusing for users. Also, when the scaffold is treated as another valve leaflet, the SDF is recomputed during valve opening and closing, even though the scaffold itself does not move. These computations for scaffold SDF are unnecessary and introduce a bit additional computational cost.

I think supporting the scaffold as an optional input, without requiring any motion data, provides a cleaner and more user-friendly interface while also avoiding these unnecessary SDF updates.

Code/Source/solver/ComMod.h
hanzhao2020: @michelebucelli Thanks for catching that, I updated the comments to Doxygen-style comments.

Code/Source/solver/uris.cpp
hanzhao2020: @michelebucelli Updated. Thank you for the suggestions!

Code/Source/solver/uris.cpp
hanzhao2020: Yes, that's a good idea! Updated.

Code/Source/solver/uris.cpp
hanzhao2020: Added comments to the input and output parameters with the Doxygen style comments.

Code/Source/solver/uris.cpp
hanzhao2020: Updated!

Code/Source/solver/uris.cpp
```diff
@@ -604,6 +603,69 @@ bool uris_point_in_bbox(const Vector<double>& xp, const Vector<double>& minb,
...
+                             const std::string& scaffold_file_path) {
+  uris_obj.scaffold_flag = true;
+  auto& scaffold_mesh = uris_obj.scaffold_msh;
+
+  try {
+    load_shell_mesh_from_file(simulation, scaffold_mesh, scaffold_file_path, true);
+  } catch (const std::exception& e) {
+    throw std::runtime_error("Failed to read URIS scaffold mesh for '" + uris_obj.name + "': " + e.what());
+  } catch (...) {
+    throw std::runtime_error("Failed to read URIS scaffold mesh for '" + uris_obj.name + "'.");
```
michelebucelli: I think it might be good to use the new exception framework introduced in #526 and #537 here.

I have introduced a `FileNotFoundException` class in #577, so maybe we can wait until that gets merged before using it here, since it seems appropriate. If so, I suggest putting a todo comment as a reminder for that (I mostly use the syntax `@todo[username]`, as in `@todo[michelebucelli]`, so that I can find them easily later, and if someone comes across it they easily know who to blame 😅)

Code/Source/solver/uris.cpp
```diff
@@ -604,6 +603,69 @@ bool uris_point_in_bbox(const Vector<double>& xp, const Vector<double>& minb,
...
+  nn::select_ele(simulation->com_mod, mesh);
+
+  if (check_element_ordering) {
+    read_msh_ns::check_ien(simulation, mesh);
+  }
+}
+
+/// @brief Load and initialize the optional URIS scaffold mesh.
+void load_scaffold_from_file(Simulation* simulation, urisType& uris_obj,
+                             const std::string& scaffold_file_path) {
```
michelebucelli: It seems to me that `uris_obj` is used by this function only to
1. choose where to store the result, and
2. construct the error message.

I think it would be better to have a more minimal interface, which would facilitate decoupling this function from the rest of the code.

Is `mshType` copy- or move-constructible? If so, this function could have the following cleaner interface:
```cpp
mshType
load_scaffold_from_file(Simulation* simulation,
                        const std::string& scaffold_file_path,
                        const std::string& uris_name)
```
Ideally, the documentation would explain that the `uris_name` argument is only used for the purpose of error messages (although honestly I'd remove this dependency too and assume that the user can figure out what valve is giving problems based on the file name alone).

In a similar vein, also taking the whole `Simulation` object in input, and as a non-`const` pointer, seems a lot. I see that it is only used as an argument to `load_shell_mesh_from_file`, which in turn forwards it to other methods, so perhaps it's not so easy to drop it altogether. However, I'd try at least to make it `const`, propagating the `const`ness down the call stack.

michelebucelli (commented):
Thanks @hanzhao2020! A couple more suggestions. Other than that looks good to me!

Code/Source/solver/uris.cpp
hanzhao2020: Sounds good! I will update this after `FileNotFoundException` is merged.

Code/Source/solver/uris.cpp
hanzhao2020: Thanks @michelebucelli, that makes sense. I updated this to decouple the helper from `urisType`. `load_scaffold_from_file` now returns the loaded `mshType`, and the caller stores it in `uris_obj.scaffold_msh`. I kept `Simulation` non-const for now to keep this change limited to uris.cpp. Making `Simulation` const can be handled separately, as it would require changes to a number of downstream functions.

michelebucelli (approved):
Looks good to me!

PR #574, opened 2026-06-24
GMRES sets success flag to true if the initial guess is already converged

### LINKED ISSUE #573: Suppress warning "The linear system solution has not converged" if no linear solver iterations are needed
### Problem

It sometimes happen that the linear solver does not need to run any iteration, because the residual for the initial guess is already below the tolerance. This is checked by an `if` statement in `gmres_s` and `gmres_v`. If this early termination condition is verified, the flag `ls.suc` is not set to `true`, so that the caller interprets this behavior as a linear solver failure, and a warning is printed.

The warning is a bit annoying, but more importantly it might be misleading, since the linear solver actually did reach convergence (or better, it was already at convergence for the provided initial guess).

### Solution

I propose setting `ls.suc = true` in the early termination branch of `gmres_s` and `gmres_v`, to suppress the unnecessary/incorrect warning.

### Additional context

- This behavior appeared for me due to inconsistent setting of non-linear and linear solver tolerances. The linear solver tolerance was accidentally higher than the non-linear one, so that the linear solver would run no iterations and the non-linear solver would stagnate. I found the warning slightly misleading in this context, as I thought something was wrong with the linear system while in fact I had just set the tolerances incorrectly.
- I have encountered this behavior for the GMRES solver, but I haven't checked if the same might happen for other solvers.

### Code of Conduct

michelebucelli:
Fixes #573:
1. the success flag is now set to true in GMRES if the initial guess is already good enough, so that no "solution has not converged" warnings are raised;
2. the success flag itself is renamed from `suc` to `success` for improved readability.


michelebucelli:
Partially related to this: I asked AI to double-check that I'd caught all instances of `suc` to be renamed to `success` (I don't have PETSc on my local build so I might have missed some), and it observed that there's a `solver/gmres.cpp` file that is nearly identical to `linear_solver/gmres.cpp`, but is never compiled (it's not listed in `solver/CMakeLists.txt`).

Perhaps it is a leftover from an older organization of the code? In which case, it might be worth removing it to clean it up, as part of this PR.

It also flagged a number of similarly dead code in the class `lsType` from `ComMod.h`.

Code/Source/linear_solver/gmres.cpp
zasexton, line 65: should the internal `gmres()` helper above get the same treatment? It is used by `ns_solver.cpp` and still throws when `err[0] == 0.0`, and does not early-return when the initial residual is nonzero but already `<= absTol`. The error message says the zero norm is probably from ill-posed boundary conditions but wouldn't this also be the case for the perfectly converged solution? In any case, this is just an extra thought. We can open up a future issue if this comment has any merit.

zasexton (approved):
These changes look complete and thoroughly address the tagged issue #573

zasexton:
@aabrown100-git @ktbolt do either of you have any remaining comments to add to this PR?

Code/Source/linear_solver/gmres.cpp
michelebucelli: I'm honestly not sure about this - I'm not sure I understand exactly what the function is doing. The error message says that the **matrix** has zero norm though, which seems to be a different kind of situation than the one addressed here.

zasexton: I'll make a note of this to come back to in the future then :) Otherwise I believe this PR should be merged

aabrown100-git (approved):
Read through the issue discussion and these changes. All looks good!

michelebucelli:
I think this is ready to merge! Unless @ktbolt you have further comments.

Issue #519, opened 2026-03-18
Object oriented refactor of active stress with optional spatially varying parameters


dseyler:
### Problem

The current active stress implementation is embedded inside inside mat_models.cpp where directional stress distribution and EP coupling are all mixed into the constitutive update flow. As discussed in a few other issues, it would be nice to expand the active stress implementation to allow more flexibility, i.e spatially varying stress distributions, magnitude, and temporal delay, but this would be unmanageable in the current procedural code.

@zasexton Let me know if you have any insights on how best to do this.

### Solution

I propose refactoring active stress handling into an object-oriented module/class (active_stress.h /.cpp) that provides an interface for spatially varying element-wise parameter lookup, similar to @aabrown100-git's implementation of spatially varying robin BCs. In the proposed solution, compute_pk2cc will request parameters from a spatial field object and apply them instead of directly reading global values. This will keep future additions to active stress contained in its own file, separate from the constitutive math. 

Element-wise parameters can be supplied in a VTU path under <Fiber_reinforcement_stress>. I think it makes the most sense to have a single VTU contain all of the spatially varying active stress fields, and those that are not supplied in the VTU will be assigned default values.

Example:
```
Fiber_reinforcement_stress type="Unsteady">
  <Temporal_values_file_path> LV_stress.dat </Temporal_values_file_path>
  <Spatial_values_file_path> eta.vtu </Spatial_values_file_path>
</Fiber_reinforcement_stress>
```

I drafted an implementation for this feature with spatially varying directional active stress distributions in [this branch](https://github.com/dseyler/svMultiPhysics/tree/active_stress).

Let me know if anyone else would find this useful or if there are any other features that folks would like included for active stress. My current list is:

- spatially varying directional distribtuion (eta_f, eta_s, eta_n)
- spatially varying stress magnitude
- spatially varying temporal delay (i.e. one way coupling) @javijv4 Maybe this could be combined with the eikonal equation solver you suggested.
- active stress "fiber" direction independent from constitutive "fiber" direction

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct and Contributing Guidelines

dseyler:
Here's a more detailed description of the changes [in my branch](https://github.com/dseyler/svMultiPhysics/tree/active_stress). Most of these changes have been focused around adding spatially varying directional stress distributions.

Previously, directional active stress was computed inline in mat_modesl.cpp using only global scalars:
```
[...]

PR #577, opened 2026-06-28
Object oriented refactoring of fcType

### LINKED ISSUE #575: Object oriented refactoring of fcType
### Problem

As part of my work on #519, I was going to implement a concrete active stress class (`UniformUnsteadyActiveStress`) that replicates what is currently implemented by the `<Fiber_reinforcement_stress>` tag, that is reads a time-dependent active tension transient from file.

It seems to me that the natural way of doing this is to place an object of type `fcType` inside `UniformUnsteadyActiveStress`, construct that object as part of the active stress object initialization, and evaluate it as part of the active tension evaluation method.

However, I encountered a few issues which I think are worth addressing in a more focused refactoring of `fcType`:
1. `fcType` is defined in `ComMod.h`, but I cannot include `ComMod.h` in `active_stress.h` (or derived), due to circular dependencies. This can probably be worked around with forward declarations and access through a pointer, but it feels needlessly convoluted.
2. `ifft` takes in input a `ComMod` instance, even though it only uses its `time` member. It would be more natural for `ifft` to take the evaluation time directly.

### Solution

The issues above prompted me to think about a cleaner restructuring of the whole `fft`, `ifft` and `fcType` infrastructure, which I think is worth addressing independently of #519, since it's more general.

I propose the following changes:
1. move `fcType` to its own file, so that it can be included and used independently of `ComMod.h`;
2. encapsulate reading data from file and calling `fft` as a method of the `fcType` class; this might be done either in the class constructor or in an initialization method (I lean in favor of the constructor); 
3. modify `ifft` so that it only takes the desired time in input, not the whole `ComMod`;
4. encapsulate evaluating the interpolated data at a given time point as a method of the `fcType` class, say `fcType::evaluate`.

After these changes, `fft` and `ifft` would become internal implementation details of `fcType`, so their code could be moved to that class.

Additionally, I suggest taking the opportunity to give the `fcType` class a more explicit name (I figure that `fc` stands for Fourier coefficients, but that might not be immediately obvious to someone unfamiliar with the code). I suggest for example `FourierInterpolation` or `TimeInterpolation` (I slightly lean in favor of the latter, because it leaves open the possibility of expanding the class to include other interpolation methods, e.g. splines or piecewise linear, while retaining the same interface).

### Additional context

_No response_

### Code of Conduct

michelebucelli:
Fixes #575.

## Current situation

The class `fcType` is currently defined in `ComMod.h`, and its interface and that of related function make its use slightly more cumbersome and less seamless than it needs to be (see discussion in #575 for more details).

## Release Notes 

1. `fcType` renamed to `FourierInterpolation`, and its data members (now private) renamed with more explicit names.
2. `FourierInterpolation` implementation moved to dedicated files `fourier_interpolation.{h,cpp}`.
3. Construction of `FourierInterpolation` object is now done through static methods `FourierInterpolation::from_time_series` and `FourierInterpolation::from_fourier_coefficients` (while before it was done "manually" everywhere this was needed).
4. Static methods `FourierInterpolation::from_time_series_file` and `FourierInterpolation::from_fourier_coefficients_file` manage reading data from file (while before it was done "manually" everywhere this was needed).
5. Distribution of data in parallel is done with the (non-static) method `FourierInterpolation::distribute`.
6. Evaluation of the interpolation is done through the methods `FourierInterpolation::value` and `FourierInterpolation::value_and_derivative`.
8. Functions `fft` and `ifft` were tightly related to `fcType` (they worked on an `fcType` instance), and they have been merged into the methods `FourierInterpolation` class (for construction and evaluation, respectively).
7. The rest of the code has been updated to adhere to this new interface.

All this should make it a bit easier to add support for reading time-dependent data from file in new classes (which is part of the aims of #519).

## Documentation

The class and its methods have been given detailed Doxygen documentation. 

## Testing

The unit test and the integration tests all pass with the new interface.

## TODO before marking as ready

- [x] expand suite of FFT-related unit tests


Code/Source/solver/fft.cpp
michelebucelli: In this PR, the functions `fft` and `ifft` have been moved from this file to `fourier_interpolation.cpp`, inside the new `FourierInterpolation` class. I think this makes sense, since those functions were tightly coupled to `fcType`.

However, this file also contains another function, `igbc`, which is now the only function left. The function is not related to Fourier transform, so it seems odd to keep the name of this file as it is.

I propose renaming it to `igbc.cpp` (and the header to `igbc.h`).

Code/Source/solver/fourier_interpolation.cpp
```diff
@@ -0,0 +1,563 @@
...
+  // Linear trend.
+  for (int i = 0; i < n_components; ++i) {
+    value[i] = linear_trend_initial_values[i] + t * linear_trend_slopes[i];
+
+    // @todo[michelebucelli] This is consistent with the old code, but is
+    //   incorrect when use_ramp = true. In that case, the derivative should be
+    //   zero outside the interpolation interval [initial_time, initial_time +
+    //   period].
+    if (evaluate_derivative)
+      derivative[i] = linear_trend_slopes[i];
```
michelebucelli: This was probably an unintended behavior of the old code.

When using the "ramp" option, the following function is computed (the implementation is formulated slightly different):

$$
v(t) = \begin{cases}
y_0 & \text{if } t < t_0 \\
y_0 + (y_{N-1} - y_{0}) \frac{t - t_0}{t_{N-1} - t_0} & \text{if } t_0 \leq t \leq t_{N-1} \\
y_{N-1} & \text{if } t > t_{N-1}
\end{cases}
$$

For times outside of the inerpolation interval $[t_0, t_{N-1}]$, I think the derivative should be set to zero:
```suggestion
    if (evaluate_derivative) {
      if (use_ramp && (time < initial_time || time > initial_time + period)) {
        derivative[i] = 0.0;
      } else {
        derivative[i] = linear_trend_slopes[i];
      }
    }
```

I tried this out and all the tests still pass, so I propose changing this behavior.

Code/Source/solver/FourierInterpolation.cpp
```diff
@@ -0,0 +1,563 @@
...
+        // Using value[j] = value[j] + ... instead of value[j] += ..., because
+        // the latter changes the order of operations enough to break some of
+        // the tests.
+        // @todo[michelebucelli] This seems pretty fragile! It happens in a few
+        //   other places in this file as well, where using an increment
+        //   operator (*= or +=) changes the order of operations resulting in
+        //   changes in the test values. This should be investigated and the
+        //   best (i.e. most accurate) operation order should be chosen.
+        value[j] = value[j] + fourier_coefficients_real(j, i) * std::cos(K) -
+                   fourier_coefficients_imaginary(j, i) * std::sin(K);
```
michelebucelli, line 593: This took me a while to figure out 😅 There's at least another place where the same thing happens. I'm honestly not sure which is the most correct option, but the fact that the order of operations makes a measurable difference probably means that this needs to be carefully evaluated.

tests/unitTests/math_tests/test_fourier_interpolation.cpp
```diff
@@ -0,0 +1,297 @@
...
+        temporal_values[0][2] +
+        (temporal_values.back()[2] - temporal_values[0][2]) * (t - x_start) /
+            (x_end - x_start);
+
+    ASSERT_NEAR(value[0], expected_y, 1e-6)
+        << "Expected value in interpolation interval to be linear "
+           "interpolation of initial and final value";
+    ASSERT_NEAR(value[1], expected_z, 1e-6)
+        << "Expected value in interpolation interval to be linear "
+           "interpolation of initial and final value";
```
michelebucelli, line 307: These assertions should be extended to include checks on the evaluated derivative, once we agree whether to change or keep the current behavior as per my previous comment.

Code/Source/solver/fft.cpp
kko27: @michelebucelli yes, I agree with this change. This function seems to be for assigning spatially varying time profiles

Code/Source/solver/fourier_interpolation.cpp
```diff
@@ -0,0 +1,622 @@
...
+    }
+  }
+
+  Vector<double> times(n_time_points);
+  Array<double> values(n_components, n_time_points);
+
+  for (unsigned int i = 0; i < n_time_points; ++i) {
+    times[i] = temporal_values[i][0];
+    for (unsigned int j = 0; j < n_components; ++j) {
+      values(j, i) = temporal_values[i][j + 1];
```
kko27: I think this was also in the previous version of the code, but would you know why we transpose the dimensions of `values`?

Code/Source/solver/fft.cpp
ktbolt: @michelebucelli You might could put `igbc` in `utils.h,cpp` or `all_fun.h,cpp`  depending on I suppose if it is `fun` or not.

tests/unitTests/math_tests/test_fourier_interpolation.cpp
kko27, line 1: I couldn't figure this out from these unit tests, but do we have tests for multiple spatial dimensions?

Code/Source/solver/FourierInterpolation.cpp
```diff
@@ -0,0 +1,622 @@
...
+  result.fourier_coefficients_real.resize(n_components,
+                                          result.n_fourier_coefficients);
+  result.fourier_coefficients_imaginary.resize(n_components,
+                                               result.n_fourier_coefficients);
+
+  // Compute the linear trend part.
+  for (unsigned int j = 0; j < n_components; ++j) {
+    result.linear_trend_initial_values[j] = values(j, 0);
+    result.linear_trend_slopes[j] =
+        (values(j, n_time_points - 1) - values(j, 0)) / result.period;
```
ktbolt, line 112: @michelebucelli Should you check if `result.period` is zero ?

Code/Source/solver/fourier_interpolation.cpp
```diff
@@ -0,0 +1,622 @@
...
+    result.fourier_coefficients_real.set_col(n, 0.0);
+    result.fourier_coefficients_imaginary.set_col(n, 0.0);
+
+    for (unsigned int i = 0; i < n_time_points - 1; ++i) {
+      const double ko = 2.0 * consts::pi * tmp * times[i] / result.period;
+      const double kn = 2.0 * consts::pi * tmp * times[i + 1] / result.period;
+
+      for (unsigned int j = 0; j < result.n_components; j++) {
+        const double s =
+            (values(j, i + 1) - values(j, i)) / (times[i + 1] - times[i]);
```
ktbolt: @michelebucelli Why all of these `const` ? Maybe a (good?) habit to guard against changing these variables later I guess.

Code/Source/solver/fourier_interpolation.cpp
```diff
@@ -0,0 +1,622 @@
...
+
+  // Compute the Fourier coefficients.
+  for (unsigned int n = 0; n < result.n_fourier_coefficients; ++n) {
+    const double tmp = static_cast<double>(n);
+    result.fourier_coefficients_real.set_col(n, 0.0);
+    result.fourier_coefficients_imaginary.set_col(n, 0.0);
+
+    for (unsigned int i = 0; i < n_time_points - 1; ++i) {
+      const double ko = 2.0 * consts::pi * tmp * times[i] / result.period;
+      const double kn = 2.0 * consts::pi * tmp * times[i + 1] / result.period;
```
ktbolt: @michelebucelli Why are we using `consts::pi` here rather than `std::numbers::pi` ?

Code/Source/solver/fourier_interpolation.cpp
```diff
@@ -0,0 +1,622 @@
...
+    }
+  }
+
+  return result;
+}
+
+FourierInterpolation
+FourierInterpolation::from_time_series_file(const std::string &file_name,
+                                            const unsigned int n_components,
+                                            const bool use_ramp) {
```
ktbolt: @michelebucelli Fundamental types are typically not declared `const`.

Code/Source/solver/FourierInterpolation.cpp
```diff
@@ -0,0 +1,622 @@
...
+            std::to_string(n_time_points) + " time points were provided.");
+  }
+
+  if (n_fourier_coefficients == 0) {
+    svmp::raise<svmp::FileFormatException>(
+        SVMP_HERE, file_name,
+        "At least 1 Fourier coefficient is required to set up Fourier "
+        "interpolation. 0 Fourier coefficients were provided.");
+  }
+
```
ktbolt, line 198: @michelebucelli `n_time_points` could be zero for an incorrect format file format: first value is non-numeric.

And `n_fourier_coefficients` could be zero if the first value is not an int or there is a format error, say a comma is used to separate values.

Code/Source/solver/FourierInterpolation.cpp
```diff
@@ -0,0 +1,622 @@
...
+}
+
+void FourierInterpolation::distribute(const CmMod &cm_mod, const cmType &cm) {
+  // Only the master knows whether the object has been initialized. Therefore,
+  // we broadcast the initialization flag to all ranks, so that the following if
+  // statement can be run correctly by all.
+  bool initialized = defined();
+  cm.bcast(cm_mod, &initialized);
+
+  if (initialized) {
```
ktbolt, line 413: @michelebucelli It is not very clear what it means for an object to be `initialized`. Does that mean it has data ?

Code/Source/solver/FourierInterpolation.cpp
ktbolt: @michelebucelli Very interesting !

What is the `measurable difference` that you are seeing ?

ktbolt (commented):
@michelebucelli Nice work !

We name files containing classes with their class name (I think).

Code/Source/solver/fourier_interpolation.cpp
kko27: This sounds good to me

kko27 (commented):
@michelebucelli This looks good to me! I have some minor comments/questions. Renaming all the Fourier interpolation variables to make them more readable was great

Code/Source/solver/FourierInterpolation.cpp
michelebucelli: Good point. That the period is positive is downstream of assuming that the time points are strictly increasing, but that is not checked anywhere.

I'll add a check on the positivity of the period to `FourierInterpolation::from_fourier_coefficients`, and a check on the monotonicity of the time series to `FourierInterpolation::from_time_series`.

Code/Source/solver/fourier_interpolation.cpp
michelebucelli: Yes, I try to follow [this guideline](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#rconst-immutable) (and the following ones) as much as possible. The idea is that it's more resistant to accidental changes as the code is edited in the future.

Code/Source/solver/fourier_interpolation.cpp
michelebucelli: No reason, as far as I can see. I inherited this from the old implementation.

Happy to replace with `std::numbers::pi` (I didn't know until now that `C++20` had introduced this, I only knew of the old `M_PI`).

I think it would be a good idea to take this chance to do it all over `svMultiPhysics` and remove `consts::pi` altogether. `consts::pi` is only used for Fourier interpolation and for the smoothed delta function in the URIS files, so it's a pretty small change.

Code/Source/solver/FourierInterpolation.cpp
michelebucelli: True, will fix.

In general, now that you mention it, there's a few unguarded reads from streams. I'll add checks on each of them so that they raise interpretable error messages if something goes wrong.

Code/Source/solver/FourierInterpolation.cpp
michelebucelli: Yes, it means that it has no data.

The need for this comes from the previous implementation, where the object would always be constructed as empty, regardless of whether it was needed, and then it would be filled if needed (i.e. if the parameters required it). Accordingly, many operations on the object, including parallel distribution, were guarded by `if (fc_type.defined())` statements, to avoid operating on empty data (I assume for the sake of both safety and efficiency).

In an ideal world, I would have disabled the default construction of `FourierInterpolation`, so that the object would always have data. However, this is incompatible with the current use of `FourierInterpolation` in `bcType`, `bfType` and `fibStrsType`, because all the three classes only have a default constructor.

One solution to this would be a larger refactoring of those classes, but I feel that's a larger effort and a bit out of scope for this PR.

Another solution would be to defer construction of `FourierInterpolation` by storing a `unique_ptr` or `shared_ptr` to it, instead of the object itself. I have a mild dislike for this solution, because of the extra indirection, and because it mostly turns `if (X.defined())` checks into `if (X != nullptr)` checks (although not the onse inside the class, admittedly).

If you prefer, I'd be happy to switch to pointers.

Code/Source/solver/FourierInterpolation.cpp
michelebucelli: It is "measurable" in the sense that, if changing this statement to the more idiomatic `value[j] += ...`, the test `fluid/carreau_yasuda` fails with 4 processors (but not with 1 or 3 processors).

The reported mismatches are as follows:
```
AssertionError: Test failed in field Velocity. Results differ by more than rtol=1e-07 in 19.3% of results. Max. rel. difference is 2.6e-06 (abs. 2.7e-06)
Test failed in field Pressure. Results differ by more than rtol=1e-06 in 32.0% of results. Max. rel. difference is 1.7e-03 (abs. 2.2e-03)
Test failed in field Traction. Results differ by more than rtol=1e-06 in 27.2% of results. Max. rel. difference is 1.2e-03 (abs. 1.8e-03)
Test failed in field WSS. Results differ by more than rtol=1e-08 in 52.4% of results. Max. rel. difference is 5.1e-07 (abs. 5.2e-07)
Test failed in field Vorticity. Results differ by more than rtol=1e-07 in 48.4% of results. Max. rel. difference is 3.3e-06 (abs. 3.4e-06)
Test failed in field Divergence. Results differ by more than rtol=1e-09 in 98.6% of results. Max. rel. difference is 1.2e-06 (abs. 1.2e-06)
```
The relative differences are not huge, but still the worst case is in the order of $10^{-3}$, which seems pretty big for a change that only affects the order of operations.

My interpretation of this is that this is a delicate operation in terms of possible error propagation, and the order of operations should be evaluated carefully, although I'm not sure how this would translate in practice to this particular statement.

Unless there's something peculiar about that specific test that amplificates the issue - I haven't investigated too deeply.

Code/Source/solver/fourier_interpolation.cpp
michelebucelli: I inherited this from the previous code without thinking too much about it, so I don't really know.

It might be that this was done in the interest of memory layout efficiency, but if that were the case then the gains are probably minimal, especially considering that this function is called only once at startup.

(Related to this, I have half a mind to change this function's signature to take directly a `Vector` for the times and an `Array` for the values, saving us this copy and transposition).

tests/unitTests/math_tests/test_fourier_interpolation.cpp
michelebucelli: `FourierInterpolation` only interpolates functions from $\mathbb{R}$ to $\mathbb{R}^n$, where $n$ is arbitrary. The input dimension is always 1 (and it is interpreted as time).

The unit test does check that the interpolation works for $n = 2$ (with the assumption that, if that is true, then it works for arbitrary $n$).

Code/Source/solver/fourier_interpolation.cpp
ktbolt: @michelebucelli Good, change `consts::pi`  to `std::numbers::pi`; little by little the code gets better ...

Code/Source/solver/FourierInterpolation.cpp
ktbolt: @michelebucelli We are indeed held hostage to the current implementation; Fortran allowed for a lot of sketchy behavior. Let's not do a refactoring of a bunch of classes; save that for a later tasks.

 Like you said using pointers still requires the same sort of checks.

When there is an object-oriented framework in place we can think about how to restructure the code.

Code/Source/solver/fourier_interpolation.cpp
michelebucelli: I have turned the inputs of `from_time_series` into a `Vector` and an `Array`, so that this conversion and transposition is no longer necessary.

michelebucelli:
Thanks for the comments @ktbolt @kko27! I think I have addressed all of them.

@ktbolt I haven't renamed the files to `FourierInterpolation.{h,cpp}` yet to avoid polluting the diffs in case you want to review again. I'll do that after approval before merging.

kko27 (approved):
Thanks! Looks good to me!

Code/Source/solver/set_bc.cpp
```diff
@@ -1068,17 +1067,16 @@ void set_bc_dir_l(ComMod& com_mod, const bcType& lBc, const faceType& lFa, Array
...
     #ifdef debug_set_bc_dir_l
     dmsg << "bType_ustd";
     #endif
-    Vector<double> dirY_v(1), dirA_v(1);
-    ifft(com_mod, lBc.gt, dirY_v, dirA_v);
-    dirY = dirY_v(0);
-    dirA = dirA_v(0);
+    const auto value_and_derivative = lBc.gt.value_and_derivative(com_mod.time);
+    dirY = value_and_derivative.first[0];
+    dirA = value_and_derivative.second[0];
```
ktbolt: @michelebucelli It would be cleaner to define `FourierInterpolation::value_and_derivative` to return `std::tuple<Vector<double>, Vector<double>>` then you could write
```
const auto [dirY, dirA] = lBc.gt.value_and_derivative(com_mod.time);
```

ktbolt (commented):
@michelebucelli Looks good, just a single suggestion.

Code/Source/solver/set_bc.cpp
michelebucelli: The same syntax can actually be used for `std::pair` as well! The reason why I wasn't doing it here is that `dirY` and `dirA` are `double`s and not `Vector<double>`, so the direct assignment is not possible.

Nonetheless, I replaced these lines with:
```cpp
    const auto [value, derivative] = lBc.gt.value_and_derivative(com_mod.time);
    dirY = value[0];
    dirA = derivative[0];
```
which is more readable than before.

ktbolt (approved):
Looks good !

Issue #549, opened 2026-05-19
Follow-up polishing from #540 and #578 (Object oriented refactoring of ionic models and active stress)


michelebucelli:
Follow up to threads from #540 and #578 that were not resolved before merging.

- [ ] Moving ionic model parameter classes to `ionic_model.h`/`ionic_model.cpp`, to improve decoupling between ionic models and the rest: [original thread](https://github.com/SimVascular/svMultiPhysics/pull/540#discussion_r3197568472).
- [ ] Remove code duplication between `ParameterLists` and ionic model parameter classes: [original thread](https://github.com/SimVascular/svMultiPhysics/pull/540#discussion_r3197583279).
- [ ] Identifying ionic models only through either `std::string`s or the `enum class ElectrophysiologyModelType`, but not a mix of both: [original thread](https://github.com/SimVascular/svMultiPhysics/pull/540#discussion_r3197416139).
- [ ] Move the ionic state variables into `IonicState`: [original thread](https://github.com/SimVascular/svMultiPhysics/pull/578#discussion_r3494484176).
- [ ] The parameter section for the selected ionic or active stress model should be required (whereas now the section can either be omitted entirely or reported in full): [original thread](https://github.com/SimVascular/svMultiPhysics/pull/578#discussion_r3605148529)

samibismar:
I’ve implemented the ionic-model identification and state-ownership cleanup discussed here on this [branch](https://github.com/samibismar/svMultiPhysics/tree/ionic-model-cleanup)

The changes:
* Use string identifiers consistently, preserving existing model names and aliases.
* Move state storage, initialization, and advancement into IonicModel.
* Keep each domain’s internal states and gates separate at shared nodes instead of averaging them.
* Update solver integration, output, and restart handling to use the model-owned state.

Voltage remains synchronized with the PDE solution. Calcium is still averaged for active-stress coupling and output, but that average is not copied back into the models’ internal states. Parameter-related work remains deferred.

All 10 cellular trajectory tests pass. The regression results after rebasing onto main are:

>CEP: 24 passed, 12 failed.
>Electromechanics: all 6 passed.

The 12 failures come from spiral_TTP_2d, cylinder_purkinje_1d3d, and slab_domains (both .dat and VTU domain inputs), each run on 1, 3, and 4 MPI ranks.

[...]

Issue #96, opened 2023-09-12
Automatically enforce coding standards


mrp089:
We currently have a ["manual" definition](https://github.com/SimVascular/svFSIplus/tree/main/Code/Source#coding_standards) of `svFSIplus` coding standards. I think it would be helpful to automatize what is written here with the widely used C++ code formatter `clang-format` ([see usage here](https://stanfordcbcl.github.io/svZeroDPlus/cpp/md_developer_guide.html#autotoc_md38)).

It seems like what we have is pretty much the [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html). If we really need to, we can define modifications in a [custom format file](https://releases.llvm.org/3.6.0/tools/clang/docs/ClangFormatStyleOptions.html).

Once defined (and all existing code is formatted to that), we can only allow commits that satisfy this format [as we do in `svZeroDPlus`](https://github.com/StanfordCBCL/svZeroDPlus/blob/master/.github/workflows/codechecks.yml). This will save developers time because they can automatically format their code and avoid future discussions about formatting when merging new code.

@ktbolt, please let me know what modifications from the Google C++ style you would like, and I can set up the rest. Thanks!

mrp089:
Adding a linter will be especially helpful now since we're moving more into refactoring. Every well-maintained, community-facing code has this:
- [PETSc](https://github.com/petsc/petsc/blob/main/.clang-format)
- [deal.II](https://github.com/dealii/dealii/blob/master/.clang-format)
- [Trilinos](https://github.com/trilinos/Trilinos/wiki/Clang%E2%80%90format)
- [FEniCS](https://github.com/FEniCS/basix/blob/main/.clang-format)
- [MOOSE](https://github.com/idaholab/moose/blob/devel/.clang-format)

This can be done easily with `clang-format` (see #118).

ktbolt:
@mrp089 I'd rather developers not have the format they use to implement code changed after they push code. And lint programs can produce a lot of noise and such that is distracting. 

And I've not seen any problems with not using lint. 

And what other open source project do does not mean that it is best. MOOSE and PETSc don't use CMake so should we not use CMake ?

alisonmarsden:
[...]

PR #578, opened 2026-06-29
Object-oriented refactoring of active stress

michelebucelli:
Partially addresses #519.

## Current situation

1. Coupling of electrophysiology and mechanics through active stress had been removed in #540, pending this refactoring;
2. only one active stress model was implemented, was tightly coupled to specific ionic models, and was not accessible through the user-exposed parameters.

## Release Notes 

Many of the implementation choices here are modeled after the implementation of `IonicModel` in #540. Here, I took some further steps towards encapsulation which might be ported to `IonicModel` as well, although maybe in a separate PR. I'll add some comments to the code about this.

1. The PR introduces the following hierarchy of classes for active stress models, with the idea that new active stress model can be added easily by deriving new classes from `ActiveStress` or `ActiveStressODE`.
```mermaid
---
config:
    class:
      hideEmptyMembersBox: true
---
classDiagram
  ActiveStress <|-- ActiveStressODE
  ActiveStressODE <|-- NashPanfilov
  ActiveStress <|-- UniformSteadyActiveStress
  ActiveStress <|-- UniformUnsteadyActiveStress
  dmnType *-- ActiveStress
  ActiveStressFactory .. ActiveStress : instantiates
  Factory~BaseClass~ <|.. ActiveStressFactory : BaseClass=ActiveStress
```

2. A new abstract `Factory` class is introduced, by templating the `IonicModelFactory`; aliases are provided for both `Factory<IonicModel>` and `Factory<ActiveStress>`.

3. The `ActiveStress` class and derived classes manage their own XML parameters through the abstract class `ActiveStressParameters` and the methods `get_parameters`, `read_parameters`, `distribute_parameters`, so that the addition of model parameters can be localized to the source files where the model is defined.

4. A (shared) pointer to `ActiveStress` is stored in `dmnType`, to allow different active stress models in different domains.

5. Time stepping of `ActiveStress` instances is managed by `Integrator::predictor`. After advancing one time step, the active tension along the three principal directions is stored in the vector `CepMod::cem.Ya_{f,s,n}`, whose entries are then passed on to the constitutive law routines.

6. The XML section for a `struct` domain now can optionally contain the following tag:
   ```xml
   <ActiveStress>
     <Model>Model name (e.g. NashPanfilov, UniformSteady, ...)</Model>

     <Directional_distribution>
       <!-- optional, and same as currently implemented for Fiber_reinforcement_stress -->
     </Directional_distribution>

     <UniformSteady>
       <Value>123</Value>
     <UniformSteady>

     <!-- other sections for every implemented model, if needed -->
   </ActiveStress>
   ```
   If present, the tag enables active stress through the `ActiveStress` class. If the active  stress model is calcium-based, and no electrophysiology equation is present, the calcium will be kept to zero.

## Documentation

New additions are documented via Doxygen comments. Documentation for `ActiveStress` and `ActiveStressODE` also includes a brief list of steps needed to implement a new model.

## Testing

Existing automatic tests pass with the new implementation, although they do not cover the newly added features.

I have tested this on a basic electromechanics example based on `cases/cep/slab_domains`. I attach below a preview of the simulation.
anim
I plan to turn this into an integration test (running one timestep is reasonably fast).

## TODO before marking as ready

- [ ] add test for uniform, steady active stress
- [x] add test for uniform, unsteady active stress
- [x] add test for coupled electromechanics with Nash-Panfilov model


Code/Source/solver/active_stress.h
```diff
@@ -0,0 +1,261 @@
...
+                                       Vector<double> &state) const = 0;
+
+  /**
+   * @brief Compute the active tension for a single node.
+   */
+  virtual double
+  compute_active_tension_local(const Vector<double> &state) const = 0;
+
+  /// State variables for the model.
+  Array<double> states;
```
michelebucelli, line 236: The state vector for the active stress model is stored into the `ActiveStress` instance itself, instead of at a more "global" scope as for the `Xion` state vector of `IonicModel`.

I think this is a slightly better design, and deals better with heterogeneous models across domain interfaces. Accordingly, I think this would be a good design to adopt for `IonicModel` too, although perhaps it's better to do it in a follow-up PR.

Code/Source/solver/active_stress.h
```diff
@@ -0,0 +1,261 @@
...
+  Vector<double> active_tension;
+
+  /// Active tension coefficient along the fiber direction.
+  double eta_f;
+
+  /// Active tension coefficient along the sheet direction.
+  double eta_s;
+
+  /// Active tension coefficient along the sheet-normal direction.
+  double eta_n;
```
michelebucelli, line 248: @dseyler for now these are domain-wide constant, same as in the old code. I think this class would be the perfect place to implement spatially-varying values for these coefficients in a way that makes that feature available to all concrete active stress models.

Code/Source/solver/active_stress_nash_panfilov.h
```diff
@@ -0,0 +1,137 @@
...
+ * @f$\eta_\text{T}@f$, which has the same units of active tension over calcium.
+ * Therefore, if the ionic model providing the calcium is phenomenological (see
+ * @ref IonicModel) and calcium is non-dimensional, this parameter may need to
+ * be rescaled as well. Similar considerations apply to @f$\xi_T@f$.
+ *
+ * **References**:
+ * 1. [Nash, Panfilov (2004)](https://doi.org/10.1016/j.pbiomolbio.2004.01.016)
+ * 2. [Goktepe, Kuhl (2009)](https://doi.org/10.1007/s00466-009-0434-z)
+ */
+class NashPanfilov : public ActiveStressODE {
```
michelebucelli, line 39: Here there's a slight departure from the previous implementation.

In the old code, the Nash-Panfilov model was driven by potential if coupled to the Aliev-Panfilov or Bueno-Orovio models, and by calcium otherwise. I imagine this choice was mostly due to the phenomenological nature of AP and BO, where calcium is not really a variable. However, both models have a variable that resembles calcium transients more closely than the potential does, which they return through their `get_calcium_index` methods.

I chose to only feed the calcium-like variables into the active stress model, and not the potential, to keep the `ActiveStress` interface more lean and closer to what makes biological sense. This means that the behavior of Nash-Panfilov coupled to AP or BO will be slightly different (in my opinion slightly more correct).

Notice that, even before #540, the coupling code was effectively dead code, so there's no test case that is affected by this change in behavior.

Code/Source/solver/active_stress_ode.cpp
```diff
@@ -0,0 +1,15 @@
...
+
+void ActiveStressODE::advance_time_step_local(const double t, const double dt,
+                                              const double calcium,
+                                              const double fiber_stretch,
+                                              const double fiber_stretch_rate,
+                                              Vector<double> &state) const {
+  // Forward Euler time stepping.
+  Vector<double> f = getf(t, state, calcium, fiber_stretch, fiber_stretch_rate);
+
+  state.add(dt, f);
```
michelebucelli: We might want to at least offer the option of a Runge-Kutta method, but it should be relatively straightforward to do that at a later stage.

Code/Source/solver/active_stress_uniform_steady.h
```diff
@@ -0,0 +1,89 @@
...
+/**
+ * @brief Uniform and steady active stress model.
+ *
+ * Defines an active tension that is constant in space and time, i.e.
+ * @f[
+ *   \Tact(t, \calcium, \fiberstretch, \fiberstretchrate, \astressstate) = g\;,
+ * @f]
+ * where @f$g@f$ is a user-defined constant value.
+ */
+class UniformSteadyActiveStress : public ActiveStress {
```
michelebucelli, line 18: @dseyler In my mind, unsteady prescribed active stress can be modeled after this class (of course with a bit of added complications to it, related e.g. to loading from file the time-dependent active tension profile, and possibly the activation times used to shift it).

I think once #577 is merged this should be rather easy.

Code/Source/solver/Parameters.h
```diff
@@ -1318,6 +1318,113 @@ class IonicModelParameters : public ParameterLists {
...
+/// @brief Parameters for a generic active stress model.
+///
+/// This class is meant to be inherited from to implement parameters for
+/// specific active stress models. Derived classes will mostly have to call
+/// add_parameter in their constructor to define the model-specific parameters.
+///
+/// In the XML file, this class, and the classes derived from it, correspond to
+/// the element <Model_name> within the <Active_stress> element, where
+/// Model_name is the name of a concrete active stress model.
+class ActiveStressModelParameters : public ParameterLists {
```
michelebucelli, line 1418: Similar to `IonicModelParameters` ([original thread](https://github.com/SimVascular/svMultiPhysics/pull/540#discussion_r3197583279)), this class might be implementing general-purpose stuff that would probably make more sense in `ParameterList`. I am still a bit confused about the parameter parsing logic, so I think this might be a sub-optimal solution.

I think similar problems are present in other classes in this file (e.g. there's a few classes redefining the `xml_element_name` member although they inherit it from `ParameterLists`), and overall parameter parsing might benefit from a bit of rationalization, but perhaps this is not the right PR for that.

Code/Source/solver/post.cpp
```diff
@@ -1898,8 +1896,13 @@ void tpost(Simulation* simulation, const mshType& lM, const int m, Array<double>
...
             double Ja;
-            
-            mat_models::compute_pk2cc(com_mod, cep_mod, eq.dmn[cDmn], F, nFn, fN, ya, S, Dm, Ja);
+
+            // @todo[michelebucelli] The active tension here was kept to zero,
+            //   so I replicated the same for fibers, sheets and normals. This
+            //   might not be the correct thing to do.
+            mat_models::compute_pk2cc(com_mod, cep_mod, eq.dmn[cDmn], F, nFn,
+                                      fN, /* ya_f = */ 0.0, /* ya_s = */ 0.0,
+                                      /* ya_n = */ 0.0, S, Dm, Ja);
```
michelebucelli: Does anyone have any insight about this?

Maybe the correct interpretation is that the active stress is kept to zero because this code aims at computing only the passive part of thes stress, but I am not sure about it.

Code/Source/solver/active_stress.h
dseyler: Awesome. This looks like it should be pretty straightforward to adapt for an unsteady, optionally spatially varying class

Code/Source/solver/active_stress_nash_panfilov.h
kko27: @michelebucelli thanks for looking into this. I agree with this change as calcium transients regulate the cross-bridging cycle. Probably beyond the scope of this PR, but it would be interesting to see the difference in using AP/BO potential vs AP/BO calcium transients as inputs

Code/Source/solver/active_stress_ode.cpp
kko27: @michelebucelli sure, let's add RK at a later time. I imagine that as we add more ODE models, we may want to use other time-stepping schemes. 

That being said, we may want to make it transparent to the user that Forward Euler time stepping is being used (maybe in the `solver.xml` file, we have `<ODE_solver>` argument).

michelebucelli: Good idea, I'll add the `<ODE_solver>` parameter to the `<Active_stress>` section.

michelebucelli: Done in this commit: c27c1f0c30ae407424770cd2a438413691c90d86

Code/Source/solver/active_stress_uniform_steady.h
michelebucelli: Added uniform unsteady active stress in this commit: f029cd33f5261f57523a7b6ac8a2582829b859df

I also updated the PR description accordingly.

Code/Source/solver/post.cpp
michelebucelli: I ended up changing this so that it uses the local active tension, for compatibility with the old behavior of `Fiber_reinforcement_stress`. I think it makes physical sense if the quantity computed here is interpreted as the total stress (rather than the purely elastic part of the stress).

Code/Source/solver/active_stress_uniform_steady.h
michelebucelli: Also, I removed the old `Fiber_reinforcement_stress` and replaced it with `UniformUnsteadyActiveStress`. The tests that previously used `Fiber_reinforcement_stress` still pass with the new implementation.

michelebucelli:
I think this PR is now ready for review and merging.

Code/Source/solver/active_stress.h
kko27: This sounds good to me! Agree with doing this for `IonicModel` at a later stage

Code/Source/solver/active_stress_ode.h
```diff
@@ -0,0 +1,155 @@
...
+   * @param[in] dt Time step size.
+   * @param[in] calcium Calcium concentration at the current node.
+   * @param[in] fiber_stretch Fiber stretch at the current node.
+   * @param[in] fiber_stretch_rate Fiber stretch rate at the current node.
+   *
+   * @return A vector containing the rate of change for each state variable.
+   */
+  virtual Vector<double> getf(const double t, const Vector<double> &state,
+                              const double calcium, const double fiber_stretch,
+                              const double fiber_stretch_rate) const = 0;
```
kko27, line 149: I just had a question regarding time `t` for future model implementation. For models with time dependence, would we need to pass the old time (i.e., `t - dt`) if we were using explicit FE time-stepping? This currently isn't an issue as Nash-Panfilov doesn't use the time `t` argument

Code/Source/solver/Vector.h
```diff
@@ -340,6 +342,25 @@ class Vector
...
 
+    /// @brief Increment by a rescaled vector.
+    ///
+    /// This is equivalent to <kbd>*this = *this + c * vec</kbd> but avoids the
+    /// temporary vector created by the operator+ and operator*.
+    Vector<T> &add(const double &c, const Vector<T> &vec)
+    {
+      if (size_ != vec.size()) {
+        svmp::raise<svmp::FE::InvalidArgumentException>(
+            "Vectors have diffrenct sizes: " + std::to_string(size_) +
```
kko27, line 354: minor typo here for error message

Code/Source/solver/Integrator.cpp
```diff
@@ -457,6 +460,77 @@ void Integrator::predictor()
   dmsg << "dFlag: " << com_mod.dFlag;
   #endif
 
+  Vector<double> fiber_stretch;
+  Vector<double> fiber_stretch_rate;
+
+  // Determine if we need to compute fiber stretch and stretch rate, by going
+  // through all domains of all equations until we find one for which active
+  // stress is enabled.
+  // @todo[michelebucelli] This should be extended to also compute fiber stretch
```
kko27: I think this comment has already been addressed (in line 488)

Code/Source/solver/Integrator.cpp
```diff
@@ -457,6 +460,77 @@ void Integrator::predictor()
...
+  // the stretch for each mesh, and then copy the mesh-local resulting vector
+  // into the global vector.
+  if (need_fiber_stretch) {
+    fiber_stretch.resize(com_mod.tnNo);
+
+    if (fiber_stretch_eq_index >= 0) {
+      for (const auto &mesh : com_mod.msh) {
+        Vector<double> tmp(mesh.nNo);
+
+        post::fib_stretch(com_mod, fiber_stretch_eq_index, mesh, Dn, tmp);
```
kko27, line 501: Probably a separate issue in itself, but do you have thoughts on whether `fib_stretch` and `fib_stretch_rate` still belongs to `post`?

kko27 (commented):
@michelebucelli, thanks so much for working on this! Documentation was excellent and the code was very easy to follow. 

I had a general comment about coding style. I noticed that in `Exception.h` and `Parameters.cpp`, many of the changes are due to indentation, whitespaces, and the position of the `*` and `&` operators. I have two questions about this. 1) Should we think about enforcing coding standards (as discussed in #96)? 2) And (more practically for me) is there a way I can review the code without looking over these minor changes?

ktbolt:
@kko27 There used to an implementation document describing implementation details and a `Coding Standards` section; seems that that was removed.

It is standard procedure in software engineering to follow the coding style used in a project; especially that used in a file. However, most developers here don't follow this practice so new code ends up looking different from one function to the next (Parameters.cpp) or the coding in an entire file does not look like that used in any other file in the project (Core/Exception.h).

I suppose we will again enable automatically enforced coding standards. Note that this may mess up the formatting of some complex mathematical statements like those used in material models.

Code/Source/solver/active_stress_ode.h
michelebucelli: Good question and good catch! I haven't thought about it too much.

I think that `getf` shouldn't worry about that (it is the code counterpart of the right-hand side function $f(t, y)$ ). If we need to evaluate the function at other time points (e.g. for forward Euler or RK4), the caller (i.e. `ActiveStressODE::advance_time_step_local) should call this function passing in different values of the time and of the state vector.

I double checked `ActiveStressODE::advance_time_step_local` to figure out whether we should call `getf` with `t` or `t - dt`. It turns out that the value of `t` passed to `advance_time_step_local` is the **new** time, while we want it to be the old. So I switched it to `t - dt`.

Code/Source/solver/Integrator.cpp
michelebucelli: Not much, but: my feeling is that in the long run the physical models could/should be encapsulated in their own class, and the computation of fiber stretch and stretch rate would make sense as a method of the structural solver class. I think that's a much wider refactoring task that this PR warrants though, so I'd leave them in `post` for now.

michelebucelli:
@kko27 @ktbolt Sorry, I have automatic formatting enabled in VS Code. It normally only touches lines of code that I've modified, but sometimes it misfires and touches the whole file without me noticing. A nice (terminal-only) tool to filter out whitespace changes in diffs is [difftastic](https://github.com/wilfred/difftastic).

Having said that, I am very much in favor of automated formatting. I've used `clang-format` in the past, and I think it goes a long way to making sure the code is clean, consistent and readable. I agree with your concern on long mathematical statements, but it can usually be worked around by breaking the statement into multiple smaller ones at no meaningful performance cost.

ktbolt:
@michelebucelli We used clang-format before so we will enable that. Thanks for your feedback !

Code/Source/solver/Integrator.cpp
```diff
@@ -481,7 +553,57 @@ void Integrator::predictor()
 
     // electrophysiology
     if (eq.phys == Equation_CEP) {
-      cep_ion::cep_integ(simulation_, iEq, e, solutions_);
+      if (fiber_stretch.size() != com_mod.tnNo)
+        svmp::raise<svmp::InternalErrorException>(
+            "Fiber stretch vector is not initialized correctly.");
```
javijv4, line 558: Do we need this? fiber_stretch is initialized with size `tnNo` a couple of lines above.

Code/Source/solver/Integrator.cpp
```diff
@@ -481,7 +553,57 @@ void Integrator::predictor()
...
+          // allow for active stress (e.g. struct) for which active stress is
+          // not enabled contribute a zero value to the average.
+          if (!supports_active_stress(dmn.phys))
+            continue;
+
+          if (dmn.active_stress != nullptr) {
+            Ta_f += dmn.active_stress->get_tension_fibers(Ac);
+            Ta_s += dmn.active_stress->get_tension_sheets(Ac);
+            Ta_n += dmn.active_stress->get_tension_sheet_normals(Ac);
+          }
```
javijv4, line 602: Shouldn't we also check whether the node `Ac` belongs to the domain?

tests/test_electromechanics.py
```diff
@@ -0,0 +1,29 @@
...
+
+# Fields to test
+fields = [
+    "Membrane_potential",
+    "Calcium",
+    "Cauchy_stress",
+    "Def_grad",
+    "Displacement",
+    "Jacobian",
+    "Jacobian",
```
javijv4: Duplicated `Jacobian`

tests/cases/electromechanics/slab/solver.xml
```diff
@@ -0,0 +1,191 @@
...
+      <NashPanfilov>
+        <ODE_solver> FE </ODE_solver>
+
+        <epsilon_0> 0.1 </epsilon_0> <!-- 1/ms -->
+        <epsilon_i> 1.0 </epsilon_i> <!-- 1/ms -->
+        <xi_T> 4.0e3 </xi_T>
+        <eta_T> 1e2 </eta_T> <!-- MPa/mM -->
+        <calcium_rest> 1.25e-4 </calcium_rest>
+        <calcium_crit> 8e-4 </calcium_crit>
+      </NashPanfilov>
```
javijv4, line 153: I noticed that if I remove this block, the code still runs (probably using the defaults in the code?).  I think it would be better if there were a check that ensured that, if `<Model>NashPanfilov</Model>` is present, then a `<NashPanfilov>` block should be present.

tests/cases/electromechanics/slab/solver.xml
```diff
@@ -0,0 +1,191 @@
...
+    <Domain_file_path> ./mesh/volume.vtu </Domain_file_path>
+
+    <Fiber_direction> (1, 0, 0)  </Fiber_direction>
+    <Fiber_direction> (0, 1, 0)  </Fiber_direction>
+    <Fiber_direction> (0, 1, 0)  </Fiber_direction>
+
+    <Quadrature_modifier_TET4> 0.34 </Quadrature_modifier_TET4>
+  </Add_mesh>
+
+  <Add_equation type="CEP">
```
javijv4, line 42: I was trying out some things with the test and noticed two things when I swapped the order of CEP and struct. 
1. The results are very similar but differ at the third decimal (for the displacement at timestep 100). I think we talked about this equation ordering issue at some point, and if these results are expected, then maybe it would be good to explain it in the documentation. 

2. The output for CEP + struct looks like
``` 
---------------------------------------------------------------------
 EP 100-1  9.100e+01  [0 1.000e+00 1.657e-02 9.622e-11]  [22 -231 1]
 ST 100-1  9.114e+01  [0 1.000e+00 2.751e+01 9.197e-12]  [158 -254 54]
 ST 100-2  9.140e+01  [-87 4.443e-05 1.222e-03 2.613e-07]  [234 -152 63]
 ST 100-3  9.147e+01  [-208 3.833e-11 1.055e-09 2.645e-01]  [3 -13 0]
 ST 100-4  9.153e+01  [-212 2.397e-11 6.594e-10 4.878e-01]  [2 -7 2]
 ST 100-5  9.160e+01  [-214 1.806e-11 4.968e-10 5.845e-01]  [2 -5 0]
---------------------------------------------------------------------
 EP 100-1s 9.169e+01  [0 1.000e+00 1.657e-02 9.622e-11]  [22 -231 1]
```
But when I swap the order (struct + CEP), I get
```
---------------------------------------------------------------------
 ST 100-1  8.753e+01  [0 1.000e+00 2.081e-01 9.947e-12]  [161 -253 31]
 ST 100-2  8.777e+01  [-86 4.700e-05 9.781e-06 1.126e-07]  [253 -9 73]
 ST 100-3  8.783e+01  [-208 3.726e-11 7.753e-12 2.035e-01]  [3 -16 1]
 ST 100-4  8.790e+01  [-212 2.284e-11 4.753e-12 4.039e-01]  [2 -9 0]
 ST 100-5  8.796e+01  [-215 1.602e-11 3.334e-12 4.965e-01]  [2 -7 0]
 ST 100-6  8.803e+01  [-218 1.241e-11 2.584e-12 5.723e-01]  [2 -6 0]
 ST 100-6s 8.805e+01  [-218 1.241e-11 2.584e-12 5.723e-01]  [2 -6 0]
```
With no EP residual shown.

javijv4 (commented):
Thanks for working on this @michelebucelli. Looks great. I just left a few comments and questions.

tests/cases/electromechanics/slab/solver.xml
michelebucelli: Thanks for trying this out!

1. I think that differences should be expected if solving the equations in a different order, although the different orders should converge to the same solution as $\Delta t \to 0$. In principle, I don't think that either ordering is more correct than the other. I tend to think in terms of EP -> struct because of "logical causality". I'd be happy to document that results are sensitive to the order of equations. I'm not sure what would be the right place for documenting it though.

2. I looked into this a while ago and, if I remember correctly, there is some logic (perhaps unintended) by which the last line of the last equation being solved is not printed. The way this works is:
    - `Integrator::step` does not print the line in which all equations converge (in your first case, that is structure iteration 6, not shown; in the second case, it is EP iteration 1, also not shown);
    - after convergence, `main` prints the last line for the first equation, adding the `s` marker if it's saving data to file;
    - I'm not sure about what is happening with the separator line in the first case (EP then struct).

    I think this behavior is a bit confusing (especially the fact that it skips some lines), and its implementation is more complicated than necessary. Maybe it's worth addressing in a separate PR?

tests/cases/electromechanics/slab/solver.xml
michelebucelli: I inherited this behavior from that of `IonicModel` (see [this comment thread](https://github.com/SimVascular/svMultiPhysics/pull/540#discussion_r3228143736)), where the section for the selected model can be removed (in which case default parameters are used), but if it is included all the model parameters must be specified.

I wouldn't mind switching to requiring the section for the selected model, but I'd like to maintain consistency with the `IonicModel` behavior (meaning that I'd change the behavior of `IonicModel` too).

Code/Source/solver/Integrator.cpp
michelebucelli: Definitely! Good catch, I'll fix this.

Code/Source/solver/Integrator.cpp
michelebucelli: I think I implemented this check before changing how `fiber_stretch` is initialized, so it is probably redundant now. Still, since the initialization logic is not trivial, I think it might be good to keep the check to safeguard against accidentally breaking this.

(This would be an ideal application for a debug-mode-only check, but I don't think debug mode is currently very used, so I'd suggest keeping it as it is for now and maybe revisiting later).

tests/cases/electromechanics/slab/solver.xml
michelebucelli: Separate but related issue: it seems that the ordering of equations messes things up when coupling to the 0D solver (something related to the various `Coupled` flags, that I understand only in part). I propose addressing that in the next PR though (coupling to the 0D solver is the next thing I'll have to work on in any case).

(Edit for future reference: this is discussed further in #592)

tests/cases/electromechanics/slab/solver.xml
javijv4: OK, yes, I think I am now convinced we should require the model parameter section for clarity (in both ionic and active stress models).  What do you think @kko27?

I am okay with leaving that outside this PR, though, and fixing it in a different issue. Whatever you think is more convenient.

kko27: Since we need to change the behavior of `IonicModel` as well, I think this falls outside of the scope of this PR. Let's create a separate issue for this

Code/Source/solver/Parameters.h
michelebucelli: (Resolving, but this is further discussed in #591).

Code/Source/solver/active_stress.h
michelebucelli: I'll add this to the task list of #549.

tests/cases/electromechanics/slab/solver.xml
michelebucelli: I'll add this to the task list of #540.

javijv4 (approved):
Looks good to me!

kko27 (approved):
Looks good!

ktbolt:
@michelebucelli Are there any unit tests for `ActiveStress` ?

ktbolt:
@michelebucelli And we name class files using the class name, like `ActiveStress.h`.

Sorry, I missed this until now.

michelebucelli:
@ktbolt 

> Are there any unit tests for `ActiveStress`?

No (or at least not yet). There are integration tests (one new electromechanics test using `NashPanfilov`, and the old active mechanics tests replacing `Fiber_reinforcement_stress` with `UniformUnsteadyActiveStress`).

> And we name class files using the class name, like ActiveStress.h.

Sorry, I forgot to update this. If it's all right with you, I'll sneak this into the PR for #592. I'll also do the same for `IonicModel`.

ktbolt:
@michelebucelli No problem; sneak it into https://github.com/SimVascular/svMultiPhysics/issues/592. 

We will need to start thinking about adding unit tests when new classes are implemented.

Issue #572, opened 2026-06-23
Spatial bounds on electrophysiology applied stimulus


michelebucelli:
### Problem

Currently, when a stimulus current is applied, it is applied to the whole domain that contains it. This implies that changing the position, size or shape of the stimulus requires regenerating the domain labels, which might be cumbersome if this needs to be done repeatedly.

### Solution

Add sub-tags to the `<Stimulus>` XML tag to constrain it within a certain subregion of the domain. Some examples might be:
- axis-aligned bounding box (only apply the stimulus if the point coordinates fall within user-defined minimum and maximum values);
- spherical stimulus (only apply the stimulus if the point falls within a sphere of user-defined center and radius).

This way the stimulus placement, size and shape will be changed by editing the XML file only. Adding other shapes as needed should be rather straightforward.

### Additional context

- The choice of adding the spatial bounds within the domain is driven mostly by backwards compatibility. In principle, the stimulus could be entirely domain-independent, but I guess this might break some existing parameter files.
- It would also be nice to have this new structure support multiple stimuli in the same domain.
- I've asked Sami Bismar (@samibismar) to work on this as a first issue in `svMultiPhysics`. Sami is collaborating with our group over the summer.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct and Contributing Guidelines

samibismar:
I pushed the current implementation to this branch:

https://github.com/samibismar/svMultiPhysics/tree/feature-cep-spatial-stimulus

The current branch adds optional spatial bounds inside the existing `<Stimulus>` XML structure while keeping the stimulus restricted to its parent domain for backwards compatibility.

Currently supported tags are:

```xml
<Box_min> x_min y_min z_min </Box_min>
<Box_max> x_max y_max z_max </Box_max>
<Sphere_center> cx cy cz </Sphere_center>
<Sphere_radius> r </Sphere_radius>
```

If no spatial bounds are provided, the behavior remains unchanged. If a box and/or sphere is provided, the stimulus is applied only to nodes inside the parent domain that also satisfy the specified geometric restriction. If both box and sphere bounds are provided, they are combined by intersection.

The screenshots below show a spherical stimulus localized inside a larger Domain 1 region in the Niederer benchmark case using only XML bounds, without changing the mesh or relabeling domains. The first image shows the initial bounded stimulus region, and the second shows subsequent propagation from that localized region.

**Initial localized spherical stimulus inside Domain 1**

[...]

PR #579, opened 2026-06-29
Add geometric filtering for CEP stimuli

samibismar:
## Current situation

Related to #572.

This PR adds optional geometric filtering for CEP applied stimuli. A CEP stimulus can now be restricted using an axis-aligned box, a sphere, or both together.

When both box and sphere bounds are provided, the stimulus is applied only to nodes that satisfy both geometric restrictions, while still respecting the parent domain.

This PR also incorporates review feedback by grouping the new XML parameters under `Spatial_bounds` and refactoring the stimulus-specific logic into `stimType`.

## Release Notes

* Added optional spatial bounds for CEP stimuli:

  * `Spatial_bounds / Box / Minimum`
  * `Spatial_bounds / Box / Maximum`
  * `Spatial_bounds / Sphere / Center`
  * `Spatial_bounds / Sphere / Radius`
* Box and sphere bounds can be used independently or together.
* When box and sphere bounds are both provided, the applied stimulus region is their intersection.
* Refactored CEP stimulus evaluation so temporal activation, spatial filtering, and MPI distribution are handled by `stimType`.
* Replaced feature-related error paths with the svMultiPhysics exception infrastructure.

Example:

```xml
<Stimulus type="Istim" >
  <Amplitude> -35.714 </Amplitude>
  <Start_time> 0.0 </Start_time>
  <Duration> 2.0 </Duration>
  <Cycle_length> 10000.0 </Cycle_length>

  <Spatial_bounds>
    <Box>
      <Minimum> 0.20 0.20 0.20 </Minimum>
      <Maximum> 1.00 1.00 1.00 </Maximum>
    </Box>

    <Sphere>
      <Center> 0.55 0.75 0.75 </Center>
      <Radius> 0.50 </Radius>
    </Sphere>
  </Spatial_bounds>
</Stimulus>
```

## Documentation

The new XML structure is documented in the `StimulusParameters` example in `Parameters.h`.

## Testing

Successfully compiled on macOS.

Validated using a Niederer benchmark CEP case with several stimulus configurations:

* no stimulus
* no spatial bounds
* box-only bounds
* sphere-only bounds
* overlapping box+sphere bounds
* disjoint box+sphere bounds
* impossible box/sphere bounds

The expected logical comparisons passed in serial and with MPI using 2 processes. In particular, the disjoint box+sphere and impossible-bound cases matched the no-stimulus case, while valid bounded cases differed from the unbounded stimulus case.

Also ran a visual validation using a spherical stimulus with 4 MPI processes and confirmed that the activation was localized and propagated as expected.

### DIFF
--- a/Code/Source/solver/CepMod.cpp
+++ b/Code/Source/solver/CepMod.cpp
@@ -134,15 +134,9 @@ void stimType::SpatialBounds::distribute(const CmMod& cm_mod, const cmType& cm)
 
 double stimType::operator()(const double time, const Vector<double>& x) const
 {
-  if (utils::is_zero(amplitude)) {
-    return 0.0;
-  }
-
-  if (!is_active(time)) {
-    return 0.0;
-  }
-
-  if (!spatial_bounds.contains(x)) {
+  if (utils::is_zero(amplitude) ||
+      !is_active(time) ||
+      !spatial_bounds.contains(x)) {
     return 0.0;
   }
 
--- a/Code/Source/solver/CepMod.cpp
+++ b/Code/Source/solver/CepMod.cpp
@@ -35,9 +35,9 @@ bool stimType::is_active(const double time) const
 {
   const double eps = std::numeric_limits<double>::epsilon();
 
-  const int icl = static_cast<int>(fmax(floor(time / CL), 0.0));
-  const double Ts_cycle = Ts + static_cast<double>(icl) * CL;
-  const double Te_cycle = Ts_cycle + Td;
+  const int icl = static_cast<int>(fmax(floor(time / cycle_length_), 0.0));
+  const double Ts_cycle = start_time_ + static_cast<double>(icl) * cycle_length_;
+  const double Te_cycle = Ts_cycle + duration_;
 
   return Ts_cycle - eps <= time && time <= Te_cycle + eps;
 }
@@ -144,7 +144,7 @@ void SpatialBounds::distribute(const CmMod& cm_mod, const cmType& cm)
 
 double stimType::operator()(const double time, const Vector<double>& x) const
 {
-  if (utils::is_zero(A)) {
+  if (utils::is_zero(amplitude_)) {
     return 0.0;
   }
 
@@ -156,26 +156,26 @@ double stimType::operator()(const double time, const Vector<double>& x) const
     return 0.0;
   }
 
-  return A;
+  return amplitude_;
 }
 
 void stimType::distribute(const CmMod& cm_mod, const cmType& cm)
 {
-  cm.bcast(cm_mod, &Ts);
-  cm.bcast(cm_mod, &Td);
-  cm.bcast(cm_mod, &CL);
-  cm.bcast(cm_mod, &A);
+  cm.bcast(cm_mod, &start_time_);
+  cm.bcast(cm_mod, &duration_);
+  cm.bcast(cm_mod, &cycle_length_);
+  cm.bcast(cm_mod, &amplitude_);
   spatial_bounds_.distribute(cm_mod, cm);
 }
 
 void stimType::read_parameters(const StimulusParameters& params, const int nsd, const double default_cycle_length)
 {
-  A = params.amplitude.value();
+  amplitude_ = params.amplitude.value();
 
-  if (!utils::is_zero(A)) {
-    Ts = params.start_time.value();
-    Td = params.duration.value();
-    CL = params.cycle_length.defined() ? params.cycle_length.value() : default_cycle_length;
+  if (!utils::is_zero(amplitude_)) {
+    start_time_ = params.start_time.value();
+    duration_ = params.duration.value();
+    cycle_length_ = params.cycle_length.defined() ? params.cycle_length.value() : default_cycle_length;
   }
 
   const auto& spatial_bounds_params = params.spatial_bounds;
@@ -199,9 +199,9 @@ void stimType::read_parameters(const StimulusParameters& params, const int nsd, 
           SVMP_HERE, "Stimulus box Minimum and Maximum must have the same coordinate dimension.");
     }
 
-    if (box_min_vals.size() < static_cast<std::size_t>(nsd)) {
+    if (box_min_vals.size() != static_cast<std::size_t>(nsd)) {
       svmp::raise<svmp::ParseException>(
-          SVMP_HERE, "Stimulus box dimension is smaller than the simulation spatial dimension.");
+          SVMP_HERE, "Stimulus box coordinate dimension must match the simulation spatial dimension.");
     }
 
     Vector<double> box_min(box_min_vals.size());
@@ -234,9 +234,9 @@ void stimType::read_parameters(const StimulusParameters& params, const int nsd, 
     const auto& sphere_center_vals = sphere_params.center.value();
     const double sphere_radius_val = sphere_params.radius.value();
 
-    if (sphere_center_vals.size() < static_cast<std::size_t>(nsd)) {
+    if (sphere_center_vals.size() != static_cast<std::size_t>(nsd)) {
       svmp::raise<svmp::ParseException>(
-          SVMP_HERE, "Stimulus sphere center dimension is smaller than the simulation spatial dimension.");
+          SVMP_HERE, "Stimulus sphere center coordinate dimension must match the simulation spatial dimension.");
     }
 
     if (sphere_radius_val < 0.0) {
--- a/Code/Source/solver/CepMod.cpp
+++ b/Code/Source/solver/CepMod.cpp
@@ -35,36 +35,41 @@ bool stimType::is_active(const double time) const
 {
   const double eps = std::numeric_limits<double>::epsilon();
 
-  const int icl = static_cast<int>(fmax(floor(time / cycle_length_), 0.0));
-  const double Ts_cycle = start_time_ + static_cast<double>(icl) * cycle_length_;
-  const double Te_cycle = Ts_cycle + duration_;
-
-  return Ts_cycle - eps <= time && time <= Te_cycle + eps;
+  const double time_relative = std::fmod(time, cycle_length);
+  return start_time - eps <= time_relative && time_relative <= start_time + duration + eps;
 }
 
-void SpatialBounds::set_box(const Vector<double>& min, const Vector<double>& max)
+void stimType::SpatialBounds::set_box(const Vector<double>& min, const Vector<double>& max)
 {
-  box_min_ = min;
-  box_max_ = max;
-  region_type_ = (region_type_ == RegionType::sphere) ? RegionType::both : RegionType::box;
+  svmp::throw_if<svmp::FE::InvalidArgumentException>(
+      min.size() == 0, "Stimulus box bounds must have at least one coordinate.");
+  svmp::throw_if<svmp::FE::InvalidArgumentException>(
+      min.size() != max.size(), "Stimulus box minimum and maximum must have the same coordinate dimension.");
+  box_min = min;
+  box_max = max;
+  has_box = true;
 }
 
-void SpatialBounds::set_sphere(const Vector<double>& center, const double radius)
+void stimType::SpatialBounds::set_sphere(const Vector<double>& center, const double radius)
 {
-  sphere_center_ = center;
-  sphere_radius_ = radius;
-  region_type_ = (region_type_ == RegionType::box) ? RegionType::both : RegionType::sphere;
+  svmp::throw_if<svmp::FE::InvalidArgumentException>(
+      center.size() == 0, "Stimulus sphere center must have at least one coordinate.");
+  svmp::throw_if<svmp::FE::InvalidArgumentException>(
+      radius < 0.0, "Stimulus sphere radius must be non-negative.");
+  sphere_center = center;
+  sphere_radius = radius;
+  has_sphere = true;
 }
 
-bool SpatialBounds::inside_box(const Vector<double>& x) const
+bool stimType::SpatialBounds::inside_box(const Vector<double>& x) const
 {
-  if (x.size() != box_min_.size()) {
+  if (x.size() != box_min.size()) {
     svmp::raise<svmp::FE::InvalidArgumentException>(
-        SVMP_HERE, "Point dimension does not match stimulus box dimension.");
+        "Point dimension does not match stimulus box dimension.");
   }
 
   for (int i = 0; i < x.size(); i++) {
-    if (x(i) < box_min_(i) || x(i) > box_max_(i)) {
+    if (x[i] < box_min[i] || x[i] > box_max[i]) {
       return false;
     }
   }
@@ -72,79 +77,64 @@ bool SpatialBounds::inside_box(const Vector<double>& x) const
   return true;
 }
 
-bool SpatialBounds::inside_sphere(const Vector<double>& x) const
+bool stimType::SpatialBounds::inside_sphere(const Vector<double>& x) const
 {
-  if (x.size() != sphere_center_.size()) {
+  if (x.size() != sphere_center.size()) {
     svmp::raise<svmp::FE::InvalidArgumentException>(
-        SVMP_HERE, "Point dimension does not match stimulus sphere dimension.");
+        "Point dimension does not match stimulus sphere dimension.");
   }
 
   double distance_squared = 0.0;
 
   for (int i = 0; i < x.size(); i++) {
-    const double dx = x(i) - sphere_center_(i);
+    const double dx = x[i] - sphere_center[i];
     distance_squared += dx * dx;
   }
 
-  return distance_squared <= sphere_radius_ * sphere_radius_;
+  return distance_squared <= sphere_radius * sphere_radius;
 }
 
-bool SpatialBounds::contains(const Vector<double>& x) const
+bool stimType::SpatialBounds::contains(const Vector<double>& x) const
 {
-  switch (region_type_) {
-    case RegionType::none:   return true;
-    case RegionType::box:    return inside_box(x);
-    case RegionType::sphere: return inside_sphere(x);
-    case RegionType::both:   return inside_box(x) && inside_sphere(x);
-  }
+  if (has_box && !inside_box(x)) return false;
+  if (has_sphere && !inside_sphere(x)) return false;
   return true;
 }
 
-void SpatialBounds::distribute(const CmMod& cm_mod, const cmType& cm)
+void stimType::SpatialBounds::distribute(const CmMod& cm_mod, const cmType& cm)
 {
-  int region_type_int = static_cast<int>(region_type_);
-  cm.bcast(cm_mod, &region_type_int);
-  region_type_ = static_cast<RegionType>(region_type_int);
+  cm.bcast(cm_mod, &has_box);
+  cm.bcast(cm_mod, &has_sphere);
 
-  if (region_type_ == RegionType::box || region_type_ == RegionType::both) {
-    int box_size = box_min_.size();
+  if (has_box) {
+    int box_size = box_min.size();
     cm.bcast(cm_mod, &box_size);
 
-    if (box_size <= 0) {
-      svmp::raise<svmp::FE::InvalidArgumentException>(
-          SVMP_HERE, "Stimulus box has invalid coordinate dimension.");
-    }
-
     if (cm.slv(cm_mod)) {
-      box_min_.resize(box_size);
-      box_max_.resize(box_size);
+      box_min.resize(box_size);
+      box_max.resize(box_size);
     }
 
-    cm.bcast(cm_mod, box_min_, "SpatialBounds box_min");
-    cm.bcast(cm_mod, box_max_, "SpatialBounds box_max");
+    cm.bcast(cm_mod, box_min, "SpatialBounds box_min");
+    cm.bcast(cm_mod, box_max, "SpatialBounds box_max");
   }
 
-  if (region_type_ == RegionType::sphere || region_type_ == RegionType::both) {
-    int sphere_center_size = sphere_center_.size();
+  if (has_sphere) {
+    int sphere_center_size = sphere_center.size();
     cm.bcast(cm_mod, &sphere_center_size);
 
-    if (sphere_center_size <= 0) {
-      svmp::raise<svmp::FE::InvalidArgumentException>(
-          SVMP_HERE, "Stimulus sphere center has invalid coordinate dimension.");
-    }
-
     if (cm.slv(cm_mod)) {
-      sphere_center_.resize(sphere_center_size);
+      sphere_center.resize(sphere_center_size);
     }
 
-    cm.bcast(cm_mod, sphere_center_, "SpatialBounds sphere_center");
-    cm.bcast(cm_mod, &sphere_radius_);
+    cm.bcast(cm_mod, sphere_center, "SpatialBounds sphere_center");
+    cm.bcast(cm_mod, &sphere_radius);
   }
 }
 
 double stimType::operator()(const double time, const Vector<double>& x) const
 {
-  if (utils::is_zero(amplitude_)) {
+  if (utils::is_zero(amplitude)) {
     return 0.0;
   }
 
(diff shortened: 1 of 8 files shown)


Code/Source/solver/CepMod.cpp
```diff
@@ -24,6 +29,124 @@ const std::map<std::string,ElectrophysiologyModelType> cep_model_name_to_type
...
+  for (int i = 0; i < com_mod.nsd; i++) {
+    const double xi = com_mod.x(i, Ac);
+
+    if (xi < box_min(i) || xi > box_max(i)) {
+      return false;
+    }
+  }
+
+  return true;
+}
```
michelebucelli: I suggest changing this function's signature to
```cpp
bool stimType::inside_box(const Vector<double>& x) const
```
with the caller responsible for extracting the point from `com_mod` (or from elsewhere). Logically, this function is a geometrical check on a point, so it makes sense that its input is a point.

This suggestion propagates up through the call stack - I think at least up to `cep_integ_l`. That is to say, also `cep_integ_l` should take in input a point, instead of `com_mod` and `Ac`, and it should be the caller of `cep_integ_l` that extracts the point from `com_mod.x`.

Similar suggestions apply to `inside_sphere`.

Code/Source/solver/CepMod.cpp
```diff
@@ -24,6 +29,124 @@ const std::map<std::string,ElectrophysiologyModelType> cep_model_name_to_type
...
+
+  return Ts_cycle - eps <= time && time <= Te_cycle + eps;
+}
+
+bool stimType::inside_box(const ComMod& com_mod, const int Ac) const
+{
+  if (box_min.size() < com_mod.nsd || box_max.size() < com_mod.nsd) {
+    svmp::raise<svmp::FE::InvalidArgumentException>(
+        SVMP_HERE, "Stimulus box dimension is smaller than the simulation spatial dimension.");
+  }
```
michelebucelli: The result of this check will not change over the course of the simulation, but will be fixed as soon as the parameters for the box are read. Accordingly, I suggest to make it right after parsing those parameters.

**EDIT**: reading further, I noticed that those checks are already in place in `read_files.cpp`. Then, the check here is somewhat redundant. What you might still want to do, after the change of signature proposed in my other comment, is to check that the size of the input point is correct (as in, matches the size of the bounds).

Code/Source/solver/read_files.cpp
```diff
@@ -1085,6 +1085,9 @@ void read_cep_domain(Simulation* simulation, EquationParameters* eq_params, Doma
   lDmn.cep.Istim.Ts = 99999.0;
   lDmn.cep.Istim.CL = 99999.0;
   lDmn.cep.Istim.Td = 0.0;
+  lDmn.cep.Istim.box_defined = false;
+  lDmn.cep.Istim.sphere_defined = false;
+  lDmn.cep.Istim.sphere_radius = 0.0;
```
michelebucelli: These seem unnecessary: the `Istim` object will (and should) already have its values initialized by whatever its (default) constructor does (in this case it will assign whatever default value is defined in the declaration of each of these members).

Code/Source/solver/CepMod.h
```diff
@@ -59,6 +63,43 @@ class stimType
...
+    Vector<double> box_max;
+
+    /// @brief Whether the stimulus is restricted to a sphere.
+    bool sphere_defined = false;
+
+    /// @brief Center coordinates for the stimulus sphere.
+    Vector<double> sphere_center;
+
+    /// @brief Radius of the stimulus sphere.
+    double sphere_radius = 0.0;
```
michelebucelli: In the interest of further decoupling the stimulus class from the rest of the code, I think it'd be better if these data members were `private`.

As far as I can tell, the only place where external access is still needed is when reading the parameters from the XML file. If this is true, those instructions can also be moved to a method `stimType::read_parameters(const StimulusParameters &params)`.

Code/Source/solver/Parameters.cpp
```diff
@@ -2126,6 +2126,84 @@ void FiberReinforcementStressParameters::print_parameters()
 // 'Stimulus' XML element used to parameters for 
 // pacemaker cells.
 
+/// @brief Define the XML element name for CEP stimulus box spatial bounds.
```
michelebucelli: I think it's better if Doxygen documentation comments are placed in the header file (`Parameters.h`).

In my opinion, if a developer wants to look some information up, the "natural" order would be first the compiled documentation (i.e. the Doxygen-generated website), then maybe the header, and only as a last resort the `cpp` source file. Because of this, it seems more natural to me to place documentation comments in the header.

Code/Source/solver/Parameters.cpp
```diff
@@ -2126,6 +2126,84 @@ void FiberReinforcementStressParameters::print_parameters()
...
+    if (name == StimulusBoxParameters::xml_element_name_) {
+      box.set_values(item);
+    } else if (name == StimulusSphereParameters::xml_element_name_) {
+      sphere.set_values(item);
+    } else {
+      svmp::raise<svmp::ParseException>(SVMP_HERE, error_msg + name + "'.");
+    }
+
+    item = item->NextSiblingElement();
+  }
```
michelebucelli: I think that using a `for` loop for this is more semantically clear, as it clearly communicates a fixed number of iterations and/or iteration through a container:
```cpp
for (auto item = xml_elem->FirstChildElement();
     item != nullptr;
     item = item->NextSiblingElement()) {
  // ...do stuff...
}
```

michelebucelli (commented):
Thank you @samibismar! I left a few suggestions.

Code/Source/solver/read_files.cpp
ktbolt: @samibismar It will probably be good to create a class to group the spatial parameters and an enum for the kind of spatial region: `box, sphere, none`. Then you would just need to check the type is not `none`.

The `bcast` would be a bit more complicated but that would be a method in the class.

Code/Source/solver/CepMod.h
ktbolt: @michelebucelli It is good if classes do not know about `Parameters`; maybe a helper to set the data.

michelebucelli: Could you elaborate a bit on why you think this would be better?

In my understanding, the parameter schema (how many, what names, what types, what grouping structure) is closely related to the internal data representation of the class, so that parameters and classes consuming them are already tightly coupled. For this reason, placing the parameter interpretation in the same place as the class seems like a good idea to me.

ktbolt: @michelebucelli It is good to maintain loose coupling between modules; if `stimType` knows about `StimulusParameters` then any changes (e.g., variable names or class structure) to that requires changes to `stimType`. 

If we later set solver parameters using JSON does `stimType` then need to know about JSON ?

The current C++ code follows the Fortran code in which the parameter processing is mixed in all over the place setting object data; would be good in future to isolate setting object data within `Parameter` classes.

Having gone on and on about this I however don't think it is totally hideous to pass `StimulusParameters` to a `stimType` method; need to clean all of this parameter processing thing later on anyway.

michelebucelli: I see, thanks for clarifying.

I still think that there is some inherent and unavoidable coupling. For example, if we change something in `stimType` (say, we add a new type of shape), we will also need to change `StimulusParameters` accordingly. That's why I think that the direction should be to bring the parameter schema for a class "closer" to that class (and "further" from all other classes, as we go on with this decoupling work).

In other words: I agree that modules should be coupled loosely, but I think that the parameters for a module should be managed by and inside of that module.

I agree with your assessment of the current state of things though, and that for what I'm describing to work there would need to be some larger refactoring of the way parameters are managed (for example, separating the abstract schema from the concrete problem of the file format used to represent it, e.g. JSON vs. XML).

If/when the cleanup of parameters processing comes to pass, I'd be happy to resume this discussion.

Code/Source/solver/read_files.cpp
```diff
@@ -1098,6 +1101,80 @@ void read_cep_domain(Simulation* simulation, EquationParameters* eq_params, Doma
...
+
+      if (box_min.size() != box_max.size()) {
+        svmp::raise<svmp::ParseException>(
+            SVMP_HERE, "Stimulus box Minimum and Maximum must have the same coordinate dimension.");
+      }
+
+      if (box_min.size() < static_cast<std::size_t>(simulation->com_mod.nsd)) {
+        svmp::raise<svmp::ParseException>(
+            SVMP_HERE, "Stimulus box dimension is smaller than the simulation spatial dimension.");
+      }
```
javijv4: This only checks that the box dimensions are not less than the simulation dimensions. Can we instead check that the box dimensions exactly match the simulation's dimensions? I know the code will only use the dimensions of the box up to `nsd` but the code should raise an issue if the user specified the wrong number of coordinates to avoid accidental typos becoming an issue.

Code/Source/solver/read_files.cpp
```diff
@@ -1098,6 +1101,80 @@ void read_cep_domain(Simulation* simulation, EquationParameters* eq_params, Doma
...
+    }
+
+    if (sphere_defined) {
+      const auto& sphere_center = sphere_params.center.value();
+      const double sphere_radius = sphere_params.radius.value();
+
+      if (sphere_center.size() < static_cast<std::size_t>(simulation->com_mod.nsd)) {
+        svmp::raise<svmp::ParseException>(
+            SVMP_HERE, "Stimulus sphere center dimension is smaller than the simulation spatial dimension.");
+      }
```
javijv4: same as above

Code/Source/solver/CepMod.h
```diff
@@ -59,6 +63,43 @@ class stimType
 
     /// @brief stimulus amplitude
     double A = 0.0;
```
javijv4: Can we use this PR to update these names to more representative ones as well? For example,
```
Ts -> start_time
Td -> stimulus_duration
CL -> cycle_length
A -> amplitude
```

Code/Source/solver/CepMod.cpp
samibismar: I moved the geometry checks into `SpatialBounds`, and `inside_box` / `inside_sphere` now take `const Vector<double>& x` rather than `ComMod` and `Ac`.

This change also propagates up to `cep_integ_l`: it now takes the point directly, and the caller extracts it with `com_mod.x.col(Ac)` before calling the local integration routine.

javijv4 (commented):
Hey @samibismar, thanks for working on this! It looks good. Just left a few small comments.

Code/Source/solver/CepMod.cpp
samibismar: I removed the old `com_mod.nsd`-based check from the geometry helper. The XML parsing/setup path now validates that the bounds coordinate dimension exactly matches the simulation spatial dimension when the parameters are read.

After changing the signature to take a point, I kept a local defensive check that the input point dimension matches the stored bounds dimension.

Code/Source/solver/CepMod.h
samibismar: I went ahead with `stimType::read_parameters(const StimulusParameters& params, ...)` as suggested. All stimulus setup is now centralized there, and the data members are no longer accessible externally.

Happy to revisit when the broader parameter-processing cleanup comes around.

Code/Source/solver/Parameters.cpp
samibismar: Changed to`for` loop.

Code/Source/solver/read_files.cpp
samibismar: I removed the direct `Istim` member initialization from `read_files.cpp`. `stimType` now owns its default values internally, and `read_files.cpp` only calls `read_parameters(...)` when a `<Stimulus>` block is defined.

I also added a dedicated `SpatialBounds` class with a `RegionType` enum (`none`, `box`, `sphere`, `both`) to group the box/sphere data and containment logic. The spatial-bound MPI distribution is now handled by `SpatialBounds::distribute(...)`.

Code/Source/solver/read_files.cpp
samibismar: Good point. I changed the box bounds validation to require the coordinate dimension to exactly match the simulation spatial dimension, rather than only checking that it is not smaller.

Code/Source/solver/CepMod.h
samibismar: Updated these to `start_time_`, `duration_`, `cycle_length_`, and `amplitude_`.

samibismar:
Thank you all for the review. Addressed the review feedback in the latest commits. The main structural changes are the new `SpatialBounds` class, making `stimType` internals private behind `read_parameters()`, and moving spatial-bound MPI distribution into `SpatialBounds`.

I replied on the individual threads with more detail. Please let me know if anything needs a different approach.

Code/Source/solver/CepMod.cpp
```diff
@@ -33,117 +35,222 @@ bool stimType::is_active(const double time) const
...
   const double eps = std::numeric_limits<double>::epsilon();
 
-  const int icl = static_cast<int>(fmax(floor(time / CL), 0.0));
-  const double Ts_cycle = Ts + static_cast<double>(icl) * CL;
-  const double Te_cycle = Ts_cycle + Td;
+  const int icl = static_cast<int>(fmax(floor(time / cycle_length_), 0.0));
+  const double Ts_cycle = start_time_ + static_cast<double>(icl) * cycle_length_;
+  const double Te_cycle = Ts_cycle + duration_;
 
   return Ts_cycle - eps <= time && time <= Te_cycle + eps;
```
michelebucelli: I think this should be equivalent to the following simpler code:
```cpp
const double time_relative = std::fmod(time, cycle_length_);
return start_time_ - eps <= time_relative &&
       time_relative <= start_time_ + duration_ + eps;
```

Code/Source/solver/read_files.cpp
michelebucelli: I think that perhaps the `enum class` is conflicting with the idea that multiple shapes can be prescribed, and the stimulus is applied to the intersection in that case. Now we only have two shapes, so `RegionType::both` is sufficient, but if we add new shapes the number of members of this `enum class` would increase combinatorially.

I think that the solution with the boolean flags is a more extensible solution if we want to keep the intersection thing. If instead we are willing to drop it, I think the `enum class` is the way to go.

Code/Source/solver/CepMod.h
```diff
@@ -48,58 +50,58 @@ class ComMod;
 class CmMod;
 class cmType;
 
-/// @brief External stimulus type
-class stimType
+/// @brief Spatial bounds for a CEP stimulus region.
+class SpatialBounds
```
michelebucelli: This class lives in the global scope, so its fully qualified name (i.e. the name with all the namespaces before it) does not clarify that these bounds are related to a stimulus.

I think the neatest solution to this would be to move this class inside `stimType`, so that the fully qualified name would be `stimType::SpatialBounds`, and there's no ambiguity there.

Code/Source/solver/CepMod.h
```diff
@@ -48,58 +50,58 @@ class ComMod;
...
   private:
-    /// @brief Check whether the stimulus is active at the given time.
-    bool is_active(const double time) const;
-
-    /// @brief Check whether a node lies inside all active spatial bounds.
-    bool contains_node(const ComMod& com_mod, const int Ac) const;
+    double start_time_ = 0.0;
+    double duration_ = 0.0;
+    double cycle_length_ = 0.0;
+    double amplitude_ = 0.0;
```
michelebucelli: Minor: I don't think we need to suffix the private members with `_`.

It would be useful if the members have a getter method for which one wants to use the name of the property itself (e.g. `double start_time() const { return start_time; }`), but that is not the case here.

Code/Source/solver/read_files.cpp
michelebucelli: (There's a more sophisticated solution in which `SpatialBounds` is abstract, and the concrete bounds are concrete classes derived from it, but I think that's way overkill unless there's a larger number of different shapes we want to allow)

Code/Source/solver/CepMod.cpp
```diff
@@ -33,117 +35,222 @@ bool stimType::is_active(const double time) const
...
 }
 
-bool stimType::inside_box(const ComMod& com_mod, const int Ac) const
+void SpatialBounds::set_box(const Vector<double>& min, const Vector<double>& max)
 {
-  if (box_min.size() < com_mod.nsd || box_max.size() < com_mod.nsd) {
+  box_min_ = min;
+  box_max_ = max;
+  region_type_ = (region_type_ == RegionType::sphere) ? RegionType::both : RegionType::box;
+}
```
michelebucelli: Here you should check that the arguments have correct (positive) and equal sizes.

Related to this, beware that #561 has changed the signature of `svmp::raise` (it should be enough to drop `SVMP_HERE`), and that it has introduced the convenient function `svmp::throw_if`.

Code/Source/solver/CepMod.cpp
```diff
@@ -33,117 +35,222 @@ bool stimType::is_active(const double time) const
...
+  region_type_ = static_cast<RegionType>(region_type_int);
+
+  if (region_type_ == RegionType::box || region_type_ == RegionType::both) {
+    int box_size = box_min_.size();
+    cm.bcast(cm_mod, &box_size);
+
+    if (box_size <= 0) {
+      svmp::raise<svmp::FE::InvalidArgumentException>(
+          SVMP_HERE, "Stimulus box has invalid coordinate dimension.");
+    }
```
michelebucelli: I don't think this check should go here, but rather in `set_box`. The same applies to the similar check for the sphere center below.

The logic behind this is: one of the advantages of encapsulation is that it allows to enforce validity constraints on the state of the object, by placing checks inside the methods that modify it. Once those checks are in place, you can and should assume that the object is always in a valid state (which in this case means: if the region is a box, then the box bounds have the correct size).

Code/Source/solver/CepMod.cpp
```diff
@@ -33,117 +35,222 @@ bool stimType::is_active(const double time) const
...
+    Vector<double> box_max(box_max_vals.size());
+
+    for (int i = 0; i < static_cast<int>(box_min_vals.size()); i++) {
+      if (box_min_vals[i] > box_max_vals[i]) {
+        svmp::raise<svmp::ParseException>(
+            SVMP_HERE, "Stimulus box Minimum values must be less than or equal to Maximum values.");
+      }
+
+      box_min(i) = box_min_vals[i];
+      box_max(i) = box_max_vals[i];
```
michelebucelli: Minor: I suggest using `operator[]` for accessing `Vector`. It does have both `operator()` and `operator[]`, and they do the same thing, but `operator()` has more ambiguous semantics (is this a function call, or a callable object, or something else?) while `operator[]` is more aligned with the language's conventions.
```cpp
box_min[i] = box_min_vals[i];
box_max[i] = box_max_vals[i];
```

michelebucelli (commented):
Thanks @samibismar! I left a few more comments, but it's mostly minor stuff.

Code/Source/solver/CepMod.cpp
samibismar: Changed `is_active()` to use `std::fmod(time, cycle_length)` as suggested.

Code/Source/solver/read_files.cpp
samibismar: Replaced the `RegionType` enum with `has_box` / `has_sphere` boolean flags. This keeps the current intersection behavior simple and avoids needing a new enum member for every possible combination of active bounds.

I agree that a more general bounds hierarchy may make sense later if many shape types are added, but for the current box/sphere case the flags seem simpler.

Code/Source/solver/CepMod.h
samibismar: Moved `SpatialBounds` inside `stimType`, so the nested type is now `stimType::SpatialBounds`.

Code/Source/solver/CepMod.h
samibismar: Removed the trailing underscores from the private `stimType` members.

Code/Source/solver/CepMod.cpp
samibismar: Updated the local `Vector` access here to use `operator[]`.

Code/Source/solver/CepMod.cpp
samibismar: Added size/equality checks to `set_box()` and `set_sphere()` using `svmp::throw_if`.

Code/Source/solver/CepMod.cpp
samibismar: Moved the validity checks into `set_box()` and `set_sphere()`. `distribute()` now assumes the object is already in a valid state, which is guaranteed by the setters.

samibismar:
Thank you for the review @michelebucelli. I addressed the latest comments in the newest commit.

Code/Source/solver/CepMod.cpp
```diff
@@ -24,6 +31,219 @@ const std::map<std::string,ElectrophysiologyModelType> cep_model_name_to_type
...
+
+  if (!is_active(time)) {
+    return 0.0;
+  }
+
+  if (!spatial_bounds.contains(x)) {
+    return 0.0;
+  }
+
+  return amplitude;
```
michelebucelli:
```suggestion
  if (utils::is_zero(amplitude) ||
      !is_active(time) ||
      !spatial_bounds.contains(x)) {
    return 0.0;
  } else {
    return amplitude;
  }
```

Code/Source/solver/CepMod.h
```diff
@@ -44,21 +46,61 @@ static std::ostream &operator << ( std::ostream& strm, ElectrophysiologyModelTyp
...
+        bool has_box = false;
+        bool has_sphere = false;
+
+        Vector<double> box_min;
+        Vector<double> box_max;
+        Vector<double> sphere_center;
+        double sphere_radius = 0.0;
+
+        bool inside_box(const Vector<double>& x) const;
+        bool inside_sphere(const Vector<double>& x) const;
```
michelebucelli, line 90: Can you please add Doxygen-style documentation to these members? A one-line comment should be enough for most of them (although maybe the functions `inside_box` and `inside_sphere` should document the behavior for when they are called with `has_box == false` or `has_sphere == false`).

Code/Source/solver/CepMod.h
```diff
@@ -44,21 +46,61 @@ static std::ostream &operator << ( std::ostream& strm, ElectrophysiologyModelTyp
...
+    double cycle_length = 0.0;
+    double amplitude = 0.0;
 
-    /// @brief cycle length
-    double CL = 0.0;
+    SpatialBounds spatial_bounds;
 
-    /// @brief stimulus amplitude
-    double A = 0.0;
+    bool is_active(const double time) const;
```
michelebucelli, line 116: Same as above, I'd include Doxygen documentation for these members.

Code/Source/solver/CepMod.h
```diff
@@ -19,6 +19,8 @@
 #include <map>
 #include <memory>
 
+class StimulusParameters;
+
```
michelebucelli: I'd move this down to line 52, after the other forward declarations.

michelebucelli (approved):
Thank you @samibismar! I approve the changes, pending the few cosmetic and documentation changes suggested below.

javijv4:
Hi @samibismar, everything looks good to me, but I have a request that might be out of scope for this PR, so feel free to skip it. I can always go back to this if I need to. 

For the application I am interested in, i.e., cardiac resynchronization therapy, I need to define two stimuli. With your geometrical definition of the stimuli, I was thinking I could just use two stimuli defined by two non-intersecting spheres. However, the current code only allows one `Stimulus` per domain, so to do this, I would have to define two domains. This is not terrible, though, since what I did before was to generate the domains to define the stimulus anyway. However, it would be great if I could avoid creating domains solely to define the stimulus, and I could just do something like this:

```
<Domain id="1">
  <Electrophysiology_model> AP </Electrophysiology_model>
  <Isotropic_conductivity> 1.0 </Isotropic_conductivity>
  <ODE_solver> Euler </ODE_solver>
  <Stimulus type="Istim1">
    <Amplitude> 10.0 </Amplitude>
    <Start_time> 0.0 </Start_time>
    <Duration> 2.0 </Duration>
    <Spatial_bounds>
      <Sphere>
        <Center> 0.1 0.2 0.0 </Center>
        <Radius> 0.05 </Radius>
      </Sphere>
    </Spatial_bounds>
  </Stimulus>
  
  <Stimulus type="Istim2">
    <Amplitude> 10.0 </Amplitude>
    <Start_time> 10.0 </Start_time>
    <Duration> 2.0 </Duration>
    <Spatial_bounds>
      <Sphere>
        <Center> 0.5 0.3 0.0 </Center>
        <Radius> 0.05 </Radius>
      </Sphere>
    </Spatial_bounds>
  </Stimulus>
</Domain>
```

The more I think about it, the more I think this should be another issue / PR, but let me know what you think. @michelebucelli, do you have any input on this?

michelebucelli:
@javijv4 that's a good suggestion! @samibismar and I have actually been talking about having a follow-up issue and PR to support multiple stimuli.

I think the structure you propose looks good, and I see no reason why to limit ourselves to two stimuli - I think the natural change would be to go from one single `stimType` instance to a `std::vector<stimType>`, populated with one element for every `<Stimulus>` tag in the XML file.

I think it might make sense to do this as a PR immediately after this one is merged, but I'm also open to adding it as a feature to this PR if you prefer and if @samibismar agrees.

On a (partially) related note, do you know what the `type` attribute to the `Stimulus` tag does?

Code/Source/solver/CepMod.cpp
samibismar: Combined the checks into a single condition as suggested.

Code/Source/solver/CepMod.h
samibismar: Moved this below the other forward declarations.

Code/Source/solver/CepMod.h
samibismar: Added Doxygen comments for the `SpatialBounds` data members and helper methods, including how the flags control the box and sphere checks.

javijv4 (approved):

javijv4:
A follow-up issue sounds good!

About the `type`, I don't really know. I thought it was similar to the `name` in `Add_face` for example, or something like an ID to identify the specific stimulus, but I looked at the code, and it doesn't seem to be doing anything. Maybe at some point there were plans to create another type of stimulus, but they were never implemented?

samibismar:
Thanks @javijv4 and @michelebucelli.

I can open a follow-up issue for supporting multiple `Stimulus` elements per domain and reference this PR for context. Would you both prefer that I open it now, or wait until this PR is merged?

Regarding the current `type` attribute, should I include removing it in the follow-up issue, unless there is a reason to preserve it?

michelebucelli:
> I can open a follow-up issue for supporting multiple Stimulus elements per domain and reference this PR for context. Would you both prefer that I open it now, or wait until this PR is merged?

Feel free to go ahead and open the issue, thanks!

> Regarding the current type attribute, should I include removing it in the follow-up issue, unless there is a reason to preserve it?

I say let's keep it for now, and maybe revisit as part of the issue discussion.

I'll go ahead and merge this PR in the meantime.

PR #584, opened 2026-07-09
Fixes for Intel OneAPI compilers and Intel MPI

### LINKED ISSUE #582: Errors with Intel OneAPI compiler
### Description

I am trying to build and run `svMultiPhysics` on the [TACC Lonestar6](https://docs.tacc.utexas.edu/hpc/lonestar6/) cluster.

Intel compilers are recommended on Intel clusters (and previous experience suggests that they can substantially outperform GCC due to better optimization), and the OneAPI family is needed (as opposed to the legacy, pre-OneAPI family) to support C++20 features. The compiler and MPI versions I am using are:
```
icx --version    --> Intel(R) oneAPI DPC++/C++ Compiler 2024.1.0 (2024.1.0.20240308)
mpirun --version --> Intel(R) MPI Library for Linux* OS, Version 2021.12 Build 20240410 (id: 49f2f80)
```

I am encountering multiple issues, from compilation to test configuration to mismatch with reference solutions. I am opening this issue to keep track of the problems that I encounter and (hopefully) of their solution.

I'd love to hear if anyone else has used `svMultiPhysics` with this compiler family, and if they've encountered the same issues.

- [x] Compilation error:
    ```
    /work/11433/mbucelli/svMultiPhysics/Code/Source/solver/mat_fun.h:54:104: error: no member named 'str' in 'std::basic_ostream<char>'
    54 |           auto dest_dims = (std::stringstream() << "(" << dest.nrows()  << "x" << dest.ncols() << ")").str();
    ```
    Easily fixed by using `std::to_string` instead of `std::stringstream`.
- [x] Parallel tests cannot run because Intel's `mpirun` does not have the `--oversubscribe` flag.
- [x] Serial tests fail due to large differences between the computed and reference solutions. There's relative differences up to $10^3$, so it seems something is catastrophically wrong. I'll report more details on this once I have addressed the issues above.
- [x] Also, it would be nice if, in the script `conftest.py`, any errors from `subprocess.call` were caught and returned in the test error message. Right now the tests fail because `svmultiphysics` generates no output, so the message is `RuntimeError: No svMultiPhysics output`, which masks the actual problem.

### Reproduction

Try to build and test `svMultiPhysics` with the compiler `Intel(R) oneAPI DPC++/C++ Compiler 2024.1.0 (2024.1.0.20240308)`.

### Expected behavior

The code should compile, and the tests should pass.

### Additional context

_No response_

### Code of Conduct

michelebucelli:
Fixes #582.

## Current situation

The code failed to build and test successfully on the TACC Lonestar6 cluster, due to switching to Intel OneAPI compilers and Intel MPI.

## Release Notes 

1. Fixed a compilation error in `mat_fun.h`.
2. Modified `conftest.py` so that it checks the version of `mpirun`, and if it is OpenMPI turns on the flag `--oversubscribe`, while if it isn't no flag is specified (in Intel MPI and `mpich`, oversubscription is enabled by default and there's no flag to control it).
3. Modified the CMake configuration scripts to set the flags `-fp-model=precise -fno-fast-math` when compiling with Intel OneAPI (otherwise, the compiler enables `-fp-model=fast` by default).
4. Modified `conftest.py` to catch the error messages raised by `svmultiphysics` if it fails, and use those as test error messages instead of `No svMultiPhysics output`.
5. Modified `conftest.py` to detect whether Trilinos or PETSc are present based on `CMakeCache.txt`, instead of whether the OS is MacOS or not.

## Documentation

None of the changes are user- or developer-facing, so they did not reflect on the Doxygen documentation. Not sure if the testing infrastructure is documented anywhere, in which case I'd be happy to amend that documentation.

## Testing

Automatic tests pass successfully on my laptop (MacOS Tahoe, compiling with Apple Clang and OpenMPI) and on TACC Lonestar6 (Linux cluster, compiling with Intel OneAPI and Intel MPI).


ktbolt (approved):
@michelebucelli Excellent updates !

tests/conftest.py
```diff
@@ -3,17 +3,78 @@
...
+    not HAS_PETSC,
+    reason="svMultiPhysics not built with PETSc (SV_PETSC_DIR empty in CMakeCache.txt)",
+)
+skip_if_no_trilinos = pytest.mark.skipif(
+    not HAS_TRILINOS,
+    reason="svMultiPhysics not built with Trilinos (SV_USE_TRILINOS=OFF in CMakeCache.txt)",
+)
+
+
+def _detect_oversubscribe_flag():
```
aabrown100-git, line 56: @michelebucelli Do you want to update the documentation (https://github.com/SimVascular/svMultiPhysics#software-dependencies) to say that Intel MPI or MPICH is also supported?

aabrown100-git (changes requested):
Very helpful changes! Just one comment about documentation.

tests/conftest.py
michelebucelli: Sure! I haven't tested `mpich` though, so I'd rather say something conservative like: "an MPI implementation, used for parallel processing. svMultiPhysics was tested against OpenMPI and Intel MPI".

We can go back and update it as soon as we confirm that it also works seamlessly with `mpich` (which it should - Intel MPI is based on `mpich`).

aabrown100-git (approved):
Approve!

PR #585, opened 2026-07-11
Fix incorrect indices in assembling weak Dirichlet conditions for the fluid

### LINKED ISSUE #583: Possible bug in weakly-applied Dirichlet conditions in fluid
While working on #582, Claude stumbled across these lines, and flagged them as a possible error:
https://github.com/SimVascular/svMultiPhysics/blob/a371185c3d2d9bfd81817eb962cf7133e49e5c60/Code/Source/solver/fluid.cpp#L460-L473

I think that the indices on the left and right hand sides of the assignments should be the same, as is done in similar loops in the same file (e.g. the similar loop in 2D). However, I'm not familiar enough with the fluid solver and with the assembly routines to be certain about that. I thought it'd be a good idea to report this in any case.

@aabrown100-git @hanzhao2020 what do you think?

michelebucelli:
Fixes #583.

There was a mismatch in the indices used to fill the local tangent in `bw_fluid_3d`, for assembling weak Dirichlet conditions.

This PR fixes the indices mismatch.

There is one test that for this feature (`fluid/pipe_RCR_weak_dir_3d`), and its reference solution needed to be updated as a result of this change.

### DIFF
--- a/Code/Source/solver/fluid.cpp
+++ b/Code/Source/solver/fluid.cpp
@@ -462,13 +462,13 @@ void bw_fluid_3d(ComMod& com_mod, const int eNoNw, const int eNoNq, const double
       double T1 = wl*Nq(a)*Nw(b);
 
       // dRc_a/du_b1
-      lK(12,a,b) = lK(13,a,b) - T1*nV(0);
+      lK(12, a, b) = lK(12, a, b) - T1 * nV(0);
 
       // dRc_a/du_b2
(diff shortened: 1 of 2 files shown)


ktbolt (approved):
Looks good, I hope the last of my indexing mistakes !

aabrown100-git (approved):
Great find! I left a comment in the issue that this change improves nonlinear convergence for `fluid/pipe_RCR_weak_dir_3d`. Very useful!

Issue #580, opened 2026-07-02
Implementing Quadrature Module


zasexton:
### Problem

Quadrature rules are still defined through Fortran-style, per-element dispatch tables that hard-code Gauss point counts, weights, and reference coordinates directly into initialization routines. This keeps quadrature setup tightly coupled to mesh storage, basis evaluation, face initialization, and physics assembly. A modular `FE/Quadrature` folder is therefore necessary because it gives us a single mesh-agnostic interface for reference-element integration rules, plus centralized construction, caching, validation, and compatibility pathways for both existing solver behavior and future finite-element extensions.

 - Quadrature setup is implemented as hard-coded element tables. `Code/Source/solver/nn_elem_gip.h` stores per-element and per-face point/weight data directly in dispatch maps, while `Code/Source/solver/nn.cpp` allocates and populates solver-facing `w` and `xi` arrays procedurally.
  - Quadrature state is embedded directly in solver containers rather than represented by reusable rule objects. `Code/Source/solver/ComMod.h` stores `nG`, `w`, `xi`, `xib`, shape tables, and quadrature modifiers inside `mshType`, `faceType`, and `fsType`, which makes integration rules hard to test or replace independently.
  -  Low-order simplex modifiers such as `qmTRI3` and `qmTET4` are encoded as explicit position-based quadrature rules with documented validity and exactness behavior.
  - Higher-order and extensible integration paths are not centralized. Efficient symmetric triangle/tetrahedron rules, tensor-product rules, surface rules, adaptive rules, and future cut or singular integration support should be developed and validated in one module instead of spreading through mesh, basis, geometry, and assembly code.

  ### Solution

To address the current solver’s quadrature infrastructure, we should create a dedicated, mesh-agnostic `FE/Quadrature` module with a common `QuadratureRule` interface for reference-space points, weights, polynomial order, reference measure, and stable cache identity. Quadrature rules should be constructed through a factory keyed by element type, requested order, and quadrature family, with caching for reusable rules and explicit compatibility paths for existing solver behavior.

[...]

PR #593, opened 2026-07-24
FE Quadrature Phase 01 abstract contract

zasexton:
## Summary

This pull request establishes Phase 01 of the finite-element Quadrature module.
It defines the immutable reference-rule contract that concrete rule generation,
selection, and consumer migration can build on in later phases.

Related to #580.

The change:

- Adds an abstract, immutable `QuadratureRule` interface for canonical
  reference cells.
- Defines ordered point/weight storage using the FE fixed-size vector type.
- Exposes cell family, reference dimension, polynomial exactness,
  reference-cell measure, and const sample access.
- Establishes one construction-time validation boundary for provider data.
- Adds focused unit coverage for the new representation, validation boundary,
  and immutable query interface.
- Registers the Quadrature sources with the solver build and the module with
  the FE documentation hierarchy.

### Immutable rule contract

`QuadratureRule` owns one complete set of ordered reference points and weights.
Successful construction establishes the invariant, after which consumers use
only const queries. Rule objects are non-copyable and non-movable and are
intended to be created once and shared through const ownership.

Concrete providers construct a protected `RuleData` payload containing:

- Declared total-degree polynomial exactness.
- Ordered canonical reference points.
- Corresponding ordered weights.

The base class derives dimension and reference-cell measure from `CellFamily`,
preventing providers from supplying redundant topology metadata.

### Supported reference cells

The Phase 01 contract covers:

- Point
- Line
- Triangle
- Quadrilateral
- Tetrahedron
- Hexahedron
- Wedge

Pyramid, polygon, polyhedron, and unknown families remain explicitly
unsupported.

### Construction validation

Construction rejects:

- Unsupported cell families.
- Negative polynomial exactness.
- Empty or mismatched point/weight storage.
- Non-finite coordinates or weights.
- Points outside the declared canonical reference cell.
- Weights that do not reproduce the reference-cell measure within the scaled
  validation tolerance.

Duplicate points and zero or negative individual weights remain admissible when
all other invariants hold.

The weight sum is evaluated exactly from the stored binary64 values using a
fixed-storage accumulator. Validation is therefore independent of point/weight
ordering, cancellation, intermediate floating-point overflow, and the host
`long double` representation. Compile-time checks document and enforce the
binary64 value model and object layout required by that accumulator; this
will be important for cross-platform compile time validation. 

### Phase 01 regression tests

The Phase 01 tests are intentionally limited to the new Quadrature
infrastructure. They verify:

- The fixed-size `QuadPoint` representation.
- The immutable `QuadratureRule` query interface.
- Supported and unsupported reference-cell families.
- Metadata and point/weight storage invariants.
- Finite values, inactive coordinates, and canonical reference-cell
  containment.
- Reference-cell measure validation, including admissible negative weights.
- Exact binary64 weight summation under cancellation, subnormal values,
  reordered inputs, and large rules.
- Construction tolerances for reference coordinates and weight normalization.

These tests do not exercise existing quadrature tables, solver-specific selection
paths, or integration consumers.

## Supporting changes

- Removes the unused speculative `QuadratureType` enumeration from the shared
  FE vocabulary.
- Adds `FE/Quadrature` sources to the solver target.
- Adds the Quadrature module to the top-level FE documentation hierarchy.
- Runs Doxygen from the repository root and removes stale generated output
  before documentation generation, ensuring relative paths resolve
  consistently.

## Scope and non-goals

This pull request intentionally does not:

- Add production concrete quadrature providers.
- Add a quadrature factory, cache, or rule-selection policy.
- Migrate solver containers or integration consumers.
- Change basis selection, reduced-integration policy, or physical-space
  mapping.

## Validation performed

- GCC build of `run_all_unit_tests` and `svmultiphysics`.
- GCC unit suite: 245 of 245 tests passed.
- Doxygen validation with no Quadrature-specific warnings.


zasexton:
@ktbolt @michelebucelli let me know what you think of the abstract class contract for quadrature. I believe the public API for this contract is minimal and clean while ensuring that explicit Quadrature rules that are created can be validated. As always, happy to have more design input here and will incorporate changes pretty quickly so we can begin making a concrete QuadratureRule class.

Code/CMake/SimVascularExternals.cmake
```diff
@@ -9,8 +9,10 @@ if(DOXYGEN_FOUND)
   configure_file(${SV_SOURCE_DIR}/../Documentation/Doxyfile
     ${SV_BINARY_DIR}/Doxyfile @ONLY)
   add_custom_target(doc
-    ${DOXYGEN_EXECUTABLE} ${SV_BINARY_DIR}/Doxyfile
-    WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+    COMMAND "${CMAKE_COMMAND}" -E remove_directory
+      "${SV_SOURCE_DIR}/../Documentation/build"
+    COMMAND "${DOXYGEN_EXECUTABLE}" "${SV_BINARY_DIR}/Doxyfile"
+    WORKING_DIRECTORY "${SV_SOURCE_DIR}/.."
```
michelebucelli, line 15: If I understand this right, the purpose of this is to fix the `make doc` recipe (whereas as of now the documentation would have to be built by calling `doxygen` manually from the repository root folder). Is this correct?

If so, is the documentation now generated into the build folder, or into the source folder as in the current main branch?

Code/Source/solver/CMakeLists.txt
```diff
@@ -266,11 +266,17 @@ file(GLOB SOLVER_FE_MATH_SRCS CONFIGURE_DEPENDS
...
+  FE/Quadrature/*.h
+)
+
 list(APPEND CSRCS
   ${SOLVER_CORE_SRCS}
   ${SOLVER_FE_COMMON_SRCS}
   ${SOLVER_FE_BASIS_SRCS}
   ${SOLVER_FE_MATH_SRCS}
+  ${SOLVER_FE_QUADRATURE_SRCS}
 )
```
michelebucelli, line 281: Wouldn't it be better to have separate `CMakeLists.txt` files in the individual subfolders, and to build separate binaries that are then linked into the executable? The main `CMakeLists.txt` would then recurse into subfolders, instead of listing all the files.

In my opinion this would encourage the clean separation of the different modules, and it might be a good idea to start implementing it at least for the newly refactored parts of the library.

Code/Source/solver/FE/Quadrature/QuadratureRule.h
```diff
@@ -0,0 +1,304 @@
...
+#define SVMP_FE_QUADRATURE_RULE_H
+
+/**
+ * @file QuadratureRule.h
+ * @brief Immutable reference-space quadrature rule contract.
+ * @ingroup FE_Quadrature
+ *
+ * This header defines the consumer-facing representation of a finite-element
+ * quadrature rule. Rule construction and validation are implemented separately
+ * so consumers depend only on the stable query interface.
```
michelebucelli: Minor: I would drop "finite-element" here, since the concept of a quadrature formula is independent of what one does with it (generic integrals or FE integrals).

Code/Source/solver/FE/Quadrature/QuadratureRule.h
```diff
@@ -0,0 +1,304 @@
...
+ * @f]
+ * A rule identifies its reference-cell family, reports its dimension and
+ * declared polynomial exactness, and keeps every point paired with its
+ * corresponding weight.
+ *
+ * The module does not choose the exactness required by an equation term, apply
+ * reduced-integration policy, select a basis, own mesh storage, embed or orient
+ * a face in its parent element, or map a reference integral into physical
+ * space. Those operations require solver, basis, mesh, and geometry context and
+ * remain at the caller's integration boundary.
```
michelebucelli: In the same spirit as my previous comment about mentioning finite elements, I think this paragraph addresses concerns that are somewhat unnatural in the context of quadrature (or at least should be), and so I feel that it might risk creating confusion in the reader, rather than clarifying. I suggest rephrasing as follows:
> This module is not concerned with selecting the quadrature rule based on the requirement of any specific application (e.g. on what functions must be integrated exactly). Furthermore, this module only deals with quadrature rules on reference cells, not on transformed cells. Those concerns are left to the consumers of this module.

More in general, I appreciate (a lot!) the level of detail in this documentation. However, in some places I feel that it uses jargon that is appropriate in a software engineering context but might be obscure to people without a structured software engineering background. Since I feel that many of the contributors to `svmp` might have a different background (e.g. applied mathematics or biomedical engineering), I think it might be beneficial to shift to a slightly more concrete and practical language. I will try to give some concrete suggestions in this direction in my other comments.

(The jargon issue might also be addressed by providing some sort of glossary, but I feel it might make things harder rather than easier to read).

Code/Source/solver/FE/Quadrature/QuadratureRule.h
```diff
@@ -0,0 +1,304 @@
...
+ * ## Public API
+ *
+ * @ref svmp::FE::quadrature::QuadPoint "QuadPoint" and the const query surface
+ * of @ref svmp::FE::quadrature::QuadratureRule "QuadratureRule" form the public
+ * API. Integration consumers read rule metadata and ordered samples; they do
+ * not derive new rules or modify rule storage.
+ *
+ * Complete-data construction, reference-cell traits, point-containment checks,
+ * exact weight summation, concrete generators, caches, and rule-selection
+ * facilities are module implementation details.
```
michelebucelli: In the vein of my previous comment, I suggest rephrasing as follows (or similar):
> The public API exposes the ordered weight-point pairs from the quadrature rule through the methods point, weight, points and weights of the QuadratureRule class. Additionally, a number of methods are exposed to access metadata about the rule such as the number of points or the degree of exactness. Other modules using quadrature formulas should exclusively rely on this information, and not on specific details of the concrete quadrature formulas being used.

(inserting Doxygen `@ref`s appropriately so that the listed methods are linked back to their documentation).

Code/Source/solver/FE/Quadrature/QuadratureRule.h
```diff
@@ -0,0 +1,304 @@
...
+ * ## Rule-provider contract
+ *
+ * Concrete rule providers are the only supported subclasses. A provider builds
+ * one complete RuleData payload before invoking the protected constructor and
+ * exposes no mutation afterward. It must advertise only exactness established
+ * for every rule it supplies through analytic moment tests. Derivation is a
+ * provider extension seam, not an integration-consumer customization point.
+ * The protected constructor is the only provider extension point; public
+ * metadata and sample queries are fixed, nonvirtual operations on validated
+ * base-class state.
```
michelebucelli: This paragraph feels a bit arcane to me too 😄 Could you try to clarify what is intended, so that we can try to make it feel a bit more concrete?

My best understanding is that this section is meant to explain how concrete quadrature rules are meant to be implemented. Personally, I find that providing a "shopping list" of things that need to be implemented (or called, or exposed) by derived classes is usually an effective way of conveying this information.

If this is indeed what this paragraph is meant to explain, a proposed rephrasing might be the following (I almost surely got something wrong in how this is intended to work, and even the part that I might have got right might change as part of this review, but it should at least give an idea of what I mean).
> ## Implementing concrete quadrature rules
>
> Concrete quadrature rules should be implemented by deriving from the QuadratureRule class. The derived class should, in its constructor, build an object of type RuleData and pass it to the protected QuadratureRule constructor. The RuleData object bundles polynomial exactness information with weights and point coordinates.

Code/Source/solver/FE/Quadrature/QuadratureRule.h
```diff
@@ -0,0 +1,304 @@
...
+ * at most @f$p@f$. A rule can integrate selected higher-degree polynomials
+ * without increasing that common guarantee. Every concrete provider is
+ * responsible for establishing its advertised exactness with analytic moment
+ * tests.
+ *
+ * Points use one fixed-size three-component representation. Providers initialize
+ * all three coordinates explicitly because the Eigen-backed vector is not
+ * zero-initialized by default. Only the first dimension() components
+ * are active, and every inactive component is zero within the coordinate
+ * tolerance. The supported canonical domains are:
```
michelebucelli: In the same spirit as before, I would recast this section as a bullet list of the constraints that the class enforces. This would, in my opinion, make it easier to parse and understand. Additionally, I suggest reporting what happens if the constraints get violated (e.g. an exception is thrown).

Furthermore, I would prioritize concreteness in defining the constraints. For example, I interpret "mismatched storage" as "weights and points are not in the same number", and the latter seems to me more concrete and immediate to interpret.

Code/Source/solver/FE/Quadrature/QuadratureRule.h
```diff
@@ -0,0 +1,304 @@
...
+ * unsupported.
+ *
+ * ## Ownership and lifetime
+ *
+ * Rule objects are non-copyable and non-movable. They are intended to be built
+ * once, retained through a const owning handle, and shared across integrations.
+ * References returned by points() and weights() remain valid for the lifetime
+ * of the rule. Consumers should retain that rule rather than copying its point
+ * count or samples into parallel authoritative state. Concurrent const access
+ * is safe while an owning handle keeps the rule alive.
```
michelebucelli: What are the reasons behind this decision?

In my opinion, quadrature rules are relatively lightweight objects (I have heard stories of monster quadrature rules with thousands of points, but I think in practice everyone mostly uses rules with a few tens of points). I fear that non-copyability might be excessively constraining in some settings, and it might be annoying to keep track of object ownership, but maybe I'm missing something about how this is going to look once concrete rules are in place.

Code/Source/solver/FE/Quadrature/QuadratureRule.h
```diff
@@ -0,0 +1,304 @@
...
+     *
+     * Integration consumers do not construct this payload. A provider computes
+     * all three fields before invoking the protected base constructor and must
+     * substantiate its declared exactness with analytic moment tests.
+     */
+    struct RuleData {
+        int polynomial_exactness{-1};    ///< Declared total-degree polynomial exactness.
+        std::vector<QuadPoint> points;   ///< Ordered canonical reference coordinates.
+        std::vector<double> weights;     ///< Weights paired with points in the same order.
+    };
```
michelebucelli: What is the reason behind passing the `RuleData` object instead of individually passing its three members? The latter, in my opinion, would be a bit more lightweight and self-documenting (in the sense that e.g. the constructor signature would be more obviously and directly interpretable without having to go to another `struct`'s documentation).

Code/Source/solver/FE/Quadrature/QuadratureRule.h
```diff
@@ -0,0 +1,304 @@
...
+     * @param family Supported canonical reference-cell family.
+     * @param data Complete exactness, point, and weight payload.
+     * @note Duplicate points and zero or negative weights remain admissible
+     * when every other rule invariant is satisfied.
+     * @throws InvalidArgumentException If the family is unsupported, exactness
+     * is negative, storage is empty or mismatched, a value is non-finite, a point
+     * is outside the reference cell, or the weights do not reproduce the
+     * reference-cell measure within the scaled measure tolerance.
+     */
+    explicit QuadratureRule(svmp::CellFamily family, RuleData data);
```
michelebucelli: How is construction intended to work for quadrature rules for which the computation of weights and points needs to be deferred?

For example, suppose I want to define a tensor-product Gaussian quadrature rule, which it makes sense to construct using only the number of points (or, equivalently, the degree of exactness) along each direction. The ideal constructor signature for that would be, in my opinion:
```cpp
QuadratureGauss(const unsigned int n_points) :
  QuadratureRule(...)
{}
```
The arguments for the `QuadratureRule` constructor would be quite annoying to build in the one-liners needed by the initializer list. There definitely are ways around this, but they feel a bit cumbersome to me.

Perhaps this question is downstream of a more general misunderstanding of how you envision concrete rules to be implemented in the future.

Code/Source/solver/FE/Quadrature/QuadratureRule.cpp
```diff
@@ -0,0 +1,444 @@
...
+            std::numeric_limits<double>::denorm_min()) == 1u,
+    "Quadrature validation requires the standard binary64 bit layout");
+
+/**
+ * @brief Exact, order-independent sum of finite binary64 values.
+ *
+ * Every finite binary64 value is an integer multiple of 2^-1074. Positive and
+ * negative coefficients are accumulated separately in fixed-width unsigned
+ * integers, so neither cancellation nor an intermediate floating-point
+ * overflow can affect the result.
```
michelebucelli: I think it would be good to add one or two sentences about why this is necessary/useful (as opposed to just summing with the normal arithmetic operators).

Code/Source/solver/FE/Quadrature/QuadratureRule.cpp
```diff
@@ -0,0 +1,444 @@
...
+        case svmp::CellFamily::Tetra:
+            return ReferenceCellTraits{3, 1.0 / 6.0};
+        case svmp::CellFamily::Hex:
+            return ReferenceCellTraits{3, 8.0};
+        case svmp::CellFamily::Wedge:
+            return ReferenceCellTraits{3, 1.0};
+        default:
+            return std::nullopt;
+    }
+}
```
michelebucelli: Is there any legitimate situation where this should be called against an unsupported reference cell? If not, I would suggest throwing an exception in the `default` case, and dropping the `std::optional` altogether.

Code/Source/solver/FE/Quadrature/QuadratureRule.cpp
```diff
@@ -0,0 +1,444 @@
...
+
+    const double x = point[0];
+    const double y = point[1];
+    const double z = point[2];
+
+    bool is_contained = false;
+    switch (family) {
+        case svmp::CellFamily::Point:
+            is_contained = true;
+            break;
```
michelebucelli: Shouldn't this return true only if the provided point is coincident with the reference cell point (which I assume to be the origin [0, 0, 0])?

Code/Source/solver/FE/Quadrature/QuadratureRule.cpp
```diff
@@ -0,0 +1,444 @@
...
+ValidationResult validate_rule_data(
+    svmp::CellFamily family,
+    const ReferenceCellTraits& traits,
+    int polynomial_exactness,
+    const std::vector<QuadPoint>& points,
+    const std::vector<double>& weights) noexcept
+{
+    if (polynomial_exactness < 0) {
+        return {"polynomial exactness must be non-negative"};
+    }
```
michelebucelli: Wouldn't this constraint be better represented by defining the polynomial exactness to be an `unsigned int`?

Code/Source/solver/FE/Quadrature/QuadratureRule.cpp
```diff
@@ -0,0 +1,444 @@
...
+{
+    if (result.valid()) {
+        return {};
+    }
+
+    std::string message{"QuadratureRule: "};
+    message.append(result.reason);
+    if (result.sample != ValidationResult::no_sample) {
+        message.append(" at sample ");
+        message.append(std::to_string(result.sample));
```
michelebucelli: What is the meaning of "sample" in this context? Does it refer to the index of the quadrature point? In that case, I think making more explicit reference to the quadrature point index might be more easy to interpret, and slightly more in line with standard nomenclature.

Code/Source/solver/FE/Quadrature/QuadratureRule.cpp
```diff
@@ -0,0 +1,444 @@
...
+
+    return {
+        family,
+        traits->dimension,
+        data.polynomial_exactness,
+        traits->reference_cell_measure,
+        std::move(data.points),
+        std::move(data.weights),
+    };
+}
```
michelebucelli: This seems a bit more complicated than necessary to me. Wouldn't it be possible to merge all three methods (constructor from cell family and `RuleData`, constructor from `ValidatedState` and `validate`) into the first constructor only.

Unless there's some use cases that I am missing?

michelebucelli (commented):
Thank you @zasexton! I have left some comments. Some of them relate to the documentation, others to the interface itself.

In general, I am not sure I have understood correctly how the concrete rules are meant to be implemented (as you might see from some of my comments). In this respect, I think it would be useful to include in this PR at least one concrete rule (I assume you have implemented some, at least for the sake of testing the infrastructure), which would serve as a demonstration of how to use the abstract framework. I think this would make review a bit easier.

Code/Source/solver/CMakeLists.txt
ktbolt: @zasexton I agree with @michelebucelli; better to use CMake to manage sub-directories.

zasexton:
Thanks for these comments @michelebucelli I'll review these today and make appropriate changes. Overall there is a good bit of important design feedback here that will help me to improve this base abstraction (which is the main purpose of me posting this draft PR). Thank you!

Code/Source/solver/FE/Quadrature/QuadratureRule.h
ktbolt: @zasexton I must admit that I am not very familiar with some of the terminology you use and have not ever seen them used in the CAD, FEA, robotics, etc. companies I've worked in. Like @michelebucelli says these terms will not be meaningful to the typical developer. 

I am not quite on board with the first 100 lines of a file being comments; maybe this should be in a separate document. 

I've seen criticism of code that needs extensive commenting; perhaps should be more expressive, clean and simple.

Code/Source/solver/FE/Quadrature/QuadratureRule.cpp
```diff
@@ -23,12 +23,12 @@ namespace svmp::FE::quadrature {
...
 struct ReferenceCellTraits {
-    int integration_dimension;
-    int coordinate_dimension;
+    int dimension;
     double zeroth_moment;
 };
 
-constexpr double construction_validation_tolerance = 1.0e-12;
+constexpr double coordinate_validation_tolerance = 1.0e-12;
+constexpr double moment_validation_tolerance = 1.0e-12;
```
ktbolt: @zasexton Where do these values come from ?

Code/Source/solver/FE/Quadrature/QuadratureRule.cpp
```diff
@@ -48,19 +48,19 @@ constexpr std::optional<ReferenceCellTraits> reference_cell_traits(
...
+            return ReferenceCellTraits{2, 4.0};
         case svmp::CellFamily::Tetra:
-            return ReferenceCellTraits{3, 3, 1.0 / 6.0};
+            return ReferenceCellTraits{3, 1.0 / 6.0};
         case svmp::CellFamily::Hex:
-            return ReferenceCellTraits{3, 3, 8.0};
+            return ReferenceCellTraits{3, 8.0};
         case svmp::CellFamily::Wedge:
-            return ReferenceCellTraits{3, 3, 1.0};
+            return ReferenceCellTraits{3, 1.0};
```
ktbolt: @zasexton What do the float numbers mean ?

ktbolt (commented):
@zasexton Why all of this complexity to implement just the few quadrature rules needed for the elements we support ?

Code/CMake/SimVascularExternals.cmake
zasexton: The changes here should remove any stale documents generated prior to regenerating a doxygen build but they should not be relocating the written destination of generated. The `Documentation/build` is still the appropriate target location since this is used by current github actions.

zasexton:
Thanks for the patience! I have made my relevant edits to the PR branch and will be going through and addressing comments directly now!

Code/Source/solver/FE/Quadrature/QuadratureRule.cpp
zasexton: These have been removed now and will instead be included for tolerance based verification of quadrature on concrete rule implementations instead of the abstract class. Originally I was using these tolerances for more niche CutCell quadrature checking (which we will not be including in this series of Quadrature refactoring)

Code/Source/solver/CMakeLists.txt
zasexton: I agree and I will make this an immediate follow-on issue and PR since it has a very clear scope and resolution.

Code/Source/solver/FE/Quadrature/QuadratureRule.h
zasexton: This change of language was made in commit 413acf3c along with substantial reduction of commenting for the abstract class

Code/Source/solver/FE/Quadrature/QuadratureRule.h
zasexton: I agree and have since removed much of the verbose documentation for simple explanations of abstract concepts through commits ee8a9272 and 413acf3c . These simplify the rule documentation as well as the design (removing `RuleProbe`). These improve the ownership model and reduce the amount of terminology needed.

Code/Source/solver/FE/Quadrature/QuadratureRule.h
zasexton: I have made `QuadratureRule` a concrete value type; therefore we no longer have separate provider interface API concerns. Commit 413acf3c

Code/Source/solver/FE/Quadrature/QuadratureRule.h
zasexton: Originally I implemented this `RuleData` for an implementation of adaptive quadrature that I had been tinkering around with as an concrete class built on top of this base abstract class. but I think this can be done in a cleaner fashion without the need for this additional object. Therefore, I have removed this subclass to eliminate this documentation section and additional abstract machinery. Commit ee8a9272

Code/Source/solver/FE/Quadrature/QuadratureRule.h
zasexton: Yes, I've made these relevant changes in commit d7172e05 and documents that `InvalidArgumentException` is thrown when a requirement is violated in rule construction.

Code/Source/solver/FE/Quadrature/QuadratureRule.h
zasexton: The RuleData object was originally added to support abstraction for more exotic downstream quadrature (which is not necessary for this refactor), so I have removed it. Commit dbc6a6ca . We now pass constructor values directly for the abstract class

Code/Source/solver/FE/Quadrature/QuadratureRule.h
zasexton: Through commits dbc6a6ca and ee8a9272 we completely remove the initializer-list and derived class construction model. This simplifies the code further and eliminates some of the otherwise annoying API that would have been employed

Code/Source/solver/FE/Quadrature/QuadratureRule.cpp
zasexton: Ultimately I decided to remove this. I was being overly pedantic about cross-platform accumulator support for standard library utilities. Commits c0beffd2 and d7172e05 simplify the quadrature weight summation and validation structure. While such utilities might become necessary/useful in the future they do not need to be a part of the base abstract quadrature rule class. This was a hold-over from my original work moving between windows and linux testing machines for unit testing.

Code/Source/solver/FE/Quadrature/QuadratureRule.cpp
zasexton: We now throw `InvalidArgumentException` for unsupported cell families. See commit ae426e80

Code/Source/solver/FE/Quadrature/QuadratureRule.cpp
zasexton: I have decided to table a wider selection fo the tolerance-based validation so that this will be accomplished in the concrete quadrature rule generators. And yes, theoretically only exact-zero values should be excepted for point validation. See changes in commit d7172e05

Code/Source/solver/FE/Quadrature/QuadratureRule.cpp
zasexton: I decided to keep the `int` polynomial exactness so that FE basis order constructors can reject negative inputs explicitly. Specifically I was worried about silent conversion of negative values. Perhaps this is unreasonable but for now I believe that invalid values will be properly caught with error messages explaining why negative values are inappropriate.

Code/Source/solver/FE/Quadrature/QuadratureRule.cpp
zasexton: I renamed "sample" to "point_index" which I think is more straight-forward. It refers to the quadrature-point index. See commit fe657df4

Code/Source/solver/FE/Quadrature/QuadratureRule.cpp
zasexton: This is now done under one constructor that initializes data and performs simple structural checks directly. See the changes in commit fe657df4

Code/Source/solver/FE/Quadrature/QuadratureRule.h
zasexton: I have changed the abstract class for `QuadratureRule` so that point and weights remain read-only but now the rules are copyable and movable to avoid potential constraints. I don't believe this will need much changing as we construct concrete rules and we can always decide to add the constraints back should we need these types of protections. Commit ee8a9272

Code/Source/solver/FE/Quadrature/QuadratureRule.cpp
zasexton: I have removed the `ReferenceCellTraits` `struct` since this was just adding to ambiguity about the stored values. Now the explicit values that are in the source file are specifically the reference measures that are inherent to the types of reference cell families (e.g. 0.5 for unit triangle and 1/6 for unit tet). See commits ae426e80 , a7193428 , and 49d0568a . I believe that this simplifies the code further and makes the explicit floating point numbers obvious in their intentions.

zasexton:
@ktbolt and @michelebucelli let me know what you think of the changes and simplifications that have been made over the last 11 commits. I believe that I have addressed many of the original comments. Overall, I think that we are getting close to a clean abstract class here to complete phase 1 of the quadrature refactor.

Code/CMake/SimVascularExternals.cmake
michelebucelli: Thanks for clarifying. I understand this is the current state of the GitHub pipelines (and that it is unrelated to this PR), but.

In my local installation, running `make doc` from `build/svMultiPhysics-build` fails to generate the documentation. If I interpret the error correctly, this is because there is a mismatch between the paths used in the `Doxyfile` and the paths that `make doc` runs with. This command changes the target so that `make doc` runs `doxygen` from the repository's source directory, so I believe that these issues should be fixed.

Having said all this, do you know the rationale behind generating the documentation in a separate build folder, rather than in the already existing, CMake-configured build folder? I would think the latter to be the more natural thing to do, and I'm generally wary of situations where running `make` in a folder generates/modifies files outside of that folder.

Code/Source/solver/FE/Quadrature/QuadratureRule.h
```diff
@@ -0,0 +1,147 @@
...
+ *
+ * | Family   | Reference cell                                      | Measure |
+ * |----------|-----------------------------------------------------|---------|
+ * | Point    | @f$(0,0,0)@f$                                      | @f$1@f$ |
+ * | Line     | @f$[-1,1]@f$                                       | @f$2@f$ |
+ * | Triangle | @f$\{(x,y):x,y\geq0,\ x+y\leq1\}@f$             | @f$1/2@f$ |
+ * | Quad     | @f$[-1,1]^2@f$                                     | @f$4@f$ |
+ * | Tetra    | @f$\{(x,y,z):x,y,z\geq0,\ x+y+z\leq1\}@f$       | @f$1/6@f$ |
+ * | Hex      | @f$[-1,1]^3@f$                                     | @f$8@f$ |
+ * | Wedge    | unit triangle @f$\times[-1,1]@f$                   | @f$1@f$ |
```
michelebucelli: Minor: fixed the table alignment.

```suggestion
 * | Family   | Reference cell                            | Measure   |
 * |----------|-------------------------------------------|-----------|
 * | Point    | @f$(0,0,0)@f$                             | @f$1@f$   |
 * | Line     | @f$[-1,1]@f$                              | @f$2@f$   |
 * | Triangle | @f$\{(x,y):x,y\geq0,\ x+y\leq1\}@f$       | @f$1/2@f$ |
 * | Quad     | @f$[-1,1]^2@f$                            | @f$4@f$   |
 * | Tetra    | @f$\{(x,y,z):x,y,z\geq0,\ x+y+z\leq1\}@f$ | @f$1/6@f$ |
 * | Hex      | @f$[-1,1]^3@f$                            | @f$8@f$   |
 * | Wedge    | unit triangle @f$\times[-1,1]@f$          | @f$1@f$   |
```

Code/Source/solver/FE/Quadrature/QuadratureRule.h
```diff
@@ -0,0 +1,147 @@
...
+
+/**
+ * @brief Owning abstract base for quadrature rules on canonical reference cells.
+ *
+ * Construction requires:
+ *
+ * - a supported cell family and non-negative polynomial exactness;
+ * - at least one point and the same number of points and weights;
+ * - finite coordinates and weights; and
+ * - inactive coordinates equal to zero.
```
michelebucelli: Minor: I suggest reordering and rephrasing as follows:
```suggestion
 * - a vector of points, with at least one element;
 * - for all points, coordinates must be finite;
 * - for all points, coordinates beyond the spatial dimension of the cell family
 *   must be equal to zero;
 * - a vector of finite weights, with as many elements as the points.
```

michelebucelli: Less minor: are any of these constraints checked upon construction, or are they only required for the correct behavior of this object, but left to the user? I think this should be documented (same as the paragraph below documents that some other constraints are not verified).

Code/Source/solver/FE/Quadrature/QuadratureRule.h
```diff
@@ -0,0 +1,147 @@
...
+     * @brief Return point @p i without bounds checking.
+     * @pre @p i is less than num_points().
+     */
+    const QuadPoint& point(std::size_t i) const noexcept { return points_[i]; }
+
+    /**
+     * @brief Return the weight paired with point @p i without bounds checking.
+     * @pre @p i is less than num_points().
+     */
+    double weight(std::size_t i) const noexcept { return weights_[i]; }
```
michelebucelli, line 122: For both of these, I would document that the index `i` is not checked (perhaps this is implicit in the `@pre` statement, but better to be explicit IMO).

Code/Source/solver/FE/Quadrature/QuadratureRule.h
```diff
@@ -0,0 +1,147 @@
...
+     * @param polynomial_exactness Declared total-degree polynomial exactness.
+     * @param points Ordered reference coordinates.
+     * @param weights Weights paired with @p points.
+     * @throws InvalidArgumentException If a construction requirement is violated.
+     */
+    explicit QuadratureRule(
+        svmp::CellFamily family,
+        int polynomial_exactness,
+        std::vector<QuadPoint> points,
+        std::vector<double> weights);
```
michelebucelli: 1. It'd be better to pass these by const reference, I think.
2. Since `points` and `weights` are required to have the same size, wouldn't it be better to store them (or at least require the caller to provide them) as `std::vector<std::pair<QuadPoint, double>> points_and_weights`? This way, the same-size constraint would be impossible to violate. Should this be used for the data members as well, I think the overhead for access through the `weight()` and `point()` getters is essentially zero, while there'd be a bit more overhead for the vector accesses `points` and `weights`.

Code/Source/solver/FE/Quadrature/QuadratureRule.cpp
```diff
@@ -0,0 +1,158 @@
...
+            return 1;
+        case svmp::CellFamily::Triangle:
+        case svmp::CellFamily::Quad:
+            return 2;
+        case svmp::CellFamily::Tetra:
+        case svmp::CellFamily::Hex:
+        case svmp::CellFamily::Wedge:
+            return 3;
+        default:
+            return -1;
```
michelebucelli: I don't remember if I had already asked this (in which case, sorry for asking again), but: is there any legitimate case where this would happen? If not, I would favor raising an exception here, so that the illegal behavior is surfaced immediately.

The same applies to `reference_measure` below, and if you end up implementing it, you can probably drop the assertions in `QuadratureRule::dimension` and `QuadratureRule::reference_cell_measure` below.

Code/Source/solver/FE/Quadrature/QuadratureRule.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+#include <cassert>
+#include <cmath>
+#include <string>
+#include <utility>
+
+namespace svmp::FE::quadrature {
+namespace {
+
+constexpr int reference_dimension(svmp::CellFamily family) noexcept
```
michelebucelli: I think that this function, as well as `reference_measure` below, would be better placed in the same file as `CellFamily` (I don't remember where that is though).

Code/Source/solver/FE/Quadrature/QuadratureRule.cpp
```diff
@@ -0,0 +1,158 @@
...
+    for (std::size_t component = 0; component < 3u; ++component) {
+        if (!std::isfinite(point[component])) {
+            svmp::raise<InvalidArgumentException>(
+                std::string{
+                    "QuadratureRule: quadrature point contains a non-finite "
+                    "coordinate at point index "} +
+                std::to_string(point_index));
+        }
+        if (component >=
+            static_cast<std::size_t>(dimension) &&
```
michelebucelli: Why not make `dimension` a `size_t` instead?

Code/Source/solver/FE/Quadrature/QuadratureRule.cpp
```diff
@@ -0,0 +1,158 @@
...
+            return -1.0;
+    }
+}
+
+void validate_point(
+    const QuadPoint& point,
+    int dimension,
+    std::size_t point_index)
+{
+    for (std::size_t component = 0; component < 3u; ++component) {
```
michelebucelli:
```suggestion
    for (std::size_t component = 0; component < point.size(); ++component) {
```
as a safeguard against future changes of that size.

Code/Source/solver/FE/Quadrature/QuadratureRule.cpp
```diff
@@ -0,0 +1,158 @@
...
+}
+
+} // namespace
+
+QuadratureRule::~QuadratureRule() = default;
+
+int QuadratureRule::dimension() const noexcept
+{
+    const int dimension = reference_dimension(cell_family_);
+    assert(dimension >= 0);
```
michelebucelli: If you opt to keep this check here, I would suggest using one of the exceptions from `Core/Exception.h` or `FE/Common/FEException.h`, ideally through `svmp::check` as below.

Code/Source/solver/FE/Quadrature/QuadratureRule.cpp
```diff
@@ -0,0 +1,158 @@
...
+        case svmp::CellFamily::Hex:
+            return 8.0;
+        case svmp::CellFamily::Wedge:
+            return 1.0;
+        default:
+            return -1.0;
+    }
+}
+
+void validate_point(
```
michelebucelli: For symmetry, I would either turn this into `validate_points` and give it an interface similar to `validate_weights` (i.e. work on the whole vector), or vice versa change `validate_weights` to `validate_weight` and have it work on a single weight at a time.

(Also, I wouldn't object to the body of these functions being moved to the constructor of `QuadratureRule`, to be honest).

michelebucelli (commented):
Thank @zasexton! I have a few leftover minor comments.

I still think it would be useful if you could provide a prototypical quadrature rule implementation, to give an idea about how you envision this in the longer run.

Code/CMake/SimVascularExternals.cmake
zasexton: I think this is a combination of my changes and the existing paths... I think the mismatch with `make doc` might have already been around, but yes this should definitely be addressed. I'll make sure that we are generating within the already built folder

Code/Source/solver/FE/Quadrature/QuadratureRule.h
zasexton: these have been aligned in a4e14cc3

Code/Source/solver/FE/Quadrature/QuadratureRule.h
zasexton: phrasing updated in this commit 5ee42cb3

zasexton: added the documentation about 261e66d4 constraint verification

Code/Source/solver/FE/Quadrature/QuadratureRule.h
zasexton: these should already be implicitly documented with the `@pre` but I'll take another look

Code/Source/solver/FE/Quadrature/QuadratureRule.cpp
zasexton: I'll add the exception infrastructure here

Code/Source/solver/FE/Quadrature/QuadratureRule.cpp
zasexton: I think I'd still prefer for these reference measures to be in the Quadrature module since these measures should be specific to quadrature rules even though they relate to specific cell typologies. Though I'll make a separate note that we might change this in the future.

Code/Source/solver/FE/Quadrature/QuadratureRule.cpp
zasexton: have made this change in commit 7c74acf2

Code/Source/solver/FE/Quadrature/QuadratureRule.cpp
zasexton: made this change in commit ec189180

Code/Source/solver/FE/Quadrature/QuadratureRule.cpp
zasexton: this is implicitly handled now via removing the assertions in commit 8c6735a8

Code/Source/solver/FE/Quadrature/QuadratureRule.cpp
zasexton: moved the body of these functions into the `QuadratureRule` and made the interfaces symmetric in commit f41839c7

zasexton:
> Thank @zasexton! I have a few leftover minor comments.
> 
> I still think it would be useful if you could provide a prototypical quadrature rule implementation, to give an idea about how you envision this in the longer run.

I've made a very simple example in the `test_QuadratureRule.cpp` file for a two-point Gauss rule just to illustrate how this might be used; however the more concrete examples will be the Gauss and Position-based rules.

Code/CMake/SimVascularExternals.cmake
michelebucelli: Thanks for the clarification, and for the updates! I confirm that now running `make doc` from `svMultiPhysics-build` generates documentation into `svMultiPhysics-build/Doxygen`.

As a final (minor, and a bit pedantic) suggestion, I would recommend placing the generated Doxyfile inside `${SV_DOCS_DIR}` instead of `${SV_BINARY_DIR}`.

Should the GitHub actions scripts be also updated following this change?

tests/unitTests/FE/Quadrature/test_QuadratureRules.cpp
```diff
@@ -41,6 +41,24 @@ class TestQuadratureRule final : public QuadratureRule {
     }
 };
 
+class TwoPointGaussLegendreRule final : public QuadratureRule {
+public:
+    TwoPointGaussLegendreRule()
+        : QuadratureRule(
+              svmp::CellFamily::Line,
+              3,
```
michelebucelli: I suggest the following, for enhanced readability:
```cpp
             /* polynomial_exactness = */ 3,
```

michelebucelli:
> I've made a very simple example in the `test_QuadratureRule.cpp` file for a two-point Gauss rule just to illustrate how this might be used; however the more concrete examples will be the Gauss and Position-based rules.

@zasexton Sorry, I'd missed that. Now I looked at it, and it seems to be clear and lightweight enough. Thanks!

ktbolt (approved):
Let's merge it.

michelebucelli (approved):

Code/CMake/SimVascularExternals.cmake
zasexton: Will add this to a small PR for changing the documentation building since this will affect the github action

Issue #594, opened 2026-07-28
Add RDQ20-MF active stress model


samibismar:
### Problem

svMultiPhysics currently does not include the RDQ20-MF mean-field active mechanics model described by Regazzoni, Dedè, and Quarteroni (2020).

Adding this model would allow simulations using a biophysically detailed sarcomere-level active stress formulation that captures calcium activation, length-dependent activation, and force–velocity behavior.

### Solution

Add RDQ20-MF as a new active stress model within the ActiveStress framework introduced in PR #578.

The implementation should include:
- the 20-state RDQ20-MF mean-field dynamics,
- model-specific parameter handling through the XML system,
- integration with the active stress model factory,
- MPI parameter distribution,
- active tension calculation,
- and an electromechanics regression test case.

The model formulation is based on:

Regazzoni, F., Dede', L., & Quarteroni, A.  
"Biophysically detailed mathematical models of multiscale cardiac active mechanics."  
*PLOS Computational Biology*, 2020.  
https://doi.org/10.1371/journal.pcbi.1008294

### Additional context

The implementation was developed using the authors' reference implementation as a comparison source:

https://github.com/FrancescoRegazzoni/cardiac-activation

Validation was performed by comparing RDQ20-MF state evolution and active tension against the reference implementation across multiple calcium, sarcomere length, and velocity protocols.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct and Contributing Guidelines

samibismar:
Implemented the RDQ20-MF active stress model on the following [branch](https://github.com/samibismar/svMultiPhysics/tree/feature-rdq20-mf):

The implementation includes:
- a new `RDQ20-MF` active stress model following the `ActiveStress` framework,
- the 20-state mean-field dynamics (16 regulatory-unit states + 4 crossbridge moments),
- XML parameter integration and MPI parameter distribution,
- active tension computation,
- and a slab electromechanics regression test case.

The model was implemented based on the formulation described in Regazzoni et al. (2020) and compared against the authors' C++ reference implementation.

Validation performed so far:
- RDQ20-MF state evolution and active tension compared against the reference implementation across calcium, sarcomere length, and velocity protocols.
- Long-horizon comparisons of all 20 states showed agreement at machine precision.
- Additional force–calcium, length-dependent activation, and force–velocity checks reproduce the expected model behavior.

[...]

PR #596, opened 2026-07-31
Add Regazzoni active stress model

samibismar:
## Current situation

This PR adds the `RegazzoniActiveStress` model, implementing the RDQ20-MF formulation described by Regazzoni, Dedè, and Quarteroni (2020), using the ActiveStress framework introduced in PR #578.

The implementation includes:

- the 20-state RDQ20-MF system, consisting of 16 regulatory-unit probabilities and 4 crossbridge moments;
- model-specific XML parameter parsing and MPI distribution;
- registration with the `ActiveStress` model factory;
- calcium-, sarcomere-length-, and velocity-dependent state evolution;
- active-tension computation;
- and an electromechanics regression test.

The local active-stress interface is extended to provide fiber stretch to each model. RDQ20-MF first computes the paper tension

$$
\widetilde T_{\mathrm{act}} = a_{\mathrm{XB}} \left(\mu_P^1+\mu_N^1\right)\phi(SL),
$$

then returns

$$
T_{\mathrm{code}} = \frac{\widetilde T_{\mathrm{act}}}{\lambda_f}.
$$

This normalization is required because svMultiPhysics assembles the returned coefficient as a second Piola–Kirchhoff stress contribution:

$$
\mathbf S_{\mathrm{act}} = T_{\mathrm{code}}\ \mathbf f_0\otimes\mathbf f_0.
$$

After conversion to first Piola–Kirchhoff stress, this gives

$$\mathbf P_{\mathrm{act}} = \widetilde T_{\mathrm{act}}\ \mathbf f\otimes\mathbf f_0,
$$

consistent with the RDQ20 paper convention.

Related to #594.

## Release Notes

- Add the RDQ20-MF active-stress model.
- Add RDQ20-MF XML parameters and MPI parameter distribution.
- Extend the local active-stress interface to receive fiber stretch.
- Normalize the RDQ20-MF returned coefficient for the existing PK2 mechanics assembly.
- Add the `slab_RDQ20MF` electromechanics regression case.

## Documentation

The RDQ20-MF class documentation describes:

- the model states and parameters
- the distinction between the paper tension $\widetilde T_{\mathrm{act}}$ and the returned PK2 coefficient $T_{\mathrm{code}}$
- the fiber-stretch normalization
- and the exception raised for zero or near-zero fiber stretch

The `slab_RDQ20MF` README documents:

- the coupled electromechanics setup
- the svMultiPhysics fiber/sheet/sheet-normal directional weighting
- the meaning of the stored `Active_tension_*` fields
- and the reference-validation procedure

## Testing

- The project build completed successfully.
- The `slab_RDQ20MF` pytest case passed for all three configured process-count variants.
- The RDQ20-MF state evolution and paper tension were validated against the authors' C++ reference implementation at commit [`26f05df`](https://github.com/FrancescoRegazzoni/cardiac-activation/commit/26f05df28891df7b3c69f16bb136cdced6b63c4d).
- After normalization, the reconstructed paper tension

 $$
\widetilde{T}_{\mathrm{act}} = \lambda_f T_{\mathrm{code}}
$$

was compared node-by-node against the same reference implementation and agreed to approximately machine precision.

- Additional checks at $\lambda_f = 0.8$, $1.0$, and $1.2$ confirmed

$$
\lambda_f T_{\mathrm{code}} = \widetilde{T}_{\mathrm{act}}
$$

  to floating-point precision.
- A local full-suite run completed with 206 tests passing and 27 skipped. The skipped tests require optional PETSc or Trilinos build configurations.
- Twelve sv0D coupling tests could not run because the local `svZeroDSolver` shared library was unavailable. The same twelve tests fail with the same missing-library error on `main` in the same environment.

## References

Regazzoni, F., Dede', L., & Quarteroni, A.  
“Biophysically detailed mathematical models of multiscale cardiac active mechanics.”  
*PLOS Computational Biology*, 2020.  
https://doi.org/10.1371/journal.pcbi.1008294


ktbolt:
@samibismar May I suggest using a more meaningful name than `RDQ20MF` ? Maybe just `MeanField` perhaps qualified by what it actually represents rather than authors and publication date ?

samibismar:
@ktbolt Would `MeanFieldActiveStress` be an appropriate class name, with `MeanField` as the factory label used in the XML input, for example `<Model>MeanField</Model>`?

I would retain “RDQ20-MF” in the documentation and citations to identify the specific Regazzoni–Dedè–Quarteroni formulation being implemented.

ktbolt:
@samibismar I think `MeanFieldActiveStress` is fine and just document it what the implementation is based on.

michelebucelli:
@samibismar @ktbolt Sorry to comment on this this late, but: I'm not sure that `MeanFieldActiveStress` is a descriptive name for the model. Mean-field based models are more a general category than a description of this specific model, and the fact that the model is based on the mean-field formalism is, I think, a relatively secondary aspect of it.

As far as I know, the standard way of referring to these cellular models is by the name of the author (and sometimes year, if the same authors have published multiple models), since (I think) it is often impossible to come up with a unique descriptive name. This is the same as with ionic models (and indeed we have the classes `AlievPanfilov`, `BuenoOrovio` and so on).

If you feel the acronym `RDQ` is a bit too obscure, I wouldn't mind using `Regazzoni` instead (or `RegazzoniActiveStress`).

samibismar:
@michelebucelli That makes sense, thanks. I agree that `MeanFieldActiveStress` is probably too broad if mean-field models are a general category rather than a unique description of this formulation.

I think `RegazzoniActiveStress` for the C++ class, with `Regazzoni` as the model/XML label, would fit the existing naming pattern while remaining clear in the active-stress code. I would continue to identify it as the RDQ20-MF formulation of Regazzoni, Dedè, and Quarteroni (2020) in the documentation and validation notes.

@ktbolt, would that naming approach work for you as well?

Code/Source/solver/active_stress.h
michelebucelli, line 234: I think that the documentation of `ActiveStressODE` currently reports the formula $T_{act} = T_{act}(\mathbf{s})$, which should be updated to $T_{act} = T_{act}(\mathbf{s}, \lambda)$ (please use the macros from `Documentation/macros.tex` to ensure consistency of notation).

Code/Source/solver/active_stress_regazzoni.h
```diff
@@ -0,0 +1,327 @@
...
+#ifndef ACTIVE_STRESS_MEAN_FIELD_H
+#define ACTIVE_STRESS_MEAN_FIELD_H
+
+#include "active_stress.h"
+
+/**
+ * @brief Mean-field active stress model (implements the RDQ20-MF formulation).
+ *
+ * This class implements the mean-field RDQ20-MF sarcomere model of cardiomyocyte
+ * force generation of Regazzoni, Dedè, and Quarteroni (2020), described in [1]
```
michelebucelli: The correct spelling is Dede', although Google Scholar and some journals get it wrong every now and then (please double-check also elsewhere).

Code/Source/solver/active_stress_regazzoni.h
```diff
@@ -0,0 +1,327 @@
...
+ * and validated against the authors' reference implementation [2]. The node-local state has 20 variables: 16
+ * regulatory-unit (RU) probabilities (entries 0-15) describing the
+ * tropomyosin/troponin configuration of a triplet of neighbouring units, and 4
+ * crossbridge (XB) moments (entries 16-19). The RU probabilities are advanced
+ * with an explicit forward-Euler substepping scheme and the XB moments with one
```
michelebucelli: I suggest expanding a little on the meaning of "substepping" here (i.e. explaining that for every "macro" time step, a certain number of smaller time steps are taken to update the RU states).

Code/Source/solver/active_stress_regazzoni.h
```diff
@@ -0,0 +1,327 @@
...
+   *
+   * Each argument is 0 or 1 and denotes the state of, respectively, the left
+   * tropomyosin unit, the central tropomyosin unit, the right tropomyosin unit
+   * and the central troponin (calcium unbound/bound). The ordering matches the
+   * reference implementation's serialization (TL outermost, CC innermost) and
+   * spans [0, 15].
+   */
+  static constexpr int ru_index(int TL, int TC, int TR, int CC) {
+    return 8 * TL + 4 * TC + 2 * TR + CC;
+  }
```
michelebucelli: Since these are indices, I would use `unsigned int` for all of them (both the inputs and the output).

Code/Source/solver/active_stress_regazzoni.h
```diff
@@ -0,0 +1,327 @@
...
+
+  /// Flat index of the XB moment state @p i (in [0, 3]), spanning [16, 19].
+  static constexpr int xb_index(int i) {
+    return static_cast<int>(n_ru_states) + i;
+  }
```
michelebucelli: Same as above.

Code/Source/solver/active_stress_regazzoni.h
```diff
@@ -0,0 +1,327 @@
...
+ * fraction of the sarcomere at sarcomere length @f$SL = SL_0 \, \lambda@f$ (with
+ * @f$\lambda@f$ the fiber stretch), and @f$a_\text{XB}@f$ is the tension
+ * upscaling factor. Because @f$\mu_P^1 + \mu_N^1@f$ and @f$\phi(SL)@f$ are
+ * dimensionless, @f$a_\text{XB}@f$ sets the units of the returned active tension;
+ * it is stored in the stress units of the simulation, so no separate output
+ * conversion is applied.
+ *
+ * The model is calibrated in a fixed unit system (calcium in [uM], time in [s],
+ * length in [um]); the svMultiPhysics inputs are converted to these units at the
+ * interface (see the conversion members below).
```
michelebucelli: My understanding is that `svMultiPhysics` leaves it to the user to choose a system of units and use it consistently. Accordingly, I do not think that this class should make any assumption regarding the units of measure, nor be concerned with any conversion between systems of units of measure.

In other words, if the ionic model is configured to return calcium in mM, then the user should also configure this class so that it accepts calcium in mM, by converting the relevant parameters accordingly.

If you introduced these conversions so that the default values of parameters are consistent with the published ones, I think we can safely represent the same values with different units, as long as this is documented.

Related to this, I found in a few places comments along the lines of "no separate output conversion is applied" for $a_{XB}$. I suggest removing those comments.

Similarly, some of the documentation comments below include units of measure. However, those units of measure are not actually binding, since the user is free to choose whatever system they prefer. I suggest removing those units from the documentation. It can still be useful to document what dimensions the value is expected to have (e.g. `[1/time]` instead of `[1/s]`).

Code/Source/solver/active_stress_regazzoni.h
```diff
@@ -0,0 +1,327 @@
...
+   * body-temperature calibration of the reference implementation, expressed in
+   * the units documented for each member. The registered value of @c a_XB,
+   * 22.894, expresses the reference calibration in MPa; the value supplied in
+   * solver.xml must instead use the stress unit of the simulation's mechanical
+   * configuration.
+   *
+   * All parameters are required. A complete @c MeanFieldActiveStress parameter block
+   * containing every parameter must be provided in solver.xml. Any value may be changed to
+   * use a different calibration, but omitting a parameter causes a parse error;
+   * the registered reference value is not used as an automatic default.
```
michelebucelli: I suggest making this documentation more concise, e.g.

```suggestion
   * Declares the parameters required by the model. All parameters are
   * marked as required, and omitting a parameter will cause a parse error.
```

Also, you should not assume that the person reading these documentation comments has access to the source code of the implementation, since this comment should be read from the Doxygen-generated website, and it should describe the external behavior of the class without requiring the reader to go into the source code. For example, "the values registered below" would have no meaning in the context of the Doxygen website, since the implementation of the constructor would not be on that same page.

Code/Source/solver/active_stress_regazzoni.h
```diff
@@ -0,0 +1,327 @@
...
+   * from the XB first moments (state entries 17 and 19) and the single-overlap
+   * fraction at @f$SL = SL_0 \lambda_f@f$, then returns
+   * @f$\widetilde{T}_\text{act} / \lambda_f@f$.
+   *
+   * The division by @f$\lambda_f@f$ is necessary because svMultiPhysics
+   * assembles the active stress as
+   * @f$\mathbf{S}_\text{act} = T_\text{code}\,\mathbf{f}_0\otimes\mathbf{f}_0@f$
+   * (second-PK), so the resulting first-PK active stress satisfies
+   * @f$\|\mathbf{P}_\text{act}\| = \widetilde{T}_\text{act}@f$, matching the
+   * paper's definition of active tension.
```
michelebucelli: This information should be moved to the general class documentation.

Also, I recommend using the LaTeX macros defined in `Documentation/macros.tex` to ensure consistency of notation across classes. Feel free to add new macros as appropriate.

Code/Source/solver/active_stress_regazzoni.h
```diff
@@ -0,0 +1,327 @@
...
+   *
+   * The returned value has the stress units of @f$a_\text{XB}@f$.
+   *
+   * @note The slab regression test uses a (0.7, 0.2, 0.1) fiber/sheet/normal
+   *   directional distribution, which is an svMultiPhysics extension of the
+   *   paper's fiber-only active stress formulation.
```
michelebucelli: This information is not related to this class, but rather it refers to one specific test and, more broadly, to the conversion from a scalar active tension to the active stress tensor, which is not something this class is or should be concerned with. Accordingly, I do not think it should be part of this class' documentation.

Code/Source/solver/active_stress_regazzoni.h
```diff
@@ -0,0 +1,327 @@
...
+  /// Calcium conversion, millimolar [mM] to micromolar [uM].
+  static constexpr double calcium_mM_to_microM = 1.0e3;
+
+  /// Time conversion, milliseconds [ms] to seconds [s].
+  static constexpr double time_ms_to_s = 1.0e-3;
+
+  /// @}
+
+  /// RU forward-Euler substep size [s].
+  static constexpr double ru_substep = 2.5e-5;
```
michelebucelli: This should not be a fixed value, but rather a user-defined parameter, not only but also because the units might be different. For example, if the user had set up the simulation in `ms` (as is the standard for electrophysiology simulations), this number would need to change.

Code/Source/solver/active_stress_regazzoni.h
```diff
@@ -0,0 +1,327 @@
...
+
+  /// Fixed reference sarcomere length [um] in the length-dependent dissociation
+  /// constant (distinct from the parameter SL0).
+  static constexpr double kd_reference_sarcomere_length = 2.15;
```
michelebucelli: Similar comment as for `ru_substep`.

tests/cases/electromechanics/slab_Regazzoni/README.md
michelebucelli, line 1: Rather than having an entirely separate test case, it should be possible to have a single test folder with multiple XMLs corresponding to multiple test configurations.   This is the case of `tests/cases/cep/niederer_benchmark_ECGs`, for example.

This way, we wouldn't need to replicate the mesh files, and I think the organization of tests would be somewhat cleaner.

Code/Source/solver/active_stress_regazzoni.cpp
```diff
@@ -0,0 +1,321 @@
...
+    const double fiber_stretch, const double fiber_stretch_rate,
+    Vector<double> &state) const {
+  // Convert the svMultiPhysics electromechanics inputs to the reference units.
+  const double calcium_microM = calcium * calcium_mM_to_microM;
+  const double dt_seconds = dt * time_ms_to_s;
+  const double sarcomere_length = SL0 * fiber_stretch;
+
+  // Calcium/stretch-independent central-tropomyosin transition rates.
+  double rates_T[2][2][2][2];
+  ru_transition_rates_tropomyosin(rates_T);
```
michelebucelli: I think that it would be better to use `std::array` rather than C-style arrays, for a number of reasons:
- they are easier to pass around: for example, you could, and should, have `ru_transition_rates_tropomyosin` return an `std::array`, instead of taking an output argument in input;
- they are aware of their own size through the `size` method;
- they can be iterated through with range-based `for` loops, instead of only index-based loops.

Not all of these will necessarily apply here, but I still think it's generally a good practice to use `std::array` preferentially.

If you are bothered by `std::array<std::array<std::array<std::array<double, 2>, 2>, 2>, 2>` (and rightly so!) you should introduce a `typedef` in the class definition:
```cpp
using Array4D = std::array<std::array<std::array<std::array<double, 2>, 2>, 2>, 2>;
```

The same suggestion applies to 2D arrays elsewhere.

Code/Source/solver/active_stress_mean_field.h
```diff
@@ -0,0 +1,327 @@
...
+ *
+ * The model is calibrated in a fixed unit system (calcium in [uM], time in [s],
+ * length in [um]); the svMultiPhysics inputs are converted to these units at the
+ * interface (see the conversion members below).
+ *
+ * **References**:
+ * 1. [Regazzoni, Dede', Quarteroni (2020)](https://doi.org/10.1371/journal.pcbi.1008294)
+ * 2. [F. Regazzoni, cardiac-activation reference implementation](https://github.com/FrancescoRegazzoni/cardiac-activation)
+ */
+class MeanFieldActiveStress : public ActiveStress {
```
michelebucelli: I don't really like that this class inherits from `ActiveStress` rather than `ActiveStressODE`, even if it is an active stress model based on solving a system of ODEs. Is it because it implements its own custom time-stepping method, or are there other reasons?

I think we should figure out if and how this can be brought under `ActiveStressODE` (although I can see that this is not exactly trivial). I can see a few ways to do that, and I don't like any of them 😅 
1. Generalize the custom time stepping scheme so that it can be moved to `ActiveStressODE`. This would involve introducing an implicit-explicit scheme, with an interface not unlike that of `IonicModel`, where state variables are partitioned into gating variables (updated implicitly with a closed-form solution) and others (updated with whatever timestepping method the user chose, preferably explicitly). I don't really like the overhead that this splitting brings.
2. Derive from `ActiveStressODE`, but then don't use the time stepping infrastructure it provides, and override it with custom time stepping. This would satisfy the "is-one-of" interpretation of inheritance, but break the interface in a number of other ways (e.g. the function `getf` would make no sense).
3. Implement better explicit time stepping in `ActiveStressODE` (say, 4th order Runge-Kutta), and see if it works well enough for this model that we can ditch the custom time stepping scheme. This option has a bit of uncertainty related to finding out if RK4 is enough or not.
4. If `ActiveStressODE` is only used by `NashPanfilov`, then maybe we can get rid of `ActiveStressODE` and have each model implement their custom timestepping as needed. I think whether this makes sense or not hinges on whether we expect to use `ActiveStressODE` in other cases: @kko27 are you using it for the Land-Niederer model, and if so did you need custom timestepping or was forward Euler enough?

I think among these option 3 (the Runge-Kutta one) would be the best, if it works.

michelebucelli (commented):
Thanks @samibismar! Please see my comments below.

Code/Source/solver/active_stress_mean_field.h
kko27: @michelebucelli I am using the `ActiveStressODE` class for my Land-Niederer implementation (for reference, here is my [branch](https://github.com/kko27/svMultiPhysics/tree/land_niederer_active_stress)) However, I had  some convergence issues with the `FE` time stepping and had to implement `RK4` instead. 

I am in favor of option 3 (and then 4, if RK4 is not good enough)

michelebucelli: Thanks @kko27! I've asked @samibismar to try out the RK4 solver from your branch on the RDQ model, if it works we'll go with option 3.

Code/Source/solver/active_stress_regazzoni.h
samibismar: I removed the internal unit conversions and updated the parameter values/documentation so that the model follows the unit system provided by the user configuration rather than assuming the reference implementation units. I also removed the output-conversion comments and changed the documentation to describe dimensions rather than fixed units.

Code/Source/solver/active_stress_mean_field.h
samibismar: I tested the RK4 approach on the Regazzoni model using both an isolated 20-state comparison and the fully coupled electromechanics slab.

In the isolated comparison, stable RK4 timesteps reproduced the refined current-scheme solution under prescribed calcium and fixed kinematics, which suggests that the unified 20-state RHS is consistent with the current model for those imposed inputs.

In the coupled slab:

dt = 1.0 ms: RK4 failed around 11 ms
dt = 0.5 ms: RK4 failed around 11.5 ms
dt = 0.25 ms: RK4 remained stable for the full simulation

The current Regazzoni implementation remained stable at dt = 1.0 ms. This is consistent with the split scheme resolving the faster RU dynamics with smaller internal steps while advancing the XB moments implicitly.

Both unstable coupled runs failed during the early calcium upstroke, with severely unphysical active tension followed by non-finite mechanics fields. The internal RU and XB states were not exported, so I could not determine which state became invalid first.

The result I found is that the shared RK4 approach works for this model only after substantially reducing the outer timestep. I’ve attached a summary figure. The experimental Regazzoni RK4 implementation is available [here](https://github.com/samibismar/svMultiPhysics/tree/experiment-regazzoni-rk4).

image

ktbolt:
@samibismar Using `RegazzoniActiveStress` for the C++ class is fine and does follow the naming convention for these sort of things.

Code/Source/solver/active_stress_mean_field.h
michelebucelli: @samibismar Thanks for trying this! It seems that the model is too stiff for this solver then.

At this point, I suggest we keep the implementation as it currently is, but leave a comment in the class documentation explaining why it inherits from `ActiveStress` instead of `ActiveStressODE`, along the lines of
> Even if this model solves an ODE system, we do not inherit from ActiveStressODE, because the model requires a customized time stepping scheme to deal with its high stiffness.

michelebucelli:
@samibismar As we discussed, this model needs some kind of stabilization to enable force-strain-rate feedback. I think this can and should be addressed in a follow-up PR, but in the meantime I suggest the following:
1. to add a `@todo` comment in the class documentation to point this out, and
2. to add a boolean parameter (read from XML) to switch off force-strain-rate feedback (setting the parameter to true would replace $d\lambda / dt$ with 0 in the equations).

Having done this, I think this is good to merge!

samibismar:
@michelebucelli I added the `@todo` for the follow-up stabilization work and the `Disable_force_strain_rate_feedback` parameter. When enabled, it replaces $\frac{d\lambda}{dt}$ with 0 in the XB update; the default is `false` so the existing behavior is preserved.

tests/test_electromechanics.py
```diff
@@ -24,5 +22,12 @@
...
-    run_with_reference(base_folder, test_folder, fields, n_proc, t_max=1)
+    run_with_reference(base_folder, "slab", fields, n_proc, t_max=1,
+                       name_inp="solver_NashPanfilov.xml",
+                       name_ref="result_NashPanfilov_001.vtu")
+
+
+def test_slab_regazzoni(n_proc):
+    run_with_reference(base_folder, "slab", fields, n_proc, t_max=1,
+                       name_inp="solver_Regazzoni.xml",
+                       name_ref="result_Regazzoni_001.vtu")
```
javijv4: Not really needed, but to follow the format of other tests, whenever there are two tests in one folder, the way to go about it is using `@pytest.mark.parametrize`. See `test_cep.py` for example.

javijv4 (commented):
Hey @samibismar, this looks good to me! Just left a small comment regarding the test. Not super important though.

javijv4:
@michelebucelli, @samibismar, I don't think this is an issue for this PR, but I was playing around with the test and saw noticeably different results when using 3 or 4 processors when I let it run longer (below is timestep 50). 
**4 proc**
Screenshot from 2026-08-11 19-40-32
**3 proc**
Screenshot from 2026-08-11 19-40-22

I checked using Nash Panfilov, and the same occurs. Then I tried changing GMRES to CG (just because, in my experience, CG works better for EP), and the differences became much smaller. Here are tables quantifying the differences. 
=== GMRES: 3 vs 4 ranks at t=50 (max abs diff) ===
model        | dVm        | dTact      | dCa        | dDisp     
-------------|------------|------------|------------|-----------
Regazzoni    | 5.0823e+00 | 3.1990e-04 | 1.0133e-04 | 8.5070e-04
NashPanfilov | 5.0823e+00 | 1.0954e-02 | 1.0133e-04 | 3.4885e-02

=== CG: 3 vs 4 ranks at t=50 (max abs diff) ===
model        | dVm        | dTact      | dCa        | dDisp     
-------------|------------|------------|------------|-----------
Regazzoni    | 2.4833e-06 | 1.5796e-11 | 8.5102e-11 | 1.8047e-11
NashPanfilov | 2.4833e-06 | 1.5706e-09 | 8.5102e-11 | 1.9004e-09

I do not really know what is causing this. I can open an independent issue if you think that is a good idea.

ktbolt:
@javijv4 Aren't the matrices produced by `CEP` symmetric and positive definite ? If so then you would use CG and not GMRES; same for the `struct` equation.

michelebucelli:
Thanks @javijv4 for running these experiments!

I agree with @ktbolt that the `CEP` matrices are symmetric and positive definite, and so CG should be the best choice to solve them. It might be a good idea to switch the two electromechanics tests to use CG by default for electrophysiology.

However, for nonlinear hyperelasticity the matrix is not symmetric in general, I think (I'm not sure whether this holds fror the generalized $\alpha$ method, but the stiffness matrix at least is generally not symmetric). Therefore, `struct` should still require GMRES.

Having said that, GMRES should still work, and should work consistently when varying the number of cores. So I agree with @javijv4 that it is quite suspicious that the results change so much between 3 and 4 processes. I think this might be another hint that something, somewhere is very sensitive to the order of operations. I have encountered other clues to this in #584 (fast math optimization broke all the tests) and #577 (using `x += ...` in place of `x = x + ...` broke some tests). When prompted about this, Claude indicated that this might be related to the GMRES implementation, but I haven't investigated this claim so take it with a very big grain of salt.

@javijv4 if you want to open an issue about this I'll be happy to add more details about the similar problems I've found, and we can investigate further. I suspect this to be independent of the changes in this PR, so I suggest merging this PR regardless.

michelebucelli (approved):

javijv4:
Sounds good! I'll open a new issue.

javijv4 (approved):

samibismar:
@michelebucelli I addressed the final test comment by parametrizing the NashPanfilov and Regazzoni slab cases with `@pytest.mark.parametrize`. All 6 test combinations pass.

PR #605, opened 2026-08-19
Allow 0D solver to be coupled to equations other than the first

### LINKED ISSUE #592: Coupling to 0D solver only supported for first equation
### Description

In a number of places, the infrastructure for coupling `svMultiPhysics` to `svZeroDSolver` assumes that the coupling is happening for the first equation. This means that e.g. in electromechanics simulations (introduced in #578) the structural mechanics equation needs to be the first. If electrophysiology is the first equation, the 0D solver doesn't run, and no error message or warning is printed to make the user aware of this.

One instance of this is
https://github.com/SimVascular/svMultiPhysics/blob/042a1552fb694604adef9ebf6aae87a4bad194bf/Code/Source/solver/initialize.cpp#L416-L422

The flag `com_mod.cplBC.coupled` is set to true only if `com_mod.eq[0].bc` contains some coupled condition. If the coupled condition is in some other equation, then `com_mod.cplBC.coupled` remains `false`, and the 0D solver is silently never called.

A similar behavior is happening in a few other places (I could find similar things in `Integrator` and `set_bc.cpp`, but there might be a few more).

### Reproduction

From the branch of #578, try running an electromechanics simulation coupled to `svZeroDSolver` for the structure boundary conditions. The first equation must be `CEP`, the second `struct`.

### Expected behavior

The coupling with the 0D solver should work independently of the ordering of the equations. The ordering of the equations should affect the order in which they are solved in partitioned schemes, but not the BCs that are applied to them.

Should that be impossible (but I don't see why it should), at least an error message should be raised to notify the user that the 0D solver is not being called despite being specified in the XML file.

### Additional context

_No response_

### Code of Conduct

michelebucelli:
Fixes #592 

## Current situation

The interface to `svZeroDSolver` assumes that the equation it is being coupled to is the first.  If it isn't, any indication of coupling to `svZeroDSolver` in the XML file is silently disregarded.

## Release Notes 

1. `svZeroDSolver` can be coupled to an arbitrary equation. The coupled equation index is tracked in `cplBCType::equation_index`. Coupling to multiple equations was and still is unsupported.
2. Several integral computation functions have been modified to select the right variable to integrate based on the equation index. These include `CoupledBoundaryCondition::compute_flowrates`, CoupledBoundaryCondition::compute_pressures`, `gnnb` and `integ`.
3. The interface to `gnnb` and `integ` has been simplified to remove redundant arguments (e.g. separately passing an `Array` and its size).
4. Perturbation magnitudes used to evaluate the derivative of the 0D solver by finite differences were hardcoded. Now they can be specified through the XML file in the section `<svZeroDSolver_interface>` (especially important for the absolute tolerance, since it is dimensional).
5. `svZeroDSolver` output files (`svZeroD_data`, `P_svZeroD` and `Q_svZeroD`) are written to the same output directory as the 3D simulation results, instead of the current directory.
6. `svZeroD_data` is reset by every simulation, instead of appended to (i.e, the file is cleared before writing the header).
7. Classes under `ActiveStress` and `IonicModel`, and their files, have been renamed to more consistently use upper camel case.

## Documentation

Modified or added data members and functions have been given Doxygen documentation.

## Testing

I have verified that the automatic tests still pass.

The coupling of the 0D solver to an equation other than the first was tested in a ventricular electromechanics simulation. This simulation should eventually become an automatic test. However, the choice of various model parameters is still very much work in progress, and I would postpone making it an automatic test until it gives plausible results, so that the test can be a reasonable starting point for electromechanics setups.


Code/Source/solver/set_bc.cpp
```diff
@@ -635,8 +647,11 @@ void rcr_init(ComMod& com_mod, const CmMod& cm_mod, const SolutionStates& soluti
       if (cplBC.initRCR) {
         auto& fa = com_mod.msh[iM].fa[iFa];
         double area = fa.area;
-        double Qo = all_fun::integ(com_mod, cm_mod, fa, Yo, 0, solutions, nsd-1, false, MechanicalConfigurationType::reference);
-        double Po = all_fun::integ(com_mod, cm_mod, fa, Yo, nsd, solutions, std::nullopt, false, MechanicalConfigurationType::reference) / area;
+        double Qo = all_fun::integ(com_mod, cm_mod, fa, Yo, eq.s, solutions,
+                                   nsd - 1, false);
```
javijv4: Following the other flow calculations, should this be using `eq.s + nsd -1`?

javijv4 (commented):
Hey @michelebucelli, sorry for the late review. I went through all the files and it looks good to me. Just the small detail I commented on.

Code/Source/solver/set_bc.cpp
michelebucelli: Yes, thanks for catching this! I just pushed a commit that fixes this.

javijv4 (approved):

aabrown100-git (approved):
Looks good to me, thanks!

Issue #602, opened 2026-08-18
Move material stiffness mat_mul outside innner loop


dseyler:
### Problem

Currently, the mat_mul operation to compute material stiffness for both struct and ustruct is contained inside an inner loop over each element's nodes. The mat_mul operation can be safely hoisted ouside the inner loop, resulting in a **~10% speedup timestep** for a 500k element simulation.

This occurs around line 750 in sv_struct.cpp:
```
  for (int b = 0; b < eNoN; b++) {

    for (int a = 0; a < eNoN; a++) {

      // Geometric stiffness
      NxSNx = Nx(0,a)*S(0,0)*Nx(0,b) + Nx(1,a)*S(1,0)*Nx(0,b) +
              Nx(2,a)*S(2,0)*Nx(0,b) + Nx(0,a)*S(0,1)*Nx(1,b) +
              Nx(1,a)*S(1,1)*Nx(1,b) + Nx(2,a)*S(2,1)*Nx(1,b) +
              Nx(0,a)*S(0,2)*Nx(2,b) + Nx(1,a)*S(1,2)*Nx(2,b) +
              Nx(2,a)*S(2,2)*Nx(2,b);

      T1 = amd*N(a)*N(b) + afu*NxSNx;

      // Material Stiffness (Bt*D*B)
      mat_mul(Dm, Bm.rslice(b), DBm);
```

A similar pattern occurs around line 1410 in ustruct.cpp:

```
for (int b = 0; b < eNoNw; b++) {
    for (int a = 0; a < eNoNw; a++) {
      NxSNx = Nwx(0,a)*Siso(0,0)*Nwx(0,b)
       + Nwx(0,a)*Siso(0,1)*Nwx(1,b) + Nwx(0,a)*Siso(0,2)*Nwx(2,b)
       + Nwx(1,a)*Siso(1,0)*Nwx(0,b) + Nwx(1,a)*Siso(1,1)*Nwx(1,b)
       + Nwx(1,a)*Siso(1,2)*Nwx(2,b) + Nwx(2,a)*Siso(2,0)*Nwx(0,b)
       + Nwx(2,a)*Siso(2,1)*Nwx(1,b) + Nwx(2,a)*Siso(2,2)*Nwx(2,b);

      auto DBm = mat_mul(Dm, Bm.rslice(b));
```

### Solution

[...]

PR #608, opened 2026-08-19
Move unnecessary mat_mul outside loops and implemented fixed-sized mat_mul

dseyler:
## Current situation
After profiling a 500k element structural mechanics simulation, I noticed several mat_mul function calls that were unnecessarily repeated within loops, as discussed in [issue 602](https://github.com/SimVascular/svMultiPhysics/issues/602).

These occurred at:

- mat_models.cpp lines 1695-1696
- sv_struct.cpp line 750
- ustruct.cpp line 144

and have now been **moved outside their loops**.

Additionally, while investigating, it was noticed that many of these mat_mul operations were of fixed-size arrays with sizes known at compilation. These can be optimized with fixed-size Eigen maps, which run **3-4x faster** than the generic mat_mul operation.

New fixed-size functions were defined in mat_mul.cpp:
- mat_mul_fixed()
- mat_mul_fixed_rows()

These functions are then **dispatched within mat_mul()** so developers do not need to distinguish between these differences.

```/// @brief Fixed-shape matrix product, C = A*B, mapped onto the existing buffers.
///
/// Used when sizes are known at compile time.
template <int M, int K, int N>
inline void mat_mul_fixed(const Array<double>& A, const Array<double>& B, Array<double>& C)
{
  Eigen::Map<const Eigen::Matrix<double, M, K>> a(A.data());
  Eigen::Map<const Eigen::Matrix<double, K, N>> b(B.data());
  Eigen::Map<Eigen::Matrix<double, M, N>>       c(C.data());

  c.noalias() = a * b;
}
```
And in mat_mul:

```void mat_mul(const Array<double>& A, const Array<double>& B, Array<double>& result)
{ 
  // Fixed-shape fast paths for the products that dominate the element loops.
  //
  //   3x3 * 3x3     F*S in struct_3d; vx*Fi, ddev*Fit, Fi*ddev_Fit and the
  //                 potential-viscosity products in mat_models; F^T*F in cep
  //   3x3 * 3xeNoN  ddev*Nx_Fi and vx_Fi*Nx_Fi in the viscous tangent
  //   6x6 * 6x3     the material stiffness product D*B in struct_3d/ustruct_3d_m
  if (A.nrows() == 3 && A.ncols() == 3 && B.nrows() == 3 &&
      result.nrows() == 3 && result.ncols() == B.ncols()) {
    if (B.ncols() == 3) {
      mat_mul_fixed<3, 3, 3>(A, B, result);
    } else {
      mat_mul_fixed_rows<3, 3>(A, B, result);
    }
    return;
  }
```

In total, these changes **decreased runtime by ~21%.**

hoist_branch

I suspect this number could increase much more if we correctly implement eigen arrays throughout sv_struct.cpp and mat_models.cpp as many arrays are of size nsd or enon, which have a discrete number of values and can be optimized similarly. I tested this on the viscosity models alone and found **another 10% improvement** (can open an issue about this, but the code is too messy for this PR)

## Release Notes 

- Moved mat_mul outside loops in several locations
- Implemented fixed-shape mat_mul functions
- mat_mul directs arrays with know sizes at compilation down these paths


Code/Source/solver/mat_fun.cpp
```diff
@@ -507,15 +507,9 @@ mat_mul(const Array<double>& A, const Array<double>& B)
...
-    for (int j = 0; j < B_num_cols; j++) {
-      double sum = 0.0;
-      for (int k = 0; k < A_num_cols; k++) {
-        sum += A(i,k) * B(k,j);
-      }
-    result(i,j) = sum;
-    }
-  }
+  // Delegate rather than repeat the loop, so callers of this form reach the
+  // fixed-shape fast paths too. Most call sites use this overload.
```
michelebucelli: Minor: I think that, for a developer that is unaware of the history of this PR, this comment might not be understandable (e.g., what loop is it talking about, since there is no loop above?).

I suggest rephrasing to something like:
```suggestion
  // Fall back onto the overload of this function that takes the result matrix
  // as an output argument.
```

Code/Source/solver/mat_fun.cpp
michelebucelli, line 494: I would suggest taking the opportunity to update the documentation of the functions being modified here, including details such as the meaning of inputs and outputs (which is fairly obvious here, but still), and whether the functions perform checks on their inputs.

In this case, this might look something like this:
```suggestion
/// @brief Multiply two matrices.
///
/// Given the input matrices @f$A@f$ and @f$B@f$, computes and returns
/// their product @f$AB@f$.
///
/// This function checks that the matrix dimensions are compatible, and
/// throws an std::runtime_error exception otherwise.
///
/// Reproduces Fortran MATMUL.
```

1. I don't know if it's appropriate to keep the Fortran routine reference (it is probably only meaningful to those who know the original Fortran codebase, but it might still be useful during the transition).
2. If the runtime error is replaced as I suggest below, the comment would have to be updated.

Additionally, I suggest moving this documentation comment (and maybe others for other modified functions in this file) to the header file, rather than the source file. As far as I know, that is the standard place to put documentation, the logic being that someone who wants to use the function will first look it up in the header file (so that they can see the signature), and only look at the `cpp` file if they really need to know the implementation details (which they shouldn't care about in most normal situations).

michelebucelli: I suggest using one of the exceptions from `Core/Exception.h` or `FE/Common/FEException.h`. I think that `InvalidArgumentException` is probably appropriate here.

I suggest to look up the function `throw_if` and use that to bundle the check and the exception raising in a single statement. The syntax of that function is somewhat obscure, but if you look up some places where it is called it should become clear.

Code/Source/solver/mat_fun.cpp
```diff
@@ -524,8 +518,62 @@ mat_mul(const Array<double>& A, const Array<double>& B)
...
+  Eigen::Map<const Eigen::Matrix<double, M, K>> a(A.data());
+  Eigen::Map<const Eigen::Matrix<double, K, Eigen::Dynamic>> b(B.data(), K, B.ncols());
+  Eigen::Map<Eigen::Matrix<double, M, Eigen::Dynamic>>       c(C.data(), M, C.ncols());
+
+  c.noalias() = a * b;
+}
+
+} // namespace
+
 void mat_mul(const Array<double>& A, const Array<double>& B, Array<double>& result)
```
michelebucelli, line 523: I suggest giving a documentation to this function, along the lines of what I suggested above. Since this function has both input and output arguments, I suggest using the Doxygen syntax `@param[in]` and `@param[out]` to document them individually, making this explicit (as opposed to, implied by their types).

Code/Source/solver/mat_fun.cpp
```diff
@@ -524,8 +518,62 @@ mat_mul(const Array<double>& A, const Array<double>& B)
 ///
 /// Compute result directly into the passed argument.
 //
+namespace {
+
+/// @brief Fixed-shape matrix product, C = A*B, mapped onto the existing buffers.
+///
+/// Used when sizes are known at compile time.
```
michelebucelli: In the same vein as above, I suggest to expand the documentation here. Here it is especially important do document what the template parameters map to (I assume they're the dimensions of the matrices, but which of the dimensions each parameter corresponds to is left implicit). I believe that Doxygen has the syntax `@tparam` for template parameters.

(Since this function is only defined in the `cpp` file, I think it's appropriate that the documentation remain here).

Code/Source/solver/mat_fun.cpp
```diff
@@ -524,8 +518,62 @@ mat_mul(const Array<double>& A, const Array<double>& B)
...
+  Eigen::Map<Eigen::Matrix<double, M, N>>       c(C.data());
+
+  c.noalias() = a * b;
+}
+
+/// @brief As mat_mul_fixed, but with the column count known only at run time.
+///
+/// Used where the right operand has one column per element node, so its width
+/// depends on the element type. The row counts are still compile-time, which is
+/// where most of the benefit comes from.
```
michelebucelli: Same as above about documenting the meaning of template parameters.

Code/Source/solver/mat_fun.cpp
michelebucelli, line 523: I think it would be cleaner to offer only one function between this (I'll call it option A below) and the overload that returns its result (which I'll call option B). I assume that both are currently used in the code, but a search-and-replace should quickly fix that if we resolve to only leave one.

The pros and cons of the two options, as far as I can tell, are as follows:
1. option A is faster than or equal to than option B, but this can probably be at least alleviated by designing option B and the `Array` class so that return value optimization takes place and move construction/move copy are possible;
2. option A has more cumbersome (old style idiom, less intuitive) syntax, and requires the caller to instantiate the target object (potentially two lines instead of one); option B supports the more modern syntax `C = mat_mul(A, B)`, which has much better readability;
3. option B supports assigning the result to `const` outputs, which in some cases might allow the compiler to better optimize things out of loops when possible. In other words, with option B you will be able to do `const auto C = mat_mul(A, B)`, while option A requires `Array<double> C(...); mat_mul(A, B, C);`, so that `C` cannot be made `const`.

Personally, I'd lean in favor of going with option A but implementing move constructor and move assignment for `Array`, so that copies can be made at basically zero cost if the object being copied-from is going to be discarded.

michelebucelli (commented):
Thanks @dseyler! I left a few comments.

Code/Source/solver/mat_fun.cpp
ktbolt: Yes, let's remove the remembrance of things past like the Fortran references. 

Documentation should be in both the header (API documentation) and .cpp (implementation details) files.

Code/Source/solver/mat_fun.cpp
```diff
@@ -524,8 +518,62 @@ mat_mul(const Array<double>& A, const Array<double>& B)
...
 { 
+  // Fixed-shape fast paths for the products that dominate the element loops.
+  //
+  //   3x3 * 3x3     F*S in struct_3d; vx*Fi, ddev*Fit, Fi*ddev_Fit and the
+  //                 potential-viscosity products in mat_models; F^T*F in cep
+  //   3x3 * 3xeNoN  ddev*Nx_Fi and vx_Fi*Nx_Fi in the viscous tangent
+  //   6x6 * 6x3     the material stiffness product D*B in struct_3d/ustruct_3d_m
+  if (A.nrows() == 3 && A.ncols() == 3 && B.nrows() == 3 &&
+      result.nrows() == 3 && result.ncols() == B.ncols()) {
+    if (B.ncols() == 3) {
```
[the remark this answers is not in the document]
dseyler: @ktbolt Do you agree?

michelebucelli (approved):

ktbolt (approved):
Let's merge it.

PR #617, opened 2026-08-26
Set MueLu number of equations to dof

### LINKED ISSUE #601: Trilinos MueLu default number of equations is incorrect for struct
### Description

In the default Trilinos ML options, the "number of equations" is hardcoded to 4 (Trilinos_impl.cpp, line 767), causing struct simulations to throw an error. Users can currently avoid this by including a custom Trilinos options file, but I think a more permanent default solution would be preferable.

Changing the "number of equations" parameter value to the number of degrees of freedom fixes this problem:
```
mueluParams.set("number of equations", 4);   // dof
```
should be
```
mueluParams.set("number of equations", dof); 
```

This is a simple one-line fix, which I implemented and tested in [this branch](https://github.com/dseyler/svMultiPhysics/tree/fix/muelu-number-of-equations-dof).

I think it would also be a good idea to include a struct Trilinos ML test case to catch this bug (and potentially others) in the future. The branch I linked includes one that is a variant of LV_HolzapfelOgden_active.

### Reproduction

Running a sufficiently large struct simulation (above coarse: max_size = 2000) with the default Trilinos ML settings results in the following error:
```
std::length_error from vector::_M_fill_insert
```

### Expected behavior

Struct simulations should run without error when using the default Trilinos ML settings.

### Additional context

_No response_

### Code of Conduct

dseyler:
Fixes [Issue #601 ](https://github.com/SimVascular/svMultiPhysics/issues/601)
## Current situation

Changed hard-coded `mueluParams.set("number of equations", 4);` to `mueluParams.set("number of equations", dof);` so that Trilinos ML works for all equation types. See [Issue #601 ](https://github.com/SimVascular/svMultiPhysics/issues/601) for a complete discussion of this change.

I also added a Trilinos ML struct test case to catch this bug. Can remove that from this PR if you feel it is unnecessary.

Some open questions remain from our discussion:

- Which Tpetra / MueLu objects can be saved across timesteps? Building these for each timestep is currently a large portion of runtime.
- Simulations with multiple equations (electromechanics) would be particularly exposed to bugs with Trilinos if these objects are saved across timesteps as they are equation-specific. May be worth adding a Trilinos ML electromechanics test case.

## Release Notes 
`mueluParams.set("number of equations", 4)` -> `mueluParams.set("number of equations", dof)`


Code/Source/solver/trilinos_impl.cpp
```diff
@@ -764,7 +764,7 @@ void setMueLuPreconditioner(Teuchos::RCP<MueLu_Preconditioner> &MueLuPrec,
 
   // Problem type
   mueluParams.set("problem: type", "unknown"); // FSI is generally nonsymmetric
-  mueluParams.set("number of equations", 4);   // dof
+  mueluParams.set("number of equations", dof);   //dof for this equation
```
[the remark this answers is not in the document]
dseyler: This was discussed in Issue #601, and I believe determined to be safe

tests/cases/struct/LV_HolzapfelOgden_active_trilinos_ml/solver.xml
```diff
@@ -0,0 +1,132 @@
...
+     <Fiber_direction> true </Fiber_direction>
+   </Output>
+
+   <LS type="GMRES" >
+      <Linear_algebra type="trilinos" >
+         <Preconditioner> trilinos-ml </Preconditioner>
+      </Linear_algebra>
+      <Tolerance> 1e-12 </Tolerance>
+      <Max_iterations> 100 </Max_iterations>
+      <Krylov_space_dimension> 50 </Krylov_space_dimension>
```
michelebucelli: Partially unrelated: I have noticed that, for Trilinos preconditioners, the parameter `Krylov_space_dimension` is disregarded, and the maximum number of iterations is just `Max_iterations`, rather than `Max_iterations * Krylov_space_dimension` as in `fsils`. Is this correct @sujaldave?

While I think that this inconsistency should be removed, I think this would require a larger discussion.

Still, for Trilinos-based test cases, I suggest removing the `Krylov_space_dimension` parameter from the XML file, since it will be unused, and increasing `Max_iterations` e.g. to 500 or 1000, which I think are more sensible values for structural mechanics simulations (not for this coarse test case of course, but for finer simulations that one might base off of this one).

michelebucelli (approved):
Thanks @dseyler! Only a minor suggestion, but otherwise approved.

Code/Source/solver/trilinos_impl.cpp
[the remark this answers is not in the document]
michelebucelli: Safe is perhaps a strong word 😄 but I agree that refactoring this would be better addressed by an object-oriented redesign of the Trilinos interface, which is beyond the scope of this PR.

tests/cases/struct/LV_HolzapfelOgden_active_trilinos_ml/solver.xml
dseyler: Thanks. I'll increase Max_iterations.

As for the Krylov_space_dimension, I think it does affect the max number of restarts. In trilinos_impl.cpp:
`int maxRestarts = int(maxIters / kspace)+1;`

michelebucelli: Oops, my bad for missing it, sorry!

sujaldave (approved):
Looks good to me! Thanks @dseyler .

PR #620, opened 2026-08-26
Implement Darcy Physics

### LINKED ISSUE #71: Tissue perfusion model
***I am moving this issue over verbatim from the Fortran svFSI repository***

I am developing a model for the perfusion of blood within a tissue volume. The model is a single-compartment Darcy model, which can be converted into a Helmholtz equation.

Details of the model in this paper: Papamanolis, L., et al. (2021). Myocardial Perfusion Simulation for Coronary Artery Disease: A Coupled Patient-Specific Multiscale Model. Annals of Biomedical Engineering, 49(5), 1432–1447. https://doi.org/10.1007/s10439-020-02681-z

Details of implementation:

The solver for the Helmholtz equation is being developed as a modification to the HEATS diffusion solver.
The tissue volume is divided into perfusion regions, each with a perfusion pressure. Other important parameters are permeability and conductance parameters.
The nodes corresponding to each perfusion region, and the corresponding perfusion pressure, depend on the vascular network that is perfusing the tissue volume. So this is user-specific, but the interface with svFSI is a list of nodes that specify each territory.
This is generally solved as a steady problem, which can easily be accomplished by setting density = 0.
The idea is to couple this with a flow simulation so that the flow pressures drive the perfusion in the tissue volume. The aim is to couple it with 3D and 0D flow solvers. Currently, the interface between the perfusion model and the flow solver will be a "hand-off" using file I/O (subject to improvement if required).

### LINKED ISSUE #616: Add implementation for darcy flow problem
### Use Case

Incorporating this physics is a necessary step for future perfusion studies, allowing users to model flow outside of vasculature. 

### Problem

Currently for perfusion applications, svMultiPhysics does not have a dedicated solver for modeling flow in the porous solid domain (aside from an old Brinkman implementation). A darcy flow solver addresses this gap in solving a Poisson equation for pressure and computing the resulting velocities. 

### Solution

I have been working on the [darcy-solver branch](https://github.com/mmegally/svMultiPhysics/tree/darcy-solver), which implements this Darcy physics module. This is a more current version of Zack's [perfusion-solver branch](https://github.com/zasexton/svMultiPhysics/tree/perfusion-solver). The solver successfully computes pressure and derives the velocity field for a simple 3D cylinder case configured to have an analytical solution. Additionally, below I show some insights from convergence tests, reflecting the expected O(h^2) pressure and O(h) velocity convergence rates on a cylindrical test case (configured to have an analytical solution). These results can be reproduced in the tests/cases/fluid/darcy_convergence in my branch. 

Image

Ultimately, want to verify before submitting a PR.

### Alternatives considered

Again, this is a more recent version of an old perfusion-solver branch that I had been using for previous perfusion studies. 

### Additional context

Note that the majority of file changes in my branch's commit history are related to the test cases implemented (file additions).

### Code of Conduct

mmegally:
## Current situation
See issue #616 on introducing darcy physics.

## Release Notes 
- Adding new equation type "darcy" to solve for pressure/flows in porous domains
Closes #616, Closes #71 


Code/Source/solver/consts.h
```diff
@@ -287,7 +287,8 @@ enum class EquationType
   phys_CMM = 209, 
   phys_CEP = 210,
   phys_ustruct = 211,  // Nonlinear elastodynamics using mixed VMS-stabilized formulation 
-  phys_stokes = 212
+  phys_stokes = 212,
+  phys_darcy = 213
```
zasexton, line 291: > A Darcy-only problem currently sets `dFlag=true`, causing the predictor to maintain a meaningless `Dn` state and changing the restart record to include displacement data. The suggested changes below address this by adding the conventional `Equation_darcy` alias and including Darcy in the first-order/non-displacement classification. With these changes applied, Darcy is handled consistently with the other first-order scalar equations.

### Suggested code changes

Add the conventional equation alias in `Code/Source/solver/consts.h` at line 294:

```diff
 constexpr auto Equation_CMM = EquationType::phys_CMM;
 constexpr auto Equation_CEP = EquationType::phys_CEP;
+constexpr auto Equation_darcy = EquationType::phys_darcy;
 constexpr auto Equation_fluid = EquationType::phys_fluid;
```

Update the non-displacement classification in `Code/Source/solver/initialize.cpp` at line 443:

```diff
-    if (std::set<EquationType>{Equation_fluid, Equation_heatF, Equation_heatS, Equation_CEP, Equation_stokes}.count(eq.phys) == 0) {
+    if (std::set<EquationType>{
+            Equation_CEP,
+            Equation_darcy,
+            Equation_fluid,
+            Equation_heatF,
+            Equation_heatS,
+            Equation_stokes
+        }.count(eq.phys) == 0) {
       dFlag = true;
     }
```

Result of the suggested change: after equation initialization, a Darcy-only configuration leaves `com_mod.dFlag == false`.

Code/Source/solver/post.cpp
```diff
@@ -1011,6 +1011,21 @@ void post(Simulation* simulation, const mshType& lM, Array<double>& res, const S
...
 
+      // MBF Flux calculation
+      // 
+      } else if (outGrp == OutputNameType::outGrp_mbfFlx) {
+        double kappa = eq.dmn[cDmn].prop[PhysicalProperyType::permeability];
+        int i = eq.s;
+        Vector<double> q(nsd);
+        for (int a = 0; a < eNoN; a++) {
+          for (int j = 0; j < nsd; j++) {
+            q(j) = q(j) + Nx(j, a) * yl(i, a);
```
zasexton: > The current reconstruction reports `MBF_flux` along global x for a fiber mesh because `nn::gnn` only populates `Nx(0,a)` when `insd=1`. The suggested changes below address this by reconstructing the physical gradient as `(dP/ds) * tangent` and then applying the Darcy mobility. This was originally something that I overlooked when initially writing these governing equations.

### Suggested code changes

Assuming one-dimensional Darcy fiber support is intentional, compute the unit physical tangent in `Code/Source/solver/post.cpp` beginning at line 917:

```diff
     Array<double> ksix(nsd,nsd);
     double Jac = 0.0;
+    Vector<double> fiber_tangent(nsd);
 
     for (int g = 0; g < lM.nG; g++) {
       if (g == 0 || !lM.lShpF) {
         auto Nx_g = lM.Nx.slice(g);
         nn::gnn(eNoN, nsd, insd, Nx_g, xl, Nx, Jac, ksix);
+
+        if (lM.lFib) {
+          fiber_tangent = 0.0;
+          for (int a = 0; a < eNoN; a++) {
+            for (int j = 0; j < nsd; j++) {
+              fiber_tangent(j) =
+                  fiber_tangent(j) + xl(j,a) * Nx_g(0,a);
+            }
+          }
+
+          const double tangent_norm = utils::norm(fiber_tangent);
+          if (utils::is_zero(tangent_norm)) {
+            throw std::runtime_error(
+                "[post] Cannot compute Darcy flux for a degenerate fiber element.");
+          }
+          fiber_tangent = fiber_tangent / tangent_norm;
+        }
       }
```

Replace the Darcy flux block beginning at `Code/Source/solver/post.cpp:1016`:

```diff
       } else if (outGrp == OutputNameType::outGrp_mbfFlx) {
-        double kappa = eq.dmn[cDmn].prop[PhysicalProperyType::permeability];
-        int i = eq.s;
-        Vector<double> q(nsd);
-        for (int a = 0; a < eNoN; a++) {
-          for (int j = 0; j < nsd; j++) {
-            q(j) = q(j) + Nx(j, a) * yl(i, a);
+        const double permeability =
+            eq.dmn[cDmn].prop[PhysicalProperyType::permeability];
+        const double viscosity =
+            eq.dmn[cDmn].prop[PhysicalProperyType::darcy_fluid_viscosity];
+        const double mobility = permeability / viscosity;
+        const int i = eq.s;
+
+        Vector<double> grad_p(nsd);
+
+        if (lM.lFib) {
+          double dp_ds = 0.0;
+          for (int a = 0; a < eNoN; a++) {
+            dp_ds = dp_ds + Nx(0,a) * yl(i,a);
+          }
+          for (int j = 0; j < nsd; j++) {
+            grad_p(j) = dp_ds * fiber_tangent(j);
+          }
+        } else {
+          for (int a = 0; a < eNoN; a++) {
+            for (int j = 0; j < nsd; j++) {
+              grad_p(j) = grad_p(j) + Nx(j,a) * yl(i,a);
+            }
           }
         }
+
         for (int j = 0; j < nsd; j++) {
-          lRes(j) = -kappa * q(j);
+          lRes(j) = -mobility * grad_p(j);
         }
```

The replacement intentionally includes the viscosity correction so the reconstructed flux implements `-(K/mu) grad(P)` for both fiber and volume meshes.

zasexton (commented):
Suggested changes for Darcy equation initialization and fiber-flux reconstruction.

Code/Source/solver/consts.h
zasexton: I missed this when originally writing the darcy equations since I was unsure of the `dflag` purpose.

Code/Source/solver/darcy.cpp
```diff
@@ -0,0 +1,297 @@
...
+  double source = dmn.prop.at(PhysicalProperyType::source_term);
+  double beta_0 = dmn.prop.at(PhysicalProperyType::media_compressibility);
+  double rho_0 = dmn.prop.at(PhysicalProperyType::fluid_density);
+  double mu = dmn.prop.at(PhysicalProperyType::darcy_fluid_viscosity);
+
+  double T1 = eq.af * eq.gam * dt;
+  double amd = eq.am / T1;
+  double wl = w * T1;
+
+  double Pd = -source;
```
zasexton: > `Source_term` is currently folded into `Pd` and the complete expression is multiplied by `Media_compressibility`. A nonzero source therefore has no effect when the default compressibility is zero, and otherwise behaves as an offset to the pressure time derivative. The changes below separate the volumetric source from the storage term in all three kernels. With these changes applied, the residual represents `rho * beta * dp/dt - rho * source` plus the Darcy diffusion term, so `Source_term` remains active in the default steady/incompressible configuration.

## Suggested code changes

Update the 1-D kernel beginning at `Code/Source/solver/darcy.cpp:164`:

```diff
-  double Pd = -source;
+  double p_dot = 0.0;
   double Px = 0.0;

   for (int a = 0; a < eNoN; a++) {
-    Pd = Pd + N(a) * al(i, a);
+    p_dot = p_dot + N(a) * al(i, a);
     Px = Px + Nx(0, a) * yl(i, a);
   }

   for (int a = 0; a < eNoN; a++) {
-    lR(0, a) = lR(0, a) + w * (rho_0 * beta_0 * N(a) * Pd +
-                               (((k * rho_0) / mu) * (Nx(0, a) * Px)));
+    lR(0, a) = lR(0, a) +
+        w * (rho_0 * N(a) * (beta_0 * p_dot - source) +
+             ((k * rho_0) / mu) * Nx(0, a) * Px);
     for (int b = 0; b < eNoN; b++) {
       lK(0, a, b) = lK(0, a, b) + wl * (rho_0 * beta_0 * N(a) * N(b) * amd +
                                        ((((rho_0 * k) / mu) * (Nx(0, a) * Nx(0, b)))));
```

Update the 2-D kernel beginning at `Code/Source/solver/darcy.cpp:219`:

```diff
-  double Pd = -source;
+  double p_dot = 0.0;
   Vector<double> Px(nsd);

   for (int a = 0; a < eNoN; a++) {
-    Pd = Pd + N(a)*al(i,a);
+    p_dot = p_dot + N(a)*al(i,a);
     Px(0) = Px(0) + Nx(0,a)*yl(i,a);
     Px(1) = Px(1) + Nx(1,a)*yl(i,a);
   }

   for (int a = 0; a < eNoN; a++) {
-    lR(0,a) = lR(0,a) + w*(rho_0*beta_0*N(a)*Pd + (((k*rho_0)/mu)*(Nx(0,a)*Px(0)
-                                                                   + Nx(1,a)*Px(1))));
+    lR(0,a) = lR(0,a) +
+        w * (rho_0 * N(a) * (beta_0 * p_dot - source) +
+             ((k * rho_0) / mu) *
+                 (Nx(0,a) * Px(0) + Nx(1,a) * Px(1)));
     for (int b = 0; b < eNoN; b++) {
       lK(0,a,b) = lK(0,a,b) + wl*(rho_0*beta_0*N(a)*N(b)*amd +
                                   ((((rho_0*k)/mu)*(Nx(0,a)*Nx(0,b) +
```

Update the 3-D kernel beginning at `Code/Source/solver/darcy.cpp:276`:

```diff
-  double Pd = -source;
+  double p_dot = 0.0;
   Vector<double> Px(nsd);

   for (int a = 0; a < eNoN; a++) {
-    Pd = Pd + N(a) * al(i,a);
+    p_dot = p_dot + N(a) * al(i,a);
     Px(0) = Px(0) + Nx(0,a) * yl(i,a);
     Px(1) = Px(1) + Nx(1,a) * yl(i,a);
     Px(2) = Px(2) + Nx(2,a) * yl(i,a);
   }

   for (int a = 0; a < eNoN; a++) {
-    lR(0,a) = lR(0, a) + w*(rho_0*beta_0*N(a)*Pd +
-                            ((k*rho_0)/mu)*(Nx(0,a)*Px(0) + Nx(1,a)*Px(1) + Nx(2,a)*Px(2)));
+    lR(0,a) = lR(0, a) +
+        w * (rho_0 * N(a) * (beta_0 * p_dot - source) +
+             ((k * rho_0) / mu) *
+                 (Nx(0,a) * Px(0) + Nx(1,a) * Px(1) +
+                  Nx(2,a) * Px(2)));
     for (int b = 0; b < eNoN; b++) {
       lK(0,a,b) = lK(0,a,b) + wl*(rho_0*beta_0*N(a)*N(b)*amd +
```

For `Media_compressibility == 0`, the source contribution at test-function node `a` becomes `-w * rho_0 * N(a) * source` instead of zero. The source remains independent of compressibility, while the existing tangent matrix remains correct because `Source_term` is prescribed data rather than an unknown.

Code/Source/solver/set_equation_props.h
```diff
@@ -274,6 +274,37 @@ SetEquationPropertiesMapType set_equation_props = {
...
+{
+    using namespace consts;
+    auto& com_mod = simulation->get_com_mod();
+    lEq.phys = consts::EquationType::phys_darcy;
+
+    propL[0][0] = PhysicalProperyType::permeability;
+    propL[1][0] = PhysicalProperyType::source_term;
+    propL[2][0] = PhysicalProperyType::solid_density;
+    propL[3][0] = PhysicalProperyType::porosity;
+    propL[4][0] = PhysicalProperyType::fluid_density;
```
zasexton: > Darcy registers `PhysicalProperyType::fluid_density`, but this branch currently reads the generic `<Density>` value for every equation except CMM. The added Darcy inputs set `<Fluid_density>1</Fluid_density>`, so their requested value is silently replaced by the default `<Density>` value of `0.5`. The change below routes Darcy through the explicit `fluid_density` parameter alongside CMM. With it applied, the documented Darcy input populates `dmn.prop[PhysicalProperyType::fluid_density]` as intended. 

## Suggested code change

Update `Code/Source/solver/read_files.cpp:1552`:

```diff
           case PhysicalProperyType::fluid_density:
-            if (lEq.phys == EquationType::phys_CMM) {
+            if (lEq.phys == EquationType::phys_CMM ||
+                lEq.phys == EquationType::phys_darcy) {
               rtmp = domain_params->fluid_density.value();
             } else {
               rtmp = domain_params->density.value();
             }
           break;
```

For the four added Darcy XML inputs, `<Fluid_density>1</Fluid_density>` produces a stored Darcy fluid density of `1.0` rather than the unrelated default `<Density>` value of `0.5`.

Code/Source/solver/set_equation_props.h
```diff
@@ -274,6 +274,37 @@ SetEquationPropertiesMapType set_equation_props = {
...
+    propL[5][0] = PhysicalProperyType::porosity_pressure;
+    propL[6][0] = PhysicalProperyType::media_compressibility;
+    propL[7][0] = PhysicalProperyType::fluid_compressibility;
+    propL[8][0] = PhysicalProperyType::darcy_fluid_viscosity;
+    propL[9][0] = PhysicalProperyType::density_pressure;
```
zasexton: > Darcy currently registers and parses `solid_density`, `porosity`, `porosity_pressure`, `fluid_compressibility`, and `density_pressure`, although no Darcy kernel consumes them. The cleanup below removes the unused properties from the Darcy property list and removes the four newly introduced parser controls that have no implementation. These other parameters were holdovers from when I was trying different models; they can be removed. 

## Suggested code changes

Restrict the Darcy property list in `Code/Source/solver/set_equation_props.h:287` to properties consumed by assembly or output:

```diff
     propL[0][0] = PhysicalProperyType::permeability;
     propL[1][0] = PhysicalProperyType::source_term;
-    propL[2][0] = PhysicalProperyType::solid_density;
-    propL[3][0] = PhysicalProperyType::porosity;
-    propL[4][0] = PhysicalProperyType::fluid_density;
-    propL[5][0] = PhysicalProperyType::porosity_pressure;
-    propL[6][0] = PhysicalProperyType::media_compressibility;
-    propL[7][0] = PhysicalProperyType::fluid_compressibility;
-    propL[8][0] = PhysicalProperyType::darcy_fluid_viscosity;
-    propL[9][0] = PhysicalProperyType::density_pressure;
+    propL[2][0] = PhysicalProperyType::fluid_density;
+    propL[3][0] = PhysicalProperyType::media_compressibility;
+    propL[4][0] = PhysicalProperyType::darcy_fluid_viscosity;
```

Remove the newly added, unused XML parameter registrations from `Code/Source/solver/Parameters.cpp:2049`:

```diff
   set_parameter("Permeability", 0.0, !required, permeability);
-  set_parameter("Porosity", 0.0, !required, porosity);
-  set_parameter("Porosity_pressure", 0.0, !required, porosity_pressure);
   set_parameter("Media_compressibility", 0.0, !required, media_compressibility);
-  set_parameter("Fluid_compressibility", 0.0, !required, fluid_compressibility);
   set_parameter("Darcy_fluid_viscosity", 1.0, !required, darcy_fluid_viscosity);
-  set_parameter("Density_pressure", 0.0, !required, density_pressure);
```

Remove their storage members from `Code/Source/solver/Parameters.h:1627`:

```diff
     Parameter<double> permeability;
-    Parameter<double> porosity;
-    Parameter<double> porosity_pressure;
     Parameter<double> media_compressibility;
-    Parameter<double> fluid_compressibility;
     Parameter<double> darcy_fluid_viscosity;
-    Parameter<double> density_pressure;
```

Remove the unused physical-property identifiers and retain explicit values for the supported identifiers in `Code/Source/solver/consts.h:421`:

```diff
   inverse_darcy_permeability = 15,
   permeability = 16,
-  porosity = 17,
-  porosity_pressure = 18,
   media_compressibility = 19,
-  fluid_compressibility = 20,
-  darcy_fluid_viscosity = 21,
-  density_pressure = 22
+  darcy_fluid_viscosity = 21
 };
```

Remove the corresponding no-op cases from `Code/Source/solver/read_files.cpp:1588`:

```diff
           case PhysicalProperyType::permeability:
             rtmp = domain_params->permeability.value();
           break;

-          case PhysicalProperyType::porosity:
-            rtmp = domain_params->porosity.value();
-          break;
-
-          case PhysicalProperyType::porosity_pressure:
-            rtmp = domain_params->porosity_pressure.value();
-          break;
-
           case PhysicalProperyType::media_compressibility:
             rtmp = domain_params->media_compressibility.value();
           break;

-          case PhysicalProperyType::fluid_compressibility:
-            rtmp = domain_params->fluid_compressibility.value();
-          break;
-
           case PhysicalProperyType::darcy_fluid_viscosity:
             rtmp = domain_params->darcy_fluid_viscosity.value();
           break;
-
-          case PhysicalProperyType::density_pressure:
-            rtmp = domain_params->density_pressure.value();
-          break;
```

The implemented Darcy model retains exactly five material inputs, and each has a runtime consumer:

- `Permeability`: Darcy assembly and flux output
- `Source_term`: Darcy residual
- `Fluid_density`: Darcy residual and tangent
- `Media_compressibility`: transient storage term
- `Darcy_fluid_viscosity`: Darcy mobility in assembly and flux output

The removed names are no longer presented as supported Darcy controls.

Code/Source/solver/CMakeLists.txt
```diff
@@ -170,6 +170,7 @@ set(CSRCS
   SimulationLogger.h
   VtkData.h VtkData.cpp
 
+  active_stress_regazzoni.h active_stress_regazzoni.cpp
```
michelebucelli: This change seems to be unrelated to the PR. I suggest reverting to keep all the active stress files close together in the list.

Code/Source/solver/consts.h
michelebucelli: Might be a good opportunity to fix this typo and rename this to `PhysicalPropertyType`.

Code/Source/solver/consts.h
```diff
@@ -347,6 +349,7 @@ enum class OutputNameType {
   outGrp_activeTensionFibers = 529,
   outGrp_activeTensionSheets = 530,
   outGrp_activeTensionNormal = 531,
+  outGrp_mbfFlx = 532,
```
michelebucelli: Since there's a correspondence between entries of the `outGrp` category and of the `out` category, I recommend using the same naming convention for both (that is, either both `mbfFlx` or `mbfFlux`). In a similar way, I'd suggest to use consistent capitalization of `MBF` (i.e. if it is all caps in `out_MBF`, I think it should be all caps in `out_MBFFlx` and `outGrp_MBFFlx`).

Finally, I would suggest avoiding abbreviations when possible, unless they're very established (that is to say: I don't mind `mbf`, but I think `Flx` should be `Flux`).

(These are not just meant as cosmetic changes, but also to prevent inexperienced users from thinking that `flx` and `flux` are two different things since they have two different names)

Code/Source/solver/consts.h
```diff
@@ -380,7 +383,9 @@ enum class OutputNameType {
   out_fibStretchRate = 570,
   out_activeTensionFibers = 569,
   out_activeTensionSheets = 568,
-  out_activeTensionNormal = 567
+  out_activeTensionNormal = 567,
+  out_MBF = 566,
+  out_mbfFlux = 565
 };
```
michelebucelli, line 389: Unrelated to this PR, but it's been bugging me for a while 😅 (also discussed in issue #548)

@ktbolt Do you know whether this `enum class` could be split into two `enum class`es, one for the `outGrp_*` entries and one for the `out_*` entries? It seems to me that they are not interchangeable, based on the way they are used, so it would seem they should be separate.

Code/Source/solver/consts.h
```diff
@@ -380,7 +383,9 @@ enum class OutputNameType {
   out_fibStretchRate = 570,
   out_activeTensionFibers = 569,
   out_activeTensionSheets = 568,
-  out_activeTensionNormal = 567
+  out_activeTensionNormal = 567,
+  out_MBF = 566,
+  out_mbfFlux = 565
```
michelebucelli: What is the difference between `MBF` and `mbfFlux`? Expanding the acronym, the latter reads "myocardial blood flow flux", and I'm not sure about its physical meaning. Perhaps with a bit of clarification we can come up with a more self-documenting name (or, failing that, we can document it).

Code/Source/solver/post.cpp
```diff
@@ -916,11 +916,29 @@ void post(Simulation* simulation, const mshType& lM, Array<double>& res, const S
...
+            }
+          }
+
+          const double tangent_norm = utils::norm(fiber_tangent);
+          if (utils::is_zero(tangent_norm)) {
+            throw std::runtime_error(
+                "[post] Cannot compute Darcy flux for a degenerate fiber element.");
+          }
+          fiber_tangent = fiber_tangent / tangent_norm;
+        }
```
michelebucelli: I am unsure about what these lines are computing (which may or may not relate to me not knowing what `mbfFlux` is meant to be). Are these fibers the same as myocardial fibers (the ones used in mechanics and electrophysiology)?

Code/Source/solver/Parameters.h
```diff
@@ -1623,6 +1623,9 @@ class DomainParameters : public ParameterLists
     Parameter<double> source_term;
     Parameter<double> time_step_for_integration;
     
+    Parameter<double> permeability;
+    Parameter<double> media_compressibility;
+    Parameter<double> darcy_fluid_viscosity;
     // Inverse of Darcy permeability. Default value of 0.0 for Navier-Stokes and non-zero for Navier-Stokes-Brinkman
     Parameter<double> inverse_darcy_permeability;
```
michelebucelli: I fear there's a potential source of confusion here, with the parameter `inverse_darcy_permeability` being used for Navier-Stokes and not for Darcy (I think), and the parameter `permeability` being used for Darcy insteady.

The structural solution to this would be for these parameters to live in different scopes, tied to the physical model they relate to. This however would be a large refactoring, that is probably beyond this PR, and will probably come more naturally once we start rethinking the assembly.

In the interim, we can try to think of better names for these. I recommend using the prefix `darcy` for all those parameters pertaining to the newly added Darcy model, and dropping it from those that don't.

Code/Source/solver/post.cpp
```diff
@@ -1011,6 +1029,38 @@ void post(Simulation* simulation, const mshType& lM, Array<double>& res, const S
...
 
+      // MBF Flux calculation
+      // 
+      } else if (outGrp == OutputNameType::outGrp_mbfFlx) {
+        const double permeability =
+            eq.dmn[cDmn].prop[PhysicalProperyType::permeability];
+        const double viscosity =
+            eq.dmn[cDmn].prop[PhysicalProperyType::darcy_fluid_viscosity];
+        const double mobility = permeability / viscosity;
+        const int i = eq.s;
```
michelebucelli: I suggest renaming this to `equation_index` to enhance readability of the lines that follow.

Code/Source/solver/post.cpp
```diff
@@ -1011,6 +1029,38 @@ void post(Simulation* simulation, const mshType& lM, Array<double>& res, const S
...
+          for (int a = 0; a < eNoN; a++) {
+            for (int j = 0; j < nsd; j++) {
+              grad_p(j) = grad_p(j) + Nx(j,a) * yl(i,a);
+            }
+           }
+         }
+
+         for (int j = 0; j < nsd; j++) {
+          lRes(j) = -mobility * grad_p(j);
+         }
```
michelebucelli: I think the indentation of this `for` loop is off, it should be one space less:
```suggestion
        for (int j = 0; j < nsd; j++) {
          lRes(j) = -mobility * grad_p(j);
        }
```
I also recommend using some automatic indentation tool to prevent this kind of thing from happening accidentally (e.g. `clang-format`).

@ktbolt On this note, are there any updates on the automatic indentation setup we discussed in #578 and #96? I'd be happy to help getting the ball rolling on that if you want, since I'm using auto-formatting a lot already (and I'd be happy to stop checking for accidental whitespace changes in my commits 😅)

Code/Source/solver/Parameters.cpp
```diff
@@ -2044,6 +2044,10 @@ DomainParameters::DomainParameters() {
 
   set_parameter("Penalty_parameter", 0.0, !required, penalty_parameter);
   set_parameter("Poisson_ratio", 0.3, !required, poisson_ratio);
+  
+  set_parameter("Permeability", 0.0, !required, permeability);
```
michelebucelli: I think that 0 is not a sensible default value for permeability, since setting the permeability to 0 would lead to an ill-posed (or at least less well-posed) problem.

Possible solutions:
1. Using a positive default value. It might not be the cleanest fix, since that value would be dimensional and thus would not be meaningful if the system of units changes.
2. Make `Permeability` required if using the Darcy equation. In that case, the user would be forced to provide a value. I think this would require somewhat larger changes to the overall parameter infrastructure, probably beyond the scope of this PR.
3. Add a check that `Permeability > 0`, if the Darcy equation is found, and leave the default to 0. This is a convoluted way of obtaining the same effect of option 2 (the parameter becomes required if using Darcy) without significant infrastructural changes. Less clean than option 2, but probably good enough pending the larger refactoring.

I lean towards option 3.

I'm not sure if similar considerations apply to the `Media_compressibility` parameter below.

Code/Source/solver/darcy.cpp
```diff
@@ -0,0 +1,303 @@
...
+    P -> Pressure
+ -------------------------------------------------------------
+ Weak form of the Single-Compartment Darcy equation:
+    -∫(∇q∇P)dΩ - λ∫qPdΩ = ∫qFdΩ - ∫q∇P⋅nvdΓ
+ where:
+    q -> Test function
+    λ -> (β0 + β1)/K
+    F -> -(β0(P_source) + β1(P_sink))/K
+    n -> Normal vector to the boundary
+*/
```
michelebucelli: 1. I would move this block to be the documentation of the `darcy` namespace, since that will make it visible from the Doxygen-generated documentation too.
2. In the same spirit, I would use Doxygen math syntax (that is, LaTeX) to style the equations, so that they will be rendered as proper math in the Doxygen-generated documentation. (I'm pretty sure AI can do the conversions from the current code pretty easily).
3. I think it would be a good idea to include one or more references, with DOI, to papers that describe this model, ideally in the cardiovascular setting.

Code/Source/solver/darcy.cpp
```diff
@@ -0,0 +1,303 @@
...
+ Assumptions:
+    - Homogeneous Permeability
+    - Homogeneous Density
+    - Isotropic Permeability
+    - Assumptions of Stokes Flow
+    - Steady-State
+ -------------------------------------------------------------
+ Strong form of the Single-Compartment Darcy equation:
+    u = -K∇(P)
+    ∇⋅u = β0(P_source - P) - β1(P - P_sink)
```
michelebucelli: Judging by the weak formulation below, the code actually implements the single-field formulation, which is obtained by taking the divergence of the first equation here, whose strong form is:

$$
-\nabla\cdot(K \nabla p) - \beta_0 (p_{source} - p) + \beta_1 (p - p_{sink}) = 0
$$

I suggest reporting this strong form instead, since it is the one actually being solved for. Otherwise, there's the risk of implying to the reader that we're using a mixed formulation in which $u$ and $p$ are primal variables in the finite element formulation.

Code/Source/solver/darcy.cpp
```diff
@@ -0,0 +1,303 @@
...
+    K -> Permeability tensor
+    P -> Pressure
+ -------------------------------------------------------------
+ Weak form of the Single-Compartment Darcy equation:
+    -∫(∇q∇P)dΩ - λ∫qPdΩ = ∫qFdΩ - ∫q∇P⋅nvdΓ
+ where:
+    q -> Test function
+    λ -> (β0 + β1)/K
+    F -> -(β0(P_source) + β1(P_sink))/K
+    n -> Normal vector to the boundary
```
michelebucelli: Is there any specific reason why this formulation is divided by the permeability $K$, as opposed to keeping it in the diffusion term, which I think is the more common approach?

The advantage of the latter would be that it lends more naturally to extensions (e.g. heterogeneous or anisotropic diffusion), which may not be relevant now but might become so in the future. The current formulation instead only works if $K$ is a constant scalar.

Code/Source/solver/darcy.cpp
```diff
@@ -0,0 +1,303 @@
...
+    - Assumptions of Stokes Flow
+    - Steady-State
+ -------------------------------------------------------------
+ Strong form of the Single-Compartment Darcy equation:
+    u = -K∇(P)
+    ∇⋅u = β0(P_source - P) - β1(P - P_sink)
+ where:
+    u -> Volume flux vector
+    K -> Permeability tensor
+    P -> Pressure
```
michelebucelli: I would suggest also defining $p_{source}$ and $p_{sink}$ in this list, and the two coefficients $\beta_0$ and $\beta_1$. I would also suggest clarifying for each what is a problem unknown, what is a quantity derived from the unknown (e.g. $u = -K\nabla p$) and what is a user-specified parameter. For user-specified parameter, it is also a good idea to document the set they belong to (e.g. $K \in (0, +\infty)$ and so on), so that the reader can understand what is spatially varying and what is assumed to be constant.

Some of this is documented by the assumptions above, e.g. "isotropic permeability", but I think it's not a bad idea to include these assumptions in the formal statement of the equation as well.

Code/Source/solver/darcy.cpp
```diff
@@ -0,0 +1,303 @@
...
+    cDmn = all_fun::domain(com_mod, lM, cEq, e);
+    auto cPhys = eq.dmn[cDmn].phys;
+    if (cPhys != EquationType::phys_darcy) {
+      continue;
+    }
+
+    // Update shape function for NURBS
+    if (lM.eType == ElementType::NRB) {
+      //CALL NRMNNX(lm, e)
+    }
```
michelebucelli: I know this is present also elsewhere, but: should we consider cleaning this up? As far as I understand it, NURBS are currently unsupported, I'm not sure there's plans to support them in the immediate future.

If we want to keep it as a placeholder, I'd at least suggest to throw a not-implemented exception here, instead of silently doing nothing, so that if the code somehow gets here the user/developer is made very aware of it.

Code/Source/solver/darcy.cpp
```diff
@@ -0,0 +1,303 @@
...
+  for (int a = 0; a < eNoN; a++) {
+    lR(0, a) = lR(0, a) +
+        w * (rho_0 * N(a) * (beta_0 * p_dot - source) +
+             ((k * rho_0) / mu) * Nx(0, a) * Px);
+    for (int b = 0; b < eNoN; b++) {
+      lK(0, a, b) = lK(0, a, b) + wl * (rho_0 * beta_0 * N(a) * N(b) * amd +
+                                       ((((rho_0 * k) / mu) * (Nx(0, a) * Nx(0, b)))));
+    }
+  }
+}
```
michelebucelli: If I am reading this code right, the problem assembled here does not match the one stated in the documentation at the top of this file. The differences I can notice are the following.
1. The terms assembled here do not have anything that corresponds to the sink term $\beta_1 (p - p_{sink})$.
2. The term multiplied by $\beta_0$ is not $\beta_0 (p - p_{source})$, but rather $\beta_0 \rho_0 \frac{\partial p}{\partial t}$
3. There's a couple of coefficients ($\rho_0$ and $\mu$) that rescale the permeability $K$.

It seems to me that the strong form of the problem assembled here is

$$
\rho_0\beta_0\frac{\partial p}{\partial t} - \nabla \cdot\left(\frac{K\rho_0} {\mu}\nabla p\right) = \rho_0 p_{source}
$$

Is this right?

If so, I would suggest confirming that this is indeed the problem that is meant to be solved, and update the documentation accordingly.

Additionally, in the above equation $\rho_0$ can be simplified out (assuming it is homogeneous and constant, which seems to be the case), so I suggest to consider whether it makes sense to remove it altogether. Similarly, $\mu$ only acts as a rescaling term on $K$, so just asking the user to provide a rescaled permemability might make for a simpler parameterization.

The same comments apply to `darcy_2d` and `darcy_3d` below.

Code/Source/solver/darcy.cpp
```diff
@@ -0,0 +1,303 @@
...
+
+void darcy_2d(ComMod& com_mod, const int eNoN, const double w, const Vector<double>& N, const Array<double>& Nx,
+              const Array<double>& al, const Array<double>& yl, Array<double>& lR, Array3<double>& lK)
```
michelebucelli, line 160: It seems to me that `darcy_1d`, `darcy_2d` and `darcy_3d` are essentially identical functions, but for the way they unroll the loops needed to compute the scalar products (and maybe the way they assemble the pressure gradient `Px`).

I think it would be rather easy to merge this into a single function, in one of these ways.
1. Templating the `darcy` function over the spatial dimension, and then using `if constexpr` statements to switch between the optimized cases for the individual dimensions. For example, the interpolation of the pressure gradient could look something like
   ```cpp
   for (int a = 0; a < eNoN; a++) {
     p_dot = p_dot + N(a)*al(i,a);

     Px(0) = Px(0) + Nx(0,a)*yl(i,a);     
     if constexpr (nsd > 1)
       Px(1) = Px(1) + Nx(1,a)*yl(i,a);
     if constexpr (nsd > 2)
       Px(2) = Px(2) + Nx(2,a)*yl(i,a);
   }
   ```
2. Better than the above, move this dimension-dependent operations (interpolation of the gradient, dot products) to dedicated helper functions, and template those on spatial dimension if necessary (I'm pretty sure that the dot product between vectors is already implemented somewhere). If manual loop unrolling is needed for efficiency, this can be hidden away in the helper functions, which will be good for readability.

Either way, this would ensure more robustly that the 1D, 2D and 3D variants of the equation behave consistently, and it will make it easier to modify the formulation of the equation, should this be needed in the future.

I am aware that the same pattern is present in several other equations, so this is perhaps a more general comment. But it might be a good idea to start applying this to new equations, leaving the improvement of other equations for incoming refactoring.

michelebucelli (commented):
Thank you @mmegally! I left a few comments.

Code/Source/solver/consts.h
ktbolt: @michelebucelli The `OutputNameType` enum class represents constants for output but the `outGrp_` and `out_` constants really do represent two different things. So go ahead and split  `OutputNameType` up if you would like to do that now; all of the output code will definitely be refactored.

Code/Source/solver/post.cpp
mmegally: See the previous review [comment ](https://github.com/SimVascular/svMultiPhysics/pull/620#discussion_r3867910359) from @zasexton.

michelebucelli: Thanks, I had missed that comment! However, I am still not quite sure I understand what this is doing, in the sense that I don't exactly understand what the mathematical quantity being computed here is, and what its physical significance is. Could you try clarifying this a bit?

Code/Source/solver/post.cpp
zasexton: Will fix here. I can also help with the auto-formatting discussions/issues so that we can avoid these in the future.

Code/Source/solver/CMakeLists.txt
mmegally: Moved the active_stress_regazzoni.cpp file to bottom of the active stress files list, aligning with current order in SimVascular:main. See commit 993bbc5f27b90570cc187ff3ac686d91867467eb

Code/Source/solver/consts.h
mmegally: Updated PhysicalProperyType to PhysicalPropertyType in all affected files (e.g. fluid.cpp, stokes.cpp, etc.). See commits See commits:  993bbc5f27b90570cc187ff3ac686d91867467eb &
d0633979ec6b66f78d66bc3f2a517a27baa38be8

Code/Source/solver/consts.h
mmegally: Recently, we were playing around with different names. Initially, I had changed it to outGrp_MBFFlux as you were suggesting here. In the most recent commit (5682189b02579ad08f93cc10715fd759524a9256) we opted to renamed it to outGrp_darcyFlux. Similarly, rather than out_MBF and out_mbfFlux, we now have out_darcyPressure and out_darcyFlux. This change was done to drop the old naming MBF naming convention, which as your future comments suggest, may have caused confusion towards the physical meaning of these outputs.

Code/Source/solver/post.cpp
mmegally: To my understanding, the calculation for the fiber tangent unit vector was relevant to the 1D case to determine the pressure gradient with respect to the pressure derivative along the local 1D coordinate. The pressure gradient is then necessary to calculate the MBF_Flux output, which we have now renamed Darcy_flux as a more self-explanatory name (see previous comment). However, we lately opted to only preserve the 2d and 3d functions to be consistent with the implementation for other problems handled in svMultiPhysics. @zasexton feel free to correct me if anything I mentioned here is not completely accurate. These deletions are viewable in commit 5682189b02579ad08f93cc10715fd759524a9256

Code/Source/solver/consts.h
mmegally: We had initially been using the name MBF for the computed pressure and MBF_Flux for the darcy flux/velocity calculated via Darcy's constitutive equation. In commit 5682189b02579ad08f93cc10715fd759524a9256 we have since renamed these variables to be output as "Darcy_pressure" and "Darcy_flux"

Code/Source/solver/Parameters.h
mmegally: Renamed variables permeability and media_compressibility to darcy_permeability and darcy_media_compressibility respectively. Changes made in commit 993bbc5f27b90570cc187ff3ac686d91867467eb

Code/Source/solver/post.cpp
mmegally: Suggested change added in commit 993bbc5f27b90570cc187ff3ac686d91867467eb

Code/Source/solver/Parameters.cpp
mmegally: As I understand it, the use of 0 as a default value in addition to what you later ask about dividing by permeability was an intentional decision such that the solver would throw an error if the user did not manually specify a permeability value. I have not touched this default value in my commit history. @zasexton is this something you have changed in your commits or are we opting to keep this current default value intact?

Code/Source/solver/darcy.cpp
mmegally: Addressed changes (1-3) in commit 993bbc5f27b90570cc187ff3ac686d91867467eb

Code/Source/solver/darcy.cpp
mmegally: Addressed in commit 993bbc5f27b90570cc187ff3ac686d91867467eb

Code/Source/solver/darcy.cpp
mmegally: Addressed in commit 993bbc5f27b90570cc187ff3ac686d91867467eb

Code/Source/solver/darcy.cpp
mmegally: See comment above: https://github.com/SimVascular/svMultiPhysics/pull/620#discussion_r3918477971

Code/Source/solver/darcy.cpp
mmegally: See comment above: https://github.com/SimVascular/svMultiPhysics/pull/620#discussion_r3918611556

Code/Source/solver/darcy.cpp
ktbolt: Please clean up. NURBS may added be some day but even so this is not how they would be implemented.

Code/Source/solver/Parameters.cpp
ktbolt: @mmegally You could add a check in `DomainParameters::set_values` to see if `Permeability` was set.

Code/Source/solver/post.cpp
zasexton: This has been removed now and validation guards have been added to specifically reject fiber meshes. This originally was bringing in other diff changes that I had made on my local branch from a while back when I was testing some embedded "fiber" models along perfusion directions. Obviously this shouldn't be included here

Code/Source/solver/Parameters.cpp
zasexton: Added in 4871ba21 for domain property validation

Code/Source/solver/darcy.cpp
zasexton: Removed in latest commits by @mmegally and the merged branch

Code/Source/solver/darcy.cpp
zasexton: Opting to keep the viscosity and permeability separate at this time since there will likely be some ongoing discussions with the wet-lab experimentalists that will likely want some 1:1 correspondence for their measurements to input parameters. Also keeping the 2d and 3d assembly kernels separate at this time to mirror the other physics files; although I agree that future refactoring should cleanup these to have better concision of code.

zasexton: Also yes the equation that we should be solving for this model is:

$$\rho\beta\frac{\partial p}{\partial t}-\nabla\cdot\left(\frac{\rho K}{\mu}\nabla p\right)=\rho s$$.

$\beta$ should be the storage/compressibility and $s$ is the prescribed volumetric source. You are correct that we do not implement the $p_{source}$ and $p_{sink}$ exchange terms from the original documentation that was written. That documentation has been updated to reflect the current assembly implementation.

Code/Source/solver/initialize.cpp
zasexton, line 1006: In `zero_init()` I believe that we are supplying the pressure assuming the fluid/velocity solution layout which is erroneous for the Darcy problem since we will only have one state variable. This should be index 0. I believe this will lead to heap allocation errors for 2D problem cases...I will take another look at this

zasexton (commented):
In my opinion we only piece requiring a bit more review is the zero initialization for the 2D cases to make sure that we are not getting heap error allocations and out-of-bounds. I'll fork this branch and double-check now to make any appropriate changes, if necessary.

Code/Source/solver/initialize.cpp
zasexton: this should now be addressed in the PR for this branch from this commit [a87e729](https://github.com/mmegally/svMultiPhysics/pull/2/commits/a87e72981ffa6de9ee9b9c16d0c95f9285e90d7a)

Code/Source/solver/darcy.cpp
```diff
@@ -44,6 +69,8 @@ void b_darcy(ComMod& com_mod, const int eNoN, const double w, const Vector<doubl
 
 void construct_darcy(ComMod& com_mod, const mshType& lM, const SolutionStates& solutions)
 {
+  validate_element_support(lM);
```
michelebucelli: Would it make logical sense for this validation to happen after parsing the mesh/element information, rather than whenever calling `construct_darcy`? Not sure what the right place in the code for that would be though.

Code/Source/solver/darcy.cpp
```diff
@@ -128,58 +147,15 @@ void construct_darcy(ComMod& com_mod, const mshType& lM, const SolutionStates& s
         darcy_3d(com_mod, eNoN, w, N, Nx, al, yl, lR, lK);
       } else if (insd == 2) {
         darcy_2d(com_mod, eNoN, w, N, Nx, al, yl, lR, lK);
-      } else if (insd == 1) {
-        darcy_1d(com_mod, eNoN, w, N, Nx, al, yl, lR, lK);
       } else {
-        throw std::runtime_error("[construct_darcy] insd must be 1, 2 or 3.");
+        throw std::runtime_error("[construct_darcy] insd must be 2 or 3.");
```
michelebucelli, line 150: I suggest using one of the exceptions from the `svmp` namespace here.

Code/Source/solver/darcy.h
```diff
@@ -1,28 +1,116 @@
...
+ *   defaults to \f$10^{-15}\f$ and must satisfy \f$K > 0\f$.
+ * - \f$\mu\f$: configured dynamic viscosity. `Darcy_fluid_viscosity` defaults
+ *   to 1 and must satisfy \f$\mu > 0\f$.
+ * - \f$\rho\f$: configured reference fluid density. `Fluid_density` defaults
+ *   to 0.5 and must satisfy \f$\rho > 0\f$.
+ * - \f$\beta\f$: configured storage/compressibility.
+ *   `Darcy_media_compressibility` defaults to 0 and must satisfy
+ *   \f$\beta \ge 0\f$.
+ * - \f$s\f$: configured volumetric source provided by `Source_term`; it
+ *   defaults to 0 and is constant within each configured domain.
```
michelebucelli, line 41: Thanks, this is much better detailed now!

A further recommendation I have is that for all these quantities it would be good to document their dimensions. For example, does the Darcy velocity $q$ have the units of a velocity (length per unit time), of a volumetric flow rate (volume, or length^3 per unit time) or mass flow rate (mass per unit time)?

Some of this can be inferred from the equation, but writing it down spares the reader from doing the inference :smile:

Code/Source/solver/darcy.cpp
michelebucelli: Thanks for clarifying! I assume that you opted to keep the density separate for the same reason as viscosity and compressibility, right?

Code/Source/solver/Parameters.cpp
```diff
@@ -2045,8 +2053,9 @@ DomainParameters::DomainParameters() {
   set_parameter("Penalty_parameter", 0.0, !required, penalty_parameter);
   set_parameter("Poisson_ratio", 0.3, !required, poisson_ratio);
   
-  set_parameter("Permeability", 0.0, !required, permeability);
-  set_parameter("Media_compressibility", 0.0, !required, media_compressibility);
+  set_parameter("Darcy_permeability", 1e-15, !required, darcy_permeability);
+  set_parameter("Darcy_media_compressibility", 0.0, !required,
```
michelebucelli, line 2057: Very minor, and somewhat pedantic: I think the correct form should be "medium compressibility". I would not object to switching to "Darcy compressibility" for brevity though.

Code/Source/solver/post.cpp
```diff
@@ -842,6 +843,10 @@ void post(Simulation* simulation, const mshType& lM, Array<double>& res, const S
   bool FSIeq = false;
   auto& eq = com_mod.eq[iEq];
 
+  if (outGrp == OutputNameType::outGrp_darcyFlux) {
+    darcy::validate_element_support(lM);
+  }
```
michelebucelli: Similar suggestion to before, this validation seems a bit out of place here. Is there a more centralized place where this can be checked during initialization?

michelebucelli (commented):
Thanks! I did a second pass, I only have a few minor suggestions left.

Code/Source/solver/Parameters.cpp
zasexton: I agree. @mmegally would you mind making this change?

zasexton: And yes, I'm using the same reasoning for the density separation as well. There will likely be more refinements this physics workflow as we apply it to a range of clinical and bench-top applications, but i think this is fine for now.

Code/Source/solver/darcy.cpp
zasexton: we could remove the validation in general here. I'm not sure that the element support should be a property of the physics formulation file. it seems more appropriate in upstream conditional setup for these problems.

Code/Source/solver/darcy.cpp
zasexton: Good catch! Thank you!

Code/Source/solver/post.cpp
mmegally: Removed validation in commit https://github.com/SimVascular/svMultiPhysics/commit/3d3d45744618de996c43960e510ac2d3418d9c98

Code/Source/solver/darcy.h
mmegally: Updated documentation in commit https://github.com/SimVascular/svMultiPhysics/commit/3d3d45744618de996c43960e510ac2d3418d9c98

Code/Source/solver/darcy.cpp
mmegally: Removed validation in commit https://github.com/SimVascular/svMultiPhysics/commit/3d3d45744618de996c43960e510ac2d3418d9c98

Code/Source/solver/darcy.cpp
mmegally: Replaced both instances of std::runtime_error that appeared in darcy.cpp with exceptions from the svmp namespace in commit https://github.com/SimVascular/svMultiPhysics/pull/620/changes/c1597932e731e76ba6239cd140a9565ba3212376

Code/Source/solver/Parameters.cpp
mmegally: Renamed to darcy_media_compressibility to darcy_compressibility within all relevant files in commits https://github.com/SimVascular/svMultiPhysics/pull/620/changes/a9bbd48d3d70a24d613fff64c8d7524097e66a13 & https://github.com/SimVascular/svMultiPhysics/pull/620/changes/a68f7c40b9034a1d7f890ac90166be8a99b69c3c

Code/Source/solver/post.cpp
zasexton: I believe this current nodal averaging will fail for quadratic elements. I know this is probably being a bit pedantic but if a user were to supply quadratic triangle or tet elements we would probably get NaN values or blow-up down at line 1118 during division. Instead we should allow an L2 projection here to resolve these nodal flux values for higher-order elements.

zasexton (commented):
We are getting very close to closing this PR I believe. I'll make one more fork to address a comment I made about the `post.cpp` file for higher-order elements. Ultimately I don't think this is a show-stopper for the regular use case of this physics but we should guarantee that it produces expected values if higher-order meshes are ever used with the darcy physics.

zasexton (approved):
I believe this PR is in a good position to merge

Code/Source/solver/darcy.cpp
```diff
@@ -135,8 +134,7 @@ void construct_darcy(ComMod& com_mod, const mshType& lM, const SolutionStates& s
         auto Nx_g = lM.Nx.slice(g);
         nn::gnn(eNoN, nsd, insd, Nx_g, xl, Nx, Jac, ksix);
         if (utils::is_zero(Jac)) {
-          throw std::runtime_error(
-              "[construct_darcy] Jacobian for element " + std::to_string(e) + " is < 0.");
+          svmp::InternalErrorException("[construct_darcy] Jacobian for element " + std::to_string(e) + " is < 0.");
         }
```
michelebucelli, line 138: I think this just constructs the exception object without throwing it. The correct version should be:
```cpp
svmp::check<svmp::InternalErrorException>(!utils::is_zero(Jac),
  "Jacobian for element " + std::to_string(e) + " is < 0.");
```
(I dropped the context tag `[construct_darcy]` because it will be part of the stack trace reported by the exception message, I believe).

Code/Source/solver/darcy.cpp
```diff
@@ -148,7 +146,7 @@ void construct_darcy(ComMod& com_mod, const mshType& lM, const SolutionStates& s
       } else if (insd == 2) {
         darcy_2d(com_mod, eNoN, w, N, Nx, al, yl, lR, lK);
       } else {
-        throw std::runtime_error("[construct_darcy] insd must be 2 or 3.");
+        svmp::raise<svmp::FE::InvalidArgumentException>("[construct_darcy] insd must be 2 or 3.");
```
michelebucelli: Same as above, I suggest dropping the context tag:
```cpp
svmp::raise<svmp::FE::InvalidArgumentException>("insd must be 2 or 3.");
```

michelebucelli (approved):
Thanks for addressing my previous comments! Just a few very minor leftovers, but otherwise approved.

zasexton:
Great! Thanks @michelebucelli for the review. @mmegally after making a few minor edits to address the lingering comments I think you should feel free to merge this code :)

Code/Source/solver/darcy.cpp
mmegally: Dropped the context tags in the last commit https://github.com/SimVascular/svMultiPhysics/pull/620/changes/ae99df477a2015b554271c9076b42a8542d92323

PR #624, opened 2026-08-27
Fix issues with history output

### LINKED ISSUE #615: Various issues with history standard output in electromechanics simulations
### Description

For an electromechanics simulation with a `cep` equation followed by a `struct` equation, the history output currently looks like this:
```
---------------------------------------------------------------------
 Eq     N-i     T       dB  Ri/R1   Ri/R0    R/Ri     lsIt   dB  %t
---------------------------------------------------------------------
 EP 1-1  2.872e-01  [0 1.000e+00 1.000e+00 2.171e-08]  [14 0 1]
 ST 1-1  2.002e+00  [0 1.000e+00 1.000e+00 9.875e-10]  [21 -90 19]
 ST 1-2  3.966e+00  [-95 1.743e-05 1.743e-05 5.515e-10]  [32 -93 30]
 ST 1-3  5.734e+00  [-153 2.127e-08 2.127e-08 7.245e-10]  [24 -91 21]
---------------------------------------------------------------------
 EP 1-1  7.703e+00  [0 1.000e+00 1.000e+00 2.171e-08]  [14 0 0]
---------------------------------------------------------------------
```
The issues with this are:
1. `EP 1-1` gets printed twice, one at the beginning (as it should), and one at the end (and this shouldn't happen);
2. there's a missing `ST 1-4` iteration (it is replaced by the duplicated `EP 1-1`);
3. as I understand it, the separator is meant to separate different timesteps, so here the first separator is spurious.

If the order of `cep` and `struct` is swapped, the output looks like this:
```
---------------------------------------------------------------------
 Eq     N-i     T       dB  Ri/R1   Ri/R0    R/Ri     lsIt   dB  %t
---------------------------------------------------------------------
 ST 1-1  1.123e+00  [0 1.000e+00 1.000e+00 5.573e-10]  [26 -93 22]
 ST 1-2  2.135e+00  [-67 4.079e-04 4.079e-04 6.330e-10]  [28 -92 27]
 ST 1-3  3.079e+00  [-113 2.002e-06 2.002e-06 8.610e-10]  [22 -91 21]
 ST 1-4  4.115e+00  [-188 3.930e-10 3.930e-10 5.195e-10]  [28 -93 27]
 ST 1-4  4.185e+00  [-188 3.930e-10 3.930e-10 5.195e-10]  [28 -93 100]
---------------------------------------------------------------------
```
The issues with this are:
1. there's no `EP 1-1` line, although electrophysiology is being solved;
2. there's a duplicate `ST 1-4` line (and I think the second one replaces the missing `EP 1-1`).

I do not know if similar problems arise with other multi-equation cases.

### Reproduction

Run an electromechanics test case (e.g. the `cases/electromechanics/slab`). That there's a missing `ST` iteration is not obvious, but can be inferred either from the timing or by adding some debug print statements in the code.

### Expected behavior

Most importantly, the output should print all the iterations that are run, and only one line for each iteration (no missing iterations, no duplicates).

On a more cosmetic side, I would like to:
1. clean up the use of separators (only one separator line between timesteps in the multi-equation cases, and no separator otherwise);
2. improve the horizontal alignment by ensuring that all column are fixed width. This would improve readability to some extent.

### Additional context

_No response_

### Code of Conduct

michelebucelli:
Fixes #615 about incorrect reporting of iterations in multi-equation simulations. Also adds some cosmetic improvements to the history table. See the issue for details.

This has no effect on calculations, but only on convergence history reporting.

### DIFF
--- a/Code/Source/solver/main.cpp
+++ b/Code/Source/solver/main.cpp
@@ -253,7 +253,6 @@ void iterate_solution(Simulation* simulation)
 
   bool l1 = false;
   bool l2 = false;
-  bool l3 = false;
 
   #ifdef debug_iterate_solution
   dmsg << "Start Outer Loop ..." << std::endl;
@@ -349,8 +348,11 @@ void iterate_solution(Simulation* simulation)
     dmsg << "Starting Newton iteration via Integrator ..." << std::endl;
     #endif
 
-    int iEqOld = cEq;
-    integrator.step();
+    // Results of this time step are written to a VTU file
+    const bool save_vtu = com_mod.saveVTK && cTS % com_mod.saveIncr == 0 &&
+                          cTS >= com_mod.saveATS;
+
+    integrator.step(save_vtu);
 
     #ifdef debug_iterate_solution
     dmsg << ">>> End of Newton iteration" << std::endl;
@@ -471,28 +473,13 @@ void iterate_solution(Simulation* simulation)
     // Writing results into the disk with VTU format
     //
     #ifdef debug_iterate_solution
-    dmsg; 
-    dmsg << "saveVTK: " << com_mod.saveVTK; 
+    dmsg;
+    dmsg << "saveVTK: " << com_mod.saveVTK;
+    dmsg << "save_vtu: " << save_vtu;
     #endif
 
-    if (com_mod.saveVTK) {
-      l2 = ((cTS % com_mod.saveIncr) == 0);
-      l3 = (cTS >= com_mod.saveATS);
-      #ifdef debug_iterate_solution
-      dmsg << "l2: " << l2; 
-      dmsg << "l3: " << l3; 
-      #endif
-
-      if (l2 && l3) {
-        output::output_result(simulation, com_mod.timeP, 3, iEqOld);
-        bool lAvg = false;
-        vtk_xml::write_vtus(simulation, solutions, lAvg);
-      } else {
-        output::output_result(simulation, com_mod.timeP, 2, iEqOld);
-      }
-
-    } else {
-      output::output_result(simulation, com_mod.timeP, 2, iEqOld);
+    if (save_vtu) {
+      vtk_xml::write_vtus(simulation, solutions, /* lAvg = */ false);
     }
 
     // [NOTE] Not implemented.
@@ -536,7 +523,7 @@ void iterate_solution(Simulation* simulation)
       }
 
       if (cm.mas(cm_mod)) {
-        if (l2 && l3) {
+        if (save_vtu) {
           uris::uris_write_vtus(com_mod);
         }
       }
--- a/Code/Source/solver/main.cpp
+++ b/Code/Source/solver/main.cpp
@@ -251,9 +251,6 @@ void iterate_solution(Simulation* simulation)
   auto& Yn = solutions.current.get_velocity();     // New variables (velocity)
   auto& Dn = solutions.current.get_displacement(); // New integrated variables (displacement)
 
-  bool l1 = false;
-  bool l2 = false;
-
   #ifdef debug_iterate_solution
   dmsg << "Start Outer Loop ..." << std::endl;
   #endif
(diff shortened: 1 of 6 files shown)


Code/Source/solver/main.cpp
```diff
@@ -253,7 +253,6 @@ void iterate_solution(Simulation* simulation)
 
   bool l1 = false;
   bool l2 = false;
```
kko27: Is there a possibility of also replacing these boolean flags with more meaningful variable names, or removing them? Not sure if that increases the scope of this PR.

Code/Source/solver/output.cpp
```diff
@@ -7,71 +7,108 @@
...
-/// \todo NOTE: This is not fully implemented.
-//
-void output_result(Simulation* simulation,  std::array<double,3>& timeP, const int co, const int iEq)
-{
-  #ifdef debug_output_result
+namespace {
+// Field widths of the history table. The rows and the header use the same
+// widths, so that they are vertically aligned. A value wider than its field
+// shifts the rest of the row rather than being truncated.
+constexpr int eq_width        = 2;  // equation symbol
```
kko27, line 22: Minor comment: do you think that in the future, we will stick to a 2-character convention for all equations being solved? Just wondering for aesthetic reasons.

Code/Source/solver/output.cpp
```diff
@@ -7,71 +7,108 @@
...
+  }
+}
+
+void output_result(const Simulation *simulation, std::array<double, 3> &timeP,
+                   const bool save_results, const int iEq) {
+#ifdef debug_output_result
   DebugMsg dmsg(__func__, com_mod.cm.idcm());
   dmsg.banner();
-  #endif
+#endif
```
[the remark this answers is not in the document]
zasexton: I agree with this comment even though this is technically a separate issue that was already in the code.

zasexton (approved):
This looks good to me

kko27 (commented):
Thanks so much @michelebucelli! The changes look great! Have two minor comments.

Code/Source/solver/output.cpp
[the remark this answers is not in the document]
michelebucelli: Fixed by mving these lines below the block of declarations that follows, so that `com_mod` is well defined.

Code/Source/solver/output.cpp
michelebucelli: One of the advantages of having this as a variable is that, should we need more characters in the future, it would be as easy as changing this `2` to some other number.

More in general, I personally prefer writing things in full rather than using acronyms, because the acronyms assume some more knowledge on the reader, and may leave some ambiguity. For example, `ST` is used for both `struct` and `ustruct`, I think, and some other abbreviations are not immediately obvious to me (e.g. `MS`, out of context, would not bring mesh motion or mesh smoothing to my mind). But then of course there's a trade-off between verbosity and precision, and where to fall between them is largely a matter of preference.

Code/Source/solver/output.cpp
```diff
@@ -7,71 +7,108 @@
 #include "output.h"
 #include "utils.h"
 
+#include <cstdio>
+#include <iomanip>
 #include <math.h>
+#include <sstream>
```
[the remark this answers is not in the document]
michelebucelli: Fixed by adding the missing header.

Code/Source/solver/main.cpp
michelebucelli: Done, as follows:

1. `l1` was only used in a statement like `l1 = expression_A; if (expression_B && l1) { ... }`. I removed the variable `l1` entirely, and replaced that statement with `if (expression_B && expression_A) { ... }`. Since the two expressions were slightly obscure due to the variables involved having somewhat cryptic names (to me at least), I included comments to explain the meaning of both (I'd rather not go on a variable renaming spree at this time 😅 ).
2. `l2` was used similarly, although its value was printed by a debug statement as well so it couldn't be removed without changing behavior. I moved its declaration down to where it is used, and gave it the more explicit name `is_restart_output_step`.
3. There was another `l1` variable, declared in the `while` loop (and thus hiding the `l1` variable here). I also renamed it to `reached_stop_time_step`.

Conceivably there's a bit more cleanup to be done in the main function code along these lines (making variables as local as possible, giving more expressive names, removing unused variables), but maybe we can save it for a more scoped issue.

kko27: @michelebucelli, thanks for doing this! These changes look good to me

kko27 (approved):
Looks good!

PR #627, opened 2026-08-28
Add standalone reference tests for cellular ODE models

### LINKED ISSUE #597: Verification and testing of cellular ODE-based models
## Problem description

Both ionic models and active stress models are based on solving the same system of ODEs at every point in the domain. These systems are usually relatively complex, due to having many equations and involved expressions, so that making typos in their implementation is not uncommon.

As of now, the correctness of their implementation is only tested through regression tests, which verify that the solution of a 3D test case is consistent with a reference solution that was computed with the same code at the time of first implementation. This, however, has a few problems, in my opinion:
1. the comparison against the reference solution does not exclude that the reference solution is itself wrong (see e.g. #535, where a typo in the implementation of the Aliev-Panfilov had gone unnoticed due to it being present when the reference solution was generated);
2. the regression tests usually involve less-than-trivial setups (e.g. the electrophysiology tests couple the ODE system with the monodomain PDE), so that if the test fails it is not immediately obvious whether the issue is in the ODE or in the PDE or elsewhere;
3. regression tests are usually 3D, which makes them relatively costly and slows down the CI/CD pipelines.

One alternative way of verifying the correctness would be to solve the ODE model as a standalone, and then compare its solution to that obtained by a trusted third-party implementation (cellular model developers often provide a reference implementation themselves). However, `svMultiPhysics` currently doesn't have a standalone ODE solver, as far as I can tell.

## Proposed solution

I propose adding tests to the `unitTests` folder that implement a standalone 0D solver for both `IonicModel` and `ActiveStress` (or `ActiveStressODE`). These tests would take care of providing meaningful input to the ODE models surrogating coupling with 3D models (e.g. by solving an additional ODE to compute the membrane potential). The result would be compared against a reference solution provided e.g. as a CSV file (or other format).

The base classes `IonicModel` and `ActiveStress` should already support this thanks to the local timestepping functions (which might need to be moved to the public scope, or made available through friendship).

## Additional context

1. @samibismar already did something of this kind as part of #596, although lacking the infrastructure that part was not committed anywhere. This proposed change would support more systematic testing.
2. I suspect that some of this overlaps with what @zasexton presented a while ago on auxiliary models (for the same reasons that `IonicModel` and `ActiveStress` have some overlap with it). I still think it might be useful to do this before auxiliary models come to the main branch, to provide testing infrastructure to cellular models that are currently under development, with the understanding that some of this might be discarded in favor of the more principled and general implementation.

samibismar:
Resolves #597.

## Current situation

The existing regression tests for cellular ODE models use reference solutions generated by svMultiPhysics itself and are generally coupled to larger 3D problems.

This PR adds standalone unit tests for ionic and active-stress models using independently generated reference trajectories.

## Release Notes

- Added shared trajectory-test infrastructure for `IonicModel` and `ActiveStress`.
- Added full-state reference tests for:
  - Aliev-Panfilov
  - FitzHugh-Nagumo
  - Bueno-Orovio EPI, ENDO, and M
  - ten Tusscher-Panfilov 2006 EPI, ENDO, and M
  - Nash-Panfilov
  - Regazzoni/RDQ20-MF
- Ionic-model tests start from their initial/rest states, apply a time-dependent stimulus, and verify the resulting transient.
- Active-stress tests use prescribed calcium and mechanical inputs to verify the full model-state trajectory.
- Added reference CSVs under `tests/unitTests/reference_data/`.
- Added independent reference generators and a verifier under `tests/unitTests/reference_generators/`.

## Testing

- Full unit-test suite passes: 252/252.
- Reference-data verification passes for all 10 trajectories.


Code/Source/solver/active_stress_nash_panfilov.h
```diff
@@ -54,6 +54,17 @@ class NashPanfilov : public ActiveStressODE {
...
+    /// Set a scalar parameter by name with full double precision.
+    ///
+    /// @c double_parameters is @c protected in @ref ActiveStressModelParameters
+    /// and accessible here because this class is a derived class.
+    /// Throws @c std::out_of_range if @p label is not a registered parameter.
+    void set(const std::string &label, double value) {
+      auto &p = double_parameters.at(label);
+      p.value_    = value;
+      p.value_set_ = true;
+    }
```
michelebucelli: Might it be better to place this in some class higher up in the hierarchy, say `ActiveStressModelParameters`? Additionally, for symmetry with `ActiveStressModelParameters::get_scalar`, I'd rename this to `set_scalar`.

Code/Source/solver/active_stress_nash_panfilov.h
```diff
@@ -54,6 +54,17 @@ class NashPanfilov : public ActiveStressODE {
...
+
+    /// Set a scalar parameter by name with full double precision.
+    ///
+    /// @c double_parameters is @c protected in @ref ActiveStressModelParameters
+    /// and accessible here because this class is a derived class.
+    /// Throws @c std::out_of_range if @p label is not a registered parameter.
+    void set(const std::string &label, double value) {
+      auto &p = double_parameters.at(label);
+      p.value_    = value;
+      p.value_set_ = true;
```
michelebucelli: `at` throws an exception if the key is not found in the map, but this exception's message is quite generic and not very helpful in finding the source of the error. I suggest changing to something like this:

```suggestion
      auto it = double_parameters.find(label);
      
      svmp::check<svmp::FE::InvalidArgumentException>(
        it != double_parameters.end(),
        "Parameter " + label + " not found.");

      it->second.value_    = value;
      it->second.value_set_ = true;
```

Code/Source/solver/CMakeLists.txt
```diff
@@ -409,6 +409,14 @@ if(ENABLE_UNIT_TEST)
     pthread   # link pthread on ubuntu20
   )
 
+  # Expose the path to the unit-test reference-data directory as a compile-time
+  # constant. CMAKE_CURRENT_SOURCE_DIR is Code/Source/solver/ regardless of
+  # where cmake is invoked, so the path is stable under the ExternalProject build.
+  # Named generically so future ionic-model tests can share the same directory.
```
michelebucelli: Some of these comments don't seem to be relevant. I suggest simplifying to:

```suggestion
  # Define the path to the unit-test reference-data directory as a preprocessor
  # macro.
```

(Compile-time constants are a somewhat larger category than preprocessor macros, as they include `constexpr` declarations and some `const` declarations that the compiler can deduce to be compile-time constant. Preprocessor macro is more accurate here)

Code/Source/solver/ionic_bueno_orovio.h
```diff
@@ -58,6 +58,8 @@ class BuenoOrovio : public IonicModel {
       add_parameter("k_so", {2.04580, 2., 2.1}, required);
       add_parameter("u_so", {0.650, 0.65, 0.6}, required);
       add_parameter("tau_s1", {2.73420, 2.7342, 2.7342}, required);
+      // svMP currently uses 2 ms for the M-cell tau_s2 default;
+      // Bueno-Orovio et al. (2008), Table 1, reports 4 ms.
```
michelebucelli: I suggest rephrasing this as:
```suggestion
      // Beware that the default here for the M-cell tau_s2 (X-th entry in
      // the list below, 2 ms) is different from the value reported in
      // Bueno-Orovio et al. (2008), Table 1, which is 4 ms.
```
(replace X-th with 2nd or 3rd depending on which is the right one).

Additionally, perhaps this should be moved to the general class documentation, so that it will be exposed in the Doxygen-generated files. I would suggest placing this in a `@warning` paragraph.

tests/unitTests/active_stress_tests/active_stress_test_helpers.h
```diff
@@ -0,0 +1,363 @@
...
+#include <vector>
+
+/**
+ * @brief Run a trusted-reference trajectory test for an ActiveStress model.
+ *
+ * A test-only adapter derived from the concrete model exposes the protected
+ * node-local interface needed to run the trajectory experiment.
+ * By default, state values are read from CSV columns @c s0 through @c sN.
+ * The constructor's state-column mapping can explicitly select other columns
+ * when an existing reference file uses a different, documented layout.
```
michelebucelli: I feel that this paragraph is a bit hand-wavy, rather than precisely and concretely describe what this test does (it may or may not be AI generated: either way, behave of AI-generated prose, as it very frequently has this issue).

Proposed rephrasing:
```
 * Runs a twitch experiment on the active stress model passed as template 
 * argument, and compares the simulation result with a reference, trusted
 * external simulation loaded from a CSV file. If the simulated result
 * doesn't match the reference up to a tolerance, an exception is raised,
 * causing the test to fail.
 * 
 * The twitch experiment uses a prescribed calcium transient to trigger
 * the contraction in the active stress model (see @ref calcium_at). The
 * fiber stretch transient is also prescribed (see @ref fiber_stretch_at)
 * 
 * The active stress model is solved as standalone (i.e. only one system
 * of ODEs is solved), not in 3D. Therefore, the test only verifies the
 * correctness of the ODE definition and timestepping.
 * 
 * ### Format of the reference solution
 * 
 * The CSV file containing the reference solution is expected to have the
 * following format:
 * [...insert what the format should be...]
 * 
 * The state variables are assumed by default to have the same order as in
 * the concrete model being tested. If necessary, an optional constructor
 * argument to this class can define a different ordering.
```

tests/unitTests/active_stress_tests/active_stress_test_helpers.h
```diff
@@ -0,0 +1,363 @@
...
+ *
+ * Checkpoint indices are zero based: checkpoint @c N is compared with the
+ * state after advance @c N, which advances the interval
+ * [N * time_step, (N + 1) * time_step].
```
michelebucelli: That an array is zero-based should be the default assumption and never have to be stated. On the other hand, I think it would be better to expand a bit on how checkpointing works, for example
1. by explicitly stating that the check is only done at the checkpoints that are found in the CSV file (as opposed e.g. to requiring that the CSV file contains all and only the times at which the simulated solution is computed);
2. by explicitly stating that the mismatch is measured point-wise (as opposed to the L2 or L1 norms);
3. by explicitly stating whether the checkpoints are required to precisely fall on simulated time steps, or whether they can also fall in the interval between two time steps, and what the behavior is in that case.

tests/unitTests/active_stress_tests/active_stress_test_helpers.h
```diff
@@ -0,0 +1,363 @@
...
+ */
+template <class ConcreteModel>
+class ActiveStressTrajectoryTest {
+  static_assert(std::is_base_of_v<ActiveStress, ConcreteModel>,
+                "ConcreteModel must derive from ActiveStress");
+
+  class TestAdapter : public ConcreteModel {
```
michelebucelli, line 163: I would provide some Doxygen documentation for this class (in this case, clarifying its intent, i.e. to expose some protected members to this test).

Similarly, I suggest to have Doxygen documentation for all members and methods of the `ActiveStressTrajectoryTest` class (I give more detailed suggestions for some below).

tests/unitTests/active_stress_tests/active_stress_test_helpers.h
```diff
@@ -0,0 +1,363 @@
...
+    if (!(time_step > 0.0) || !std::isfinite(time_step))
+      throw std::invalid_argument("ActiveStress test time step must be finite and positive");
+
+    const double step_count = final_time / time_step;
+    const double rounded_step_count = std::round(step_count);
+    const double scale = std::max(1.0, std::fabs(step_count));
+    if (std::fabs(step_count - rounded_step_count) >
+        1.0e-12 * scale)
+      throw std::invalid_argument(
+          "ActiveStress test final time must be an integer multiple of the time step");
```
michelebucelli: I suggest to replace all these `throw` statements with the appropriate `svmp::raise` call. Even better, if you have something like `if (X) throw ...`, I suggest to replace that with `svmp::throw_if<...>(...)`.

tests/unitTests/active_stress_tests/active_stress_test_helpers.h
```diff
@@ -0,0 +1,363 @@
...
+
+    if (header.empty())
+      throw std::runtime_error("ActiveStress reference CSV has no header: " + path);
+    if (checkpoints.empty())
+      throw std::runtime_error("ActiveStress reference CSV has no checkpoints: " +
+                               path);
+    return checkpoints;
+  }
+
+  static double calcium_at(double time)
```
michelebucelli, line 458: The documentation for this function should report the expression of the calcium twitch that this function defines. The same applies for `fiber_stretch_at`.

Imagine someone adds a new model and wants to implement the automatic test for it. This is information that they likely need to generate the reference solution, and asking them to deduce it from the function implementation (as opposed to reading the expression in the Doxygen documentation webpage) adds friction that might make them say "forget it, I'll just skip implementing the test".

tests/unitTests/active_stress_tests/test_active_stress_nash_panfilov.cpp
```diff
@@ -0,0 +1,61 @@
...
+//         epsilon(Phi) = eps0 + (eps_inf - eps0)*exp(-exp(-xi*(Phi - Phi_bar)))
+//       Both equations are expressed in terms of transmembrane potential Phi,
+//       not calcium; see Figure 3.
+//
+// svMultiPhysics follows the Goktepe-Kuhl form but substitutes intracellular
+// calcium concentration Ca for Phi. The parameter-role mapping is
+//   eta_T        <-> k_sigma       calcium_rest <-> Phi_r
+//   xi_T         <-> xi            calcium_crit <-> Phi_bar.
+// This calcium substitution is an svMultiPhysics-specific adaptation not
+// documented in [1] or [2].
```
michelebucelli: This information should duplicate that which is already present in the documentation for the `NashPanfilov` class, does it not? If there is some information here that is not there, then it should be added there, and these comments should be removed. Similar considerations apply (to a lesser extent) to other files like this.

The test documentation, I think, should only discuss where the reference solution comes from, and point to the model class for additional details.

Also, it seems that Doxygen has a [keyword `@test`](https://www.doxygen.nl/manual/commands.html#cmdtest) for documenting test cases, which might be a good idea to use for this and similar scenarios.

tests/unitTests/ionic_model_tests/ionic_model_test_helpers.h
michelebucelli, line 1: Similar considerations apply to this file as to `active_stress_test_helpers.h`.

In addition to those, it seems to me that the two files aim at the same goal (verification of ODE models against reference solutions), and while they have a similar large-scale structure, there are some small dissimilarities (e.g. the trajectory configuration `struct` is here but not there, and similar). I suggest removing the dissimilarities as much as possible.

While not technically necessary, this has a few advantages.
1. If someone understands the testing infrastructure for one of the two, their understanding transfers immediately to the other (whereas now they could think that it does, but then find out that there are subtle differences, and this may lead them down a wrong path).
2. Future migration to a unified setting where `ActiveStress` and `IonicModel` are both under an abstract `AuxiliaryState` or `ODE` class will be easier. As of now it is not trivial to have a unified helper, due to the existing differences between `ActiveStress` and `IonicModel`, and the fact that they do not have a common base class. But I think both of these obstacles will gradually go away.

tests/unitTests/reference_data/active_stress_nash_panfilov_twitch.csv
```diff
@@ -0,0 +1,27 @@
...
+#     substitute intracellular calcium Ca for Phi, with parameter roles
+#       eta_T <-> k_sigma, calcium_rest <-> Phi_r,
+#       xi_T <-> xi, calcium_crit <-> Phi_bar.
+#     The oracle evaluates
+#       dT/dt = epsilon(Ca)*(eta_T*(Ca-calcium_rest)-T).
+# Parameters: epsilon_0=0.1, epsilon_i=1.0, xi_T=4e3, eta_T=1e2,
+#             calcium_rest=1.25e-4, calcium_crit=8e-4 (slab calibration).
+# Ca transient: double-exponential, c0=1e-4 mM, cmax=9e-4 mM,
+#               tau_rise=20ms, tau_decay=50ms, onset=10ms. dt=1ms, 200 steps.
+# Values extracted from the existing test without modification.
```
michelebucelli: I appreciate the documentation effort, but as far as I know CSV files do not have a standard comment syntax, meaning that these files might not be readable by other tools. I suggest removing these, and instead placing the information they contain in the documentation for the test case in the associated `cpp` file.

tests/unitTests/reference_generators/active_stress/nash_panfilov/generate_nash_panfilov.py
michelebucelli, line 1: I think that the generation of reference solutions should be delegated to trusted third-parties, if possible. Otherwise there's a risk that we are simply comparing two different solvers made by the same people, and both of them might contain the same errors.

I can see that this particular flavor of the Nash-Panfilov model is not the vanilla one (as in the original paper), but it is the slightly modified version from Goktepe and Kuhl, and so the CellML-generated code might not match exactly.

Still, I suggest taking that CellML code (I think it's [this one](https://preview.models.physiomeproject.org/exposures/d96a64b94d824692955e06ec878a2d09), but do double-check) and applying the modifications there, rather than rewriting the thing from scratch.

A similar suggestion applies to other files as appropriate.

tests/unitTests/reference_generators/active_stress/regazzoni/generate_regazzoni.py
michelebucelli, line 1: If I understand this right, this file is adapted, if not copied, from [this repository](https://github.com/FrancescoRegazzoni/cardiac-activation). If so, the license statement at the top is not correct (the copyright of at least part of this file is of the original author, not ours).

Having said that, I think for this case it is sufficient to document (e.g. in the README and in the Doxygen documentation) the repository that was used, and the commit hash that was used, rather than including the ready-to-use generator script. I fear that the Python-building-C++-code setup might be quite frail if the system configuration changes (e.g. compilers not found, linking issues, etc.), and it would be an unnecessary hassle to make it more robust.

tests/unitTests/reference_generators/verify_reference_data.py
```diff
@@ -0,0 +1,267 @@
+#!/usr/bin/env python3
+# SPDX-FileCopyrightText: Copyright (c) Stanford University, The Regents of the
+# University of California, and others. SPDX-License-Identifier: BSD-3-Clause
+
+"""Regenerate and verify all unit-test reference trajectories."""
```
michelebucelli: What is the verification doing here? I suggest expanding the documentation to explain this more clearly.

tests/unitTests/reference_generators/active_stress/nash_panfilov/generate_nash_panfilov.py
michelebucelli: Also beware that the copyright of CellML-generated code probably still lies with the Cell-ML authors, and so the generated scripts should not contain the header about copyright being held by Stanford University etc.

The CellML model page (at least on the new website) has some licensing and authorship information, and that should be followed in the scripts I think.

michelebucelli (commented):
Thank you @samibismar! This is a very useful contribution. I have left some comments below.

tests/unitTests/reference_generators/active_stress/nash_panfilov/generate_nash_panfilov.py
samibismar: Thanks you! I looked into the other reference generators before changing anything. The same general issue applies to all of them: the current generators run independently of svMultiPhysics, but the model equations were still transcribed locally.

I found the following external implementations and the minimum adaptations that would be needed:

| Model | Trusted external base | Minimal adaptations needed | Plan |
|---|---|---|---|
| **Nash–Panfilov** | [Physiome Nash–Panfilov 2004 CellML](https://models.physiomeproject.org/exposure/d96a64b94d824692955e06ec878a2d09/nash_panfilov_2004.cellml/source_text) | Replace the original piecewise, voltage-driven active-tension law with the Göktepe–Kuhl Gompertz form used by svMP; replace voltage activation with intracellular calcium; apply the existing svMP calibration and FE test protocol. | **Use the CellML-generated implementation as the trusted starting point and make only the explicit changes needed to reproduce the svMP Nash variant.** |
| **FitzHugh–Nagumo** | [Published Physiome FitzHugh 1961 CellML](https://models.physiomeproject.org/exposure/cf32346a9e5c4b2cdb559b11da5f1ae1/fitzhugh_1961.cellml/source_text) | The equations already match after the parameter/time transformation. Apply the svMP parameters and time scaling, then use the existing FE test protocol. | **Replace the handwritten FN equations with the CellML-generated equations and keep only the svMP-specific test protocol locally.** |
| **TP06 EPI/ENDO/M** | Physiome TP06 CellML: [EPI](https://models.physiomeproject.org/exposure/a7179d94365ff0c9c0e6eb7c6a787d3d/ten_tusscher_model_2006_IK1Ko_epi_units.cellml/source_text) / [ENDO](https://models.physiomeproject.org/exposure/a7179d94365ff0c9c0e6eb7c6a787d3d/ten_tusscher_model_2006_IK1Ko_endo_units.cellml/source_text) / [M](https://models.physiomeproject.org/exposure/a7179d94365ff0c9c0e6eb7c6a787d3d/ten_tusscher_model_2006_IK1Ko_M_units.cellml/source_text) | Keep the generated TP06 equations and phenotype parameters, but use the test’s FE/Rush–Larsen stepping, stimulus convention, phenotype initial states, and state ordering. | **Replace the handwritten TP06 equations with the CellML-generated equations while keeping the svMP-specific numerical/test protocol.** |
| **Aliev–Panfilov** | [Pigreads `aliev1996simple.cl`](https://gitlab.com/pigreads/pigreads/-/blob/43346f8b429ec7d9a09bf94d313272531259ac77/models/aliev1996simple.cl) | Adapt the original AP formulation to the Göktepe–Kuhl/svMP form: separate the `alpha` and `b` parameter roles, apply the svMP voltage/time scaling, initial state, and stimulus convention, remove diffusion, and use the existing FE test protocol. | **Use Pigreads as the trusted external starting implementation if you think it is an appropriate source.** |
| **Bueno–Orovio EPI/ENDO/M** | [openCARP `Bueno.model`](https://git.opencarp.org/openCARP/openCARP/-/blob/ac4e96792db082958fa9a830341f6e0642cc6bb8/physics/limpet/models/Bueno.model) | The equations already match closely. Apply the EPI/ENDO/M parameter sets, intentionally retain svMP’s M-cell `tau_s2 = 2 ms` value instead of the published `4 ms`, and use the existing FE test protocol. | **Use the pinned openCARP implementation as the trusted external oracle, but keep openCARP external rather than copying its implementation into svMultiPhysics.** |

I also checked the source licensing/authorship because of your follow-up comment. The Physiome sources have explicit attribution requirements, Pigreads is MIT-licensed, and openCARP uses a more restrictive academic license.

Does the Nash/FN/TP06 plan look reasonable to you?

For **Aliev–Panfilov**, do you think [Pigreads](https://gitlab.com/pigreads/pigreads/-/blob/43346f8b429ec7d9a09bf94d313272531259ac77/models/aliev1996simple.cl) is a sufficiently trustworthy external source for this purpose? I also found the [Physiome AP workspace](https://models.physiomeproject.org/workspace/241), but it is a poorer match to the svMP formulation and would require more adaptation.

For **Bueno–Orovio**, [openCARP](https://git.opencarp.org/openCARP/openCARP/-/blob/ac4e96792db082958fa9a830341f6e0642cc6bb8/physics/limpet/models/Bueno.model) is a much closer match than the older [Physiome bueno_2007 model](https://models.physiomeproject.org/workspace/bueno_2007), but its [license](https://opencarp.org/download/license) is more restrictive. Would you prefer that I handle it more like Regazzoni—document the exact external source/version used to generate the committed reference trajectory rather than include/adapt the implementation inside svMultiPhysics—or use another approach you think would be better?

michelebucelli (approved):

javijv4 (commented):
@samibismar, this is very useful thanks for working on it. I went through the code and looks good to me! Just a quick request, the `README.md` in `tests/unitTests/` hasn't been updated and does not mention these new tests. Can you add a section briefly explaining the new tests and how to run them?

javijv4 (approved):

PR #630, opened 2026-08-28
Fix GMRES early termination result

### LINKED ISSUE #623: GMRES returns residual as solution if it terminates with zero iterations
### Problem description

`gmres_s` and `gmres_v` use the argument `Vector<double> &R` as both input (the system's right hand side vector) and output (the system's solution upon convergence). If the functions reach the `for` loop for the actual GMRES iterations, once that loop is over, they set `R = X`, and everything is fine.

However, if the termination-with-no-iterations condition is triggered (that is, if the initial guess for the solution is already within the tolerance), the assignment `R = X` is never executed, and the functions just return.

The caller expects `R` to contain the solution, but it still contains the right hand side instead, which is incorrect.

### Solution

Two possible fixes:
1. set `R = X` before returning in the early termination branch;
2. make `gmres_s` and `gmres_v` behave the same as `gmres`, by taking separate arguments for `R` (input) and `X` (output).

### Additional context

1. This doesn't destroy the solver because the correct solution would be zero (we're always solving for the increment between nonlinear iterations) and, if the early termination condition is true, the residual `R` will be small.
2. This explains a behavior I've observed a few times where the linear solver would do zero iterations but the residual reported by the nonlinear solver would keep changing. One example was reported by @kko27 in #606 (see the convergence history from [this comment](https://github.com/SimVascular/svMultiPhysics/issues/606#issuecomment-5420867672)).

michelebucelli:
Fixes #623.

## Current situation

`gmres_s` and `gmres_v` use the argument `Vector<double> &R` as both input (the system's right hand side vector) and output (the system's solution upon convergence). If the functions reach the for loop for the actual GMRES iterations, once that loop is over, they set `R = X`, and everything is fine.

However, if the termination-with-no-iterations condition is triggered (that is, if the initial guess for the solution is already within the tolerance), the assignment `R = X` is never executed, and the functions just return.

The caller expects `R` to contain the solution, but it still contains the right hand side instead, which is incorrect.

## Release Notes 

1. Fixes the issue by setting `R = X` before the early return in `gmres_s` and `gmres_v`.
2. Regenerates reference solution for the test `fluid/driven_cavity_2d_porous`, which exhibited solution drift during nonlinear iterations with zero linear iterations (see [this comment](https://github.com/SimVascular/svMultiPhysics/issues/623#issuecomment-5454200428) for details).

## Documentation

There are no user-facing changes.

## Testing

The test suite passes (up to changing the reference solution of one test as per release note 2 above).

### DIFF
--- a/Code/Source/linear_solver/gmres.cpp
+++ b/Code/Source/linear_solver/gmres.cpp
@@ -340,6 +340,7 @@ void gmres_s(fsi_linear_solver::FSILS_lhsType& lhs, fsi_linear_solver::FSILS_sub
     ls.callD = std::numeric_limits<double>::epsilon();
     ls.dB = 0.0;
     ls.success = true;
+    R = X;
     return; 
   }
 
@@ -494,6 +495,7 @@ void gmres_v(fsi_linear_solver::FSILS_lhsType& lhs, fsi_linear_solver::FSILS_sub
     ls.callD = std::numeric_limits<double>::epsilon();
     ls.dB = 0.0;
     ls.success = true;
+    R = X;
     return; 
   }
 
--- a/Code/Source/linear_solver/gmres.cpp
+++ b/Code/Source/linear_solver/gmres.cpp
@@ -26,9 +26,8 @@ namespace gmres {
 
 namespace {
 
-void bc_pre(fsi_linear_solver::FSILS_lhsType& lhs, fsi_linear_solver::FSILS_subLsType& ls, const int dof,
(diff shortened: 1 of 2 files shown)


zasexton (approved):
Simple fix. `X` is zero-initialized, and assigning it to `R` matches the normal return contract. Fine to merge

kko27 (approved):
Looks good! Thanks for fixing this!

PR #631, opened 2026-08-29
Improvements to VTU output

### LINKED ISSUE #618: Improvements to VTU output
### Problem

1. VTU output files do not carry any time information, but only their timestep number (inferred from their file name). Therefore, to properly interpret the data in the VTU file, one needs to manually keep track of the time step size used for that simulation. Additionally, Paraview does not use the timestep indices from filenames, but rescales everything to 0, 1, 2, ..., so that also the output frequency is not immediately accessible in the viewer.
2. The initial condition is never exported. Since Paraview indexes file series starting at 0, this causes a shift between what the simulation computes and what Paraview reports (i.e. the first exported file corresponds to timestep 1, but is shown as 0 in Paraview, and so on).
3. Only the first exported frame contains the `DOMAIN_ID` field. I imagine this is useful to keep the exported files small. However, in some situations it is useful to filter on the `DOMAIN_ID` also on frames after the first. This can be obtained with some processing (e.g. using the `Append attributes Paraview filter), but it might be nice to also have it natively from the output.

### Solution

I propose the following solutions to the above points.

1. Adding a `TimeValue` field data array (i.e. a scalar) to the output files. This is the standard way VTK communicates time information, and it will ensure that Paraview visualizes both the timestep index and the physical time which that timestep is associated to.
2. Adding a call to `write_vtus` before the time loop starts (if not restarting and  `Start_saving_after_time_step == 0`). Initial output can still be disabled by setting `Start_saving_after_time_step` to any value greater than 0.
3. Add an option, disabled by default, to export the `DOMAIN_ID` at every time step.

While not particularly critical, I think these improvements will make the inspection of `svmp` solutions in viewers more seamless and less error prone.

### Additional context

_No response_

### Code of Conduct

michelebucelli:
Fixes #618.

## Current situation

1. VTU output files do not carry any time information, but only their timestep number (inferred from their file name). Therefore, to properly interpret the data in the VTU file, one needs to manually keep track of the time step size used for that simulation. Additionally, Paraview does not use the timestep indices from filenames, but rescales everything to 0, 1, 2, ..., so that also the output frequency is not immediately accessible in the viewer.
2. The initial condition is never exported. Since Paraview indexes file series starting at 0, this causes a shift between what the simulation computes and what Paraview reports (i.e. the first exported file corresponds to timestep 1, but is shown as 0 in Paraview, and so on).
3. Only the first exported frame contains the DOMAIN_ID field.

### Major

1. Exported VTU files now contain a `TimeValue` field attribute containing the time, which viewers such as ParaView use to properly show the physical time in the UI.
2. If `Start_saving_after_time_step == 0`, then the initial condition is exported to `result_000.vtu`. This improves the alignment between timestep indices in the simulation and viewers (on top of allowing to view the initial solution). This can be disabled by setting `Start_saving_after_time_step` to anything strictly greater than 0.
3. A new parameter `General_simulation_parameters/Save_domain_ID_in_every_file` toggles exporting the `Domain_ID` field to every timestep, instead of just the first. It is disabled by default (meaning that the default is the same as the current state). If enabled, it allows e.g. thresholding the simulation results based on subdomains.

### Minor

1. `Increment_in_saving_VTK_files` and `Increment_in_saving_restart_files` defaulted to the invalid value 0, and there was no check that they are `>= 1`. I changed the default value to `1`, and introduced those assertions.
2. Removed the unused variable `ComMod::sepOutput`.

## Documentation

Newly added or modified variables and functions have been given Doxygen documentation.

## Testing

Automatic tests all pass after the changes.

### DIFF
--- a/Code/Source/solver/VtkData.cpp
+++ b/Code/Source/solver/VtkData.cpp
@@ -42,7 +42,7 @@ void VtkData::read_file(const std::string &file_name) {
                                            "The file has no elements.");
 
     // Get the cell type.
-    auto cell = vtkGenericCell::New();
+    auto cell = vtkSmartPointer<vtkGenericCell>::New();
     vtk_data->GetCell(0, cell);
     num_points_per_elem_ = cell->GetNumberOfPoints();
     elem_type_ = cell->GetCellType();
--- a/Code/Source/solver/VtkData.cpp
+++ b/Code/Source/solver/VtkData.cpp
@@ -327,8 +327,7 @@ void VtkData::copy_points(Array<double> &points) const {
 
 void VtkData::copy_point_data(const std::string &data_name,
                               Array<double> &mesh_data) const {
-  const auto vtk_array = vtkDoubleArray::SafeDownCast(
-      vtk_data->GetPointData()->GetArray(data_name.c_str()));
+  const auto vtk_array = vtk_data->GetPointData()->GetArray(data_name.c_str());
   svmp::check<svmp::FE::InvalidArgumentException>(
       vtk_array != nullptr,
       "There is no double-valued point data array named '" + data_name +
@@ -405,8 +404,7 @@ void VtkData::copy_point_data(const std::string &data_name,
 
 void VtkData::copy_cell_data(const std::string &data_name,
                              Array<double> &mesh_data) const {
-  const auto vtk_array = vtkDoubleArray::SafeDownCast(
-      vtk_data->GetCellData()->GetArray(data_name.c_str()));
+  const auto vtk_array = vtk_data->GetCellData()->GetArray(data_name.c_str());
   svmp::check<svmp::FE::InvalidArgumentException>(
       vtk_array != nullptr,
       "There is no double-valued element data array named '" + data_name +
--- a/Code/Source/solver/VtkData.cpp
+++ b/Code/Source/solver/VtkData.cpp
@@ -3,14 +3,17 @@
 
 #include "VtkData.h"
 
+#include "Core/Exception.h"
+#include "FE/Common/FEException.h"
+
 #include <cstring>
-#include <stdexcept>
 #include <string>
 
 #include <vtkCellArray.h>
 #include <vtkCellData.h>
 #include <vtkCellType.h>
 #include <vtkDoubleArray.h>
+#include <vtkErrorCode.h>
 #include <vtkFieldData.h>
 #include <vtkGenericCell.h>
 #include <vtkIntArray.h>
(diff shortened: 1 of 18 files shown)


Code/Source/solver/VtkData.h
michelebucelli, line 1: It seems to me that the classes in this file (`VtkData`, `VtkVtpData`, `VtkVtuData`) could be unified by exploiting polymorphism of `VTK` classes, or at least substantial duplication could be removed.

For example, `vtkPolyData` and `vtkUnstructuredGrid` both derive from `vtkPointSet`, and these classes mostly use the `vtkPointSet` interface. Accordingly, most of the methods could be defined in the base class, and only the truly VTP- or VTU-specific functionality would need to be overridden in the derived classes.

Doing this unification would make additions such as this one even simpler in the future.

mrp089 (changes requested):
Thank you, @michelebucelli! Looks good. Only idea I would have is add `TimeValue` to some of the integration tests as a check.

Code/Source/solver/VtkData.h
michelebucelli: I went ahead and implemented this unification in the last few commits.

The main changes are as follows:
1. `VtkData` now stores a pointer to `vtkPointSet` (a common ancestor to `vtkPolyData` and `vtkUnstructuredGrid`), and most of the logic works through that pointer and is implemented in `VtkData` directly;
2. `VtkVtpData` and `VtkVtuData` only implement custom reading and writing, which is tightly related to the specific file format, plus a few internal helpers; the amount of code duplication in the files has reduced considerably;
3. all these classes consistently use exceptions from `svmp` namespace instead of `throw std::runtime_error`;
4. added some correctness checks, with appropriate `svmp` exceptions, that were not in the previous implementation;
5. all classes, members and methods now have Doxygen documentation.

The files are still somewhat large, but not as much as before, and code duplication is almost entirely absent now. All the tests pass as before (and they exercise both reading and writing of VTP and VTU files).

These changes are a bit more invasive than the others, so if you think it's too much for a single PR let me know and I can revert them.

Code/Source/solver/VtkData.cpp
```diff
@@ -2,814 +2,498 @@
...
-  int num_comp = data.nrows();
+void VtkData::set_point_data(const std::string &data_name,
+                             const Vector<int> &data) {
+  const int num_vals = data.size();
+
+  svmp::check<svmp::FE::InvalidArgumentException>(
+      num_vals == vtk_data->GetNumberOfPoints(),
+      "The point data array named '" + data_name + "' holds " +
+          std::to_string(num_vals) + " values, while the mesh has " +
+          std::to_string(vtk_data->GetNumberOfPoints()) + " points.");
```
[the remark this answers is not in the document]
michelebucelli: This comment seems to be triggered by the fact that, in the function `write_vtp` from `vtk_xml.cpp`, there's a call `vtk_writer->set_point_data("GlobalElementID", lFa.gE)`, where a cell data array is incorrectly set by calling `set_point_data`.

I fixed this by introducing another overload of `set_element_data` for `Vector<int>` and calling that instead.

It seems that this is only triggered by remeshing, which is not covered by the test suite, and that's probably why this was never caught.

mrp089 (approved):

ktbolt (approved):
Looks good !

Code/Source/solver/VtkData.h
ktbolt: @michelebucelli Nice fix !

Issue #634, opened 2026-09-01
Shape function gradients are incorrectly assumed to be constant within wedge elements


dseyler:
### Description

In `Code/Source/solver/nn_elem_props.h`, the flag `mshType::lshpF` is set to indicate whether shape functions are linear (including wedge elements). This flag is then used in several places throughout the code to decide whether the derivative of the shape functions is constant across the element or needs to be computed per Gauss point. This is true for `LIN1`, `TRI3`, and `TET4` elements but not for `WDG` elements.

```cpp
for (int g = 0; g < lM.nG; g++) {
  if (g == 0 || !lM.lShpF) {
    nn::gnn(eNoN, nsd, nsd, Nx_g, xl, Nx, Jac, ksix);   // Nx and Jac
  }
  double w = lM.w(g) * Jac;
  ...
}
```

This bug appears 35 times in the codebase, and there are currently no wedge element test cases to catch it.

| file | line number |
|---|---|
| `post.cpp` | 263, 433, 581, 678, 762, 921, 1775, 2060 |
| `fluid.cpp` | 634, 642, 706, 715 |
| `fsi.cpp` | 181, 189, 281, 290 |
| `all_fun.cpp` | 390, 589, 656 |
| `stokes.cpp` | 137, 169, 178 |
| `ustruct.cpp` | 330, 363, 371 |
| `cep.cpp` | 490 |
| `cmm.cpp` | 876 |
| `eq_assem.cpp` | 248 |
| `heatf.cpp` | 112 |
| `heats.cpp` | 89 |
| `Integrator.cpp` | 1171 |
| `l_elas.cpp` | 112 |
| `mesh.cpp` | 114 |
| `set_bc.cpp` | 1416 |
| `sv_struct.cpp` | 297 |

### Reproduction

There are no wedge element test cases to reproduce this bug.

### Expected behavior

Either a new flag should be added to indicate that an element has a constant shape function gradient or these locations in the code should be gated on element type rather than `lshpF`

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct and Contributing Guidelines

ktbolt:
@dseyler This is a C++ conversion error made by yours truly; the Fortran has ` lM%lShpF   = .FALSE.`

ktbolt:
Note that we should be using wedge elements for boundary layer meshing.

dseyler:
[...]

Issue #633, opened 2026-08-31
Templated viscous stress models


dseyler:
### Problem

The Newtonian and potential viscosity models currently run very slow, accounting for nearly half of the assembly runtime for structural mechanics simulations. It looks like this slowdown is mostly due to heap allocation of several runtime-sized arrays for every gauss point that should be moved to the stack. Furthermore, the viscosity models can be templated on `nsd`, similarly for what is already done for `compute_pk2cc`. Fixing this **cuts assembly runtime by ~30%** (20% of total runtime)

### Solution

I implemented `nsd` templated viscosity models in this [branch](https://github.com/dseyler/svMultiPhysics/tree/perf/visc-stress-templated) with fixed-size Eigen arrays used in place of the previous heap-allocated arrays.

Other minor improvements include:
- The column count of dynamic arrays can be capped at 27, as this is the largest possible value for `eNoN`:
```
/// @brief Largest element node count the fixed-size views below allow (HEX27).
constexpr int MAX_ELEMENT_NODES = 27;

/// @brief An nsd x eNoN matrix whose column count is bounded at compile time,
/// so that it is held on the stack rather than heap allocated.
template <int nsd>
using MatNodes = Eigen::Matrix<double, nsd, Eigen::Dynamic, 0, nsd, MAX_ELEMENT_NODES>;
```
- `nsd` templated overloads of `mat_dev` and `mat_symm`

```
template <int nsd>
Matrix<nsd> mat_symm(const Matrix<nsd>& A) {
    return 0.5 * (A + A.transpose());
}

template <int nsd>
Matrix<nsd> mat_dev(const Matrix<nsd>& A) {
    return A - (A.trace() / nsd) * Matrix<nsd>::Identity();
}
```

Below is the proposed potential viscosity implementation (Newtonian viscosity follows a similar pattern):

```
template <int nsd>
void compute_visc_stress_potential_impl(const double mu, const int eNoN, const Array<double>& Nx,
                        const Array<double>& vx, const Array<double>& F,
                        Array<double>& Svis, Array3<double>& Kvis_u, Array3<double>& Kvis_v) {
    using MatNsd = mat_fun::Matrix<nsd>;

    // Initialize Svis, Kvis_u, Kvis_v to zero
    Svis = 0.0;
    Kvis_u = 0.0;
    Kvis_v = 0.0;

    Eigen::Map<const MatNsd> F_map(F.data());
    Eigen::Map<const MatNsd> vx_map(vx.data());
[...]

PR #640, opened 2026-09-04
Templated viscous stress functions + linear shape function fix 

dseyler:
## Current situation
Address issues #633 and #634, which involve templating viscosity models on `nsd` and skipping unnecessary computation when shape function gradients are constant across an element. These changes reduce assembly runtime by ~30% and fix a bug where shape function gradients were assumed to be uniform within wedge elements.

## Release Notes 
- `compute_visc_stress_potential` are now templated on `nsd` in `mat_models.cpp`
- Eigen views are implemented throughout viscosity models to keep intermediate array computations on the stack and condense several for-loops into one-line expressions. 
- `compute_visc_stress_and_tangent` dispatches to templated models by `nsd`
- `mat_symm` and `mat_dev` are also `nsd` templated in `mat_fun.h`
- `Svis`, `Kvis_u`, and `Kvis_v` are allocated outside of the Gauss point loop, written over for each Gauss point, and only zeroed when no viscosity model is defined for an element (sv_struct.cpp, ustruct.cpp, fsi.cpp) or when viscosity arrays can be reused.
- `recompute_visc` flag is added to indicate when viscosity computation can be skipped (shape function gradient are uniform for all Gauss points within an element, so `Kvis_u`, `Kvis_v`, and `Svis` values can be reused).
- `compute_visc_stress_and_tangent` skips viscosity computation when `recompute_visc == False`
- default case added to `compute_visc_stress_and_tangent` so viscosity arrays are zeroed and do not carry the previous element's values when an element doesn't have a viscosity model defined
- `struct_2d`, `struct_3d`, `ustruct_2d_m`, `ustruct_3d_m` now require four new arguments: `Svis`, `Kvis_u`, `Kvis_v`, and `recompute_visc`
- `compute_visc_stress_and_tangent` now takes `recompute_visc`
- Fixed bug where `lShpF` should be False for wedge elements.

## Testing
All test cases pass. Eigen operations were verified in isolation to be within machine precision of the operations they replaced.


aabrown100-git (approved):
Approving!

Code/Source/solver/mat_models.cpp
```diff
@@ -1641,71 +1638,58 @@ void compute_visc_stress_potential(const double mu, const int eNoN, const Array<
...
+    Eigen::Map<const Matrix<nsd>> vx_map(vx.data());
+    Eigen::Map<const NodalMatrix<nsd>> Nx_map(Nx.data(), nsd, eNoN);
 
     // Get identity matrix, Jacobian, and F^-1
-    auto Idm = mat_id(nsd);
-    auto J = mat_det(F, nsd);
-    auto Fi = mat_inv(F, nsd);
+    const auto Idm = Matrix<nsd>::Identity();
+    const double J = F_map.determinant();
+    const Matrix<nsd> Fi = F_map.inverse();
```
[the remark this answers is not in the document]
dseyler: I think this is fine, since a singular Jacobian would already be caught upstream in `struct3d`, so this guard would never be reached anyway

Code/Source/solver/fsi.cpp
```diff
@@ -186,7 +191,10 @@ void construct_fsi(ComMod& com_mod, CepMod& cep_mod, const mshType& lM, const So
         }
       }
 
-      if (g == 0 || !fs_1[0].lShpF) {
+      // Viscosity is constant at all Gauss points for linear elements
```
michelebucelli, line 194: Minor suggestion:
```suggestion
      // Viscosity is constant at all Gauss points for linear tetrahedral elements
```
Linear hexahedral elements are actually bi- or trilinear, so I don't think that this would apply to them, and I think it's a good idea to leave a reminder in the comment.

(I know that technically those are not linear elements, but they are sometime called that, so it might be better to be redundantly explicit just in case).

Code/Source/solver/mat_fun.h
```diff
@@ -169,6 +169,35 @@ namespace mat_fun {
...
 
     Array<double> mat_symm(const Array<double>& A, const int nd);
+
+    /**
+     * @brief Symmetric part of a 2nd order tensor, 0.5 * (A + A^T).
+     *
+     * Fixed-size overload for the Eigen matrices used by the element kernels.
+     *
+     * @tparam nsd Number of spatial dimensions.
+     * @param[in] A Second order tensor.
```
michelebucelli, line 179: Very minor:
```suggestion
     * @param[in] A second order tensor.
```
Same for the function below.

Code/Source/solver/mat_models.cpp
```diff
@@ -1559,65 +1559,62 @@ void g_vol_pen(const ComMod& com_mod, const dmnType& lDmn, const double p,
   }
 }
 
+namespace {
+
+/// @brief Largest element node count the fixed-size views below allow (HEX27).
+constexpr int MAX_ELEMENT_NODES = 27;
```
michelebucelli, line 1565: I'm not sure I have a clean solution to this, but: this number depends on the "largest" local finite element basis supported by the library, and this information is not logically pertinent to this module (the material model evaluation) but to the finite element basis module.

I think having this constant here introduces possible unexpected bugs (e.g. if larger finite element spaces are added in the future), and in general introduces implicit coupling between the two modules (implicit in the sense that the coupling is somewhat hidden).

One somewhat cleaner solution to this, I think, would be to move this definition to the basis function module (`source/Code/FE/Basis`). I am not sure what the appropriate file for that would be, but maybe @zasexton has a clearer picture.

Code/Source/solver/mat_models.cpp
```diff
@@ -1559,65 +1559,62 @@ void g_vol_pen(const ComMod& com_mod, const dmnType& lDmn, const double p,
...
-
+template <int nsd>
+void compute_visc_stress_potential(const double mu, const int eNoN, const Array<double>& Nx,
+                           const Array<double>& vx, const Array<double>& F,
+                           Array<double>& Svis, Array3<double>& Kvis_u, Array3<double>& Kvis_v) {
+    // Alias the caller's storage; no copies. Svis, Kvis_u and Kvis_v are
+    // written in full below, so they are not zeroed first.
+    Eigen::Map<const Matrix<nsd>> F_map(F.data());
+    Eigen::Map<const Matrix<nsd>> vx_map(vx.data());
+    Eigen::Map<const NodalMatrix<nsd>> Nx_map(Nx.data(), nsd, eNoN);
```
michelebucelli, line 1603: If I understand this right, these lines are constructing `Eigen` views over the data stored by the arrays, so that the subsequent operations use `Eigen` for efficiency but don't do any copy.

If that is right,
1. Would it make sense to change the code in the caller to directly use `Eigen` structures for the local data, instead of custom `Array`s? This would be a somewhat larger change perhaps, but I think it might make sense, reduce the (presumably very small, if even detectable) overhead associated to the construction of these views, and make the code a bit simpler (this function would just compute the viscosity, without any boilerplate code related to the technicalities of conversion). 
2. Should you choose not to follow the above suggestion, I would recommend writing a small helper function to do this, as a form of synctactic sugar, to turn these lines into e.g. `auto F_eigen = to_eigen_map(F)`.

A similar suggestion applies to `compute_visc_stress_newton`.

Code/Source/solver/mat_models.cpp
```diff
@@ -1559,65 +1559,62 @@ void g_vol_pen(const ComMod& com_mod, const dmnType& lDmn, const double p,
...
-
-    // Initialize Svis, Kvis_u, Kvis_v to zero
-    Svis = 0.0;
-    Kvis_u = 0.0;
-    Kvis_v = 0.0;
-
+template <int nsd>
+void compute_visc_stress_potential(const double mu, const int eNoN, const Array<double>& Nx,
+                           const Array<double>& vx, const Array<double>& F,
+                           Array<double>& Svis, Array3<double>& Kvis_u, Array3<double>& Kvis_v) {
```
michelebucelli, line 1598: If I interpret this right, you removed the definition of this function and of `compute_visc_stress_newton` from the header because they are not meant to be called from outside this file.

If that is correct, I think it is a good decision, but I would also place both functions in the anonymous namespace above (my understanding of anonymous namespaces is that, among other things, they're a way of specifying that a certain name should only have internal visibility).

Code/Source/solver/mat_models.cpp
```diff
@@ -1749,15 +1731,39 @@ void compute_visc_stress_newtonian(const double mu, const int eNoN, const Array<
...
+                                 Array<double>& Svis, Array3<double>& Kvis_u, Array3<double>& Kvis_v,
+                                 const bool recompute_visc) {
 
     switch (lDmn.solid_visc.viscType) {
       case consts::SolidViscosityModelType::viscType_Newtonian:
-        compute_visc_stress_newtonian(lDmn.solid_visc.mu, eNoN, Nx, vx, F, Svis, Kvis_u, Kvis_v);
+        // Viscosity is constant at all Gauss points for linear elements
+        if (!recompute_visc) {
+          return;
+        }
```
michelebucelli, line 1742: I think that it would be better if this check was performed by the caller of this function, rather than by the function itself.

In other words, I think this function should always do what its name says (compute the viscosity), and the caller, who is aware of the context, should be the one to decide whether the viscosity needs to be recomputed (and thus the function needs to be called) or not.

Code/Source/solver/sv_struct.cpp
```diff
@@ -229,6 +229,9 @@ void construct_dsolid(ComMod& com_mod, CepMod& cep_mod, const mshType& lM, const
                 bfl(nsd,eNoN), fN(nsd,nFn), pS0l(nsymd,eNoN), Nx(nsd,eNoN), lR(dof,eNoN);
   Array3<double> lK(dof*dof,eNoN,eNoN);
 
+  Array<double> Svis(nsd,nsd);
+  Array3<double> Kvis_u(nsd*nsd,eNoN,eNoN), Kvis_v(nsd*nsd,eNoN,eNoN);
```
michelebucelli, line 233: It seems to me that these arrays are not used by this function itself, but rather they are only used as "scratch" writing space for `struct_3d` and `struct_2d`, and they are declared here to avoid repeated allocations.

One downside is that this forces us to declare variables in a scope where those variables are not meaningful, which might make it harder to interpret the code.

An alternative solution to that would be to declare these variables inside the functions that need them (`struct_3d` and `struct_2d`) but declare them as `static`. This way, they will not be reallocated between subsequent calls, but they will remain local to the function's scope. You might have to write a bit of logic to ensure that they always have the correct size between calls, but I think it could be a good solution.

michelebucelli (commented):
Thank you @dseyler! I've left some suggestions.

As a general comment, I appreciate that `Eigen` brings some nice performance improvements, and that some of these changes also add some efficiency by avoiding unnecessary repeated allocations. However, there's a risk that both these kinds of changes increase the amount of "noise" in the code, by which I mean that the functions not only do the mathematical computation that they are supposed to do, but also spend quite a few lines in data format conversion and general bookkeeping, which may obfuscate their meaning a bit.

Some of the comments below go into the direction of mitigating this kind of issue (they are of course very much open to debate). I think if we can figure out some nice solution to this, then it can be used in the other performance improvements that you've worked on.

Code/Source/solver/mat_models.cpp
dseyler: Agreed. A max size of 27 is also set in fluid.cpp in two separate places. Would be nice to define this in a single place in the code to avoid bugs.

Code/Source/solver/mat_models.cpp
dseyler: That's correct. I think we should move toward option 1 eventually unless we plan to rewrite the Array class eventually. I sketched out a version of this branch that also changes all of 3d_struct() to use Eigen, and it simplifies many of the functions in mat_models.{cpp,h} because they already convert inputs to Eigen, and it shaves another 10% off the runtime. It would be a much larger PR (~500 lines changed), but I could fold that in here if you think that would be the better option.

Code/Source/solver/mat_models.cpp
dseyler: Sounds good. Would you rather add recompute_visc as a new argument for the potential and Newtonian viscosity models? I placed it within each case rather than throwing an early return at the top of the function, in case a different viscosity model is added in the future that does differ between Gauss points within linear elements. Alternatively, could just scrap the recompute_visc change if the 5% runtime savings isn't worth the risks/complexity of the assembly routine being treated differently for certain element types

Code/Source/solver/mat_models.cpp
michelebucelli: I think this PR would be a good place to do that. What are other people's thoughts? @ktbolt @zasexton

Code/Source/solver/mat_models.cpp
michelebucelli: The place where I think this would be cleanest (although maybe still not very clean 😅 ), right now, is to pass `recompute_visc` to `struct_2d` and `struct_3d` (as you do already), but not to `mat_models::compute_visc_stress_and_tangent`. So the `if (recompute_visc)` statement should wrap the call to `compute_visc_stress_and_tangent`, in my opinion.

It is true that this might break for future models (which is why I say it still isn't entirely clean). But then, should new viscosity models be implemented, it would be a good idea to give them an object-oriented refactoring, and perhaps this sort of caching could be encapsulated into the hypothetical viscous model class.

Code/Source/solver/mat_models.cpp
ktbolt: @michelebucelli I say let's do it; good to move Eigen as far up as possible.

dseyler:
Addressing comments here and the many changes that have been made to the branch:

- struct3d/2d and the ustruct per-Gauss-point assembly routines have been refactored using Eigen. This considerably condensed much of the arithmetic and made it much more readable. To facilitate this, the following changes were made:
    - `eigen_view` helper functions were added to construct read-only eigen views, reducing the amount of Eigen jargon in the assembly routine.
    - Aliases for fixed-size and bounded dynamic Eigen matrices were added in `mat_fun`
    - compute_pk2cc is now templated on `nsd` to accept fixed-size Eigen. An array overload still exists, used only in `post.cpp`. If preferred, this overload could be removed and the arrays can be converted to Eigen in `post.cpp` with a few loc before being passed into the templated functions.
    - `const int MaxNoN = 27` is added in `consts.h` and used to bound dynamic Eigen matrices. @zasexton confirmed this is the appropriate location. This value was previously defined in several different files and is now consolidated in one place. If larger element sizes are supported in the future, this will need to be updated, but it is now much easier to catch than before. In the future, this could be linked to `nn_elem_props.h` so that it automatically updates if larger elements are added.
    - `recompute_visc` is checked in the assembly routine and not passed to the viscosity function.
    - Scratch arrays for Kvis_u and Kvis_v were previously declared too far upstream of their use. Rather than move this into the `struct_3d/2d` where complicated resizing logic would be necessary to make use of `recompute_visc`, they are now owned by a new ViscousReponse class which hides the resize from the assembly routine. The viscosity call sites now read `visc.update(..)` and `Svis/Kvis_u/Kvis_v` are accessed as `visc.S()/visc.du/visc.dv`. Maybe this is a bit more of a refactor than necessary, but I think is much more readable.
    - Faster tensor operations were added in `mat_fun.h`. Maybe this should be a separate PR, but they were already in my branch that I merged with the viscosity branch to address @ktbolt and @michelebucelli's request to move Eigen as far up as possible.


A few more notes:

- Chasing Eigen conversions further upstream gets complicated, as it would require editing `nn::gnn` which has a wide impact on many modules. I also think it would have diminishing returns, as the functions edited in this branch are called per-Gauss point while the rest of the assembly routine is per-element or less frequent.
- Many of these same changes could be made to `b_struct*` and would make those functions easier to read, but not necessarily much faster.
- While implementing Eigen in `ustruct`, I found many bugs in `ustruct_2d` where arrays are clearly indexed out of bounds, either due to incorrect indexing conversion from FORTRAN or from copying 3d code into the 2d routines. There are currently no 2d ustruct test cases, so these bugs would not have been caught. I tried to fix all of the obvious bugs during the Eigen conversion but can not promise that more don't exist.

zasexton:
General question, when building the solver and running the Eigen Matrices and Tensors are you compiling with the `-march` flag in cmake? I believe this may be necessary for the Eigen vectorization optimizations.

zasexton:
It would be nice if the `maxNoN` was of type `size_t` since we would otherwise expect this to always be a positive integer value; however I understand that Eigen templating is built around `int` type for dimension sizing and because dynamic runtime allocation is represented by `-1`.

dseyler:
> General question, when building the solver and running the Eigen Matrices and Tensors are you compiling with the `-march` flag in cmake? I believe this may be necessary for the Eigen vectorization optimizations.

I was not. But it looks like Apple Silicon vectorizes with Eigen by default with ARM NEON. Is there a reason why this isn't the default on Sherlock though?

dseyler:
> It would be nice if the `maxNoN` was of type `size_t` since we would otherwise expect this to always be a positive integer value; however I understand that Eigen templating is built around `int` type for dimension sizing and because dynamic runtime allocation is represented by `-1`.

Sounds good. A lot of the templates were previously on `size_t` and I switched them to `int` per Claude's recommendation (for the reason you mentioned). I can switch them back though. I think it should be safe to use `size_t` as long as it's always positive. Will this be fine for the places where the templated parameter has arithmetic in it, i.e. `Matrix<3*(nsd-1)>`?

zasexton:
> > General question, when building the solver and running the Eigen Matrices and Tensors are you compiling with the `-march` flag in cmake? I believe this may be necessary for the Eigen vectorization optimizations.
> 
> I was not. But it looks like Apple Silicon vectorizes with Eigen by default with ARM NEON. Is there a reason why this isn't the default on Sherlock though?

It seems like someone made a note of this at some point https://github.com/SimVascular/svMultiPhysics/blob/c9838bdd0fbbbf613e801ecdb78a7c01fd859cb4/CMakeLists.txt#L48 

In general, I'd suspect that the `-march` flag is not used by default because it specifies hardware specific accelerators on the machine it is compiled on. This might cause problems if you compile against a specific CPU architecture but then you run a simulation across nodes that might not share the exact same type of architecture. I think we currently have two different CPU architectures on the sherlock partition (`MLN` and `SKX`).

zasexton:
> > It would be nice if the `maxNoN` was of type `size_t` since we would otherwise expect this to always be a positive integer value; however I understand that Eigen templating is built around `int` type for dimension sizing and because dynamic runtime allocation is represented by `-1`.
> 
> Sounds good. A lot of the templates were previously on `size_t` and I switched them to `int` per Claude's recommendation (for the reason you mentioned). I can switch them back though. I think it should be safe to use `size_t` as long as it's always positive. Will this be fine for the places where the templated parameter has arithmetic in it, i.e. `Matrix<3*(nsd-1)>`?

So long as any arithmetic is properly guarded for underflow then yes I believe that `size_t` should work. But before making any changes we can see what @ktbolt and @michelebucelli think about the `nsd` type preferences for `maxNoN`?

michelebucelli:
> So long as any arithmetic is properly guarded for underflow then yes I believe that size_t should work. But before making any changes we can see what @ktbolt and @michelebucelli think about the nsd type preferences for maxNoN?

@zasexton @dseyler I generally lean in favor of using `unsigned` types (like `size_t`) for quantities that are supposed to be non-negative. I see this first and foremost as a declarative practice, as in: the person writing the code declares to the person reading it that a number is assumed to be non-negative, with the variable type (as opposed to a documentation comment) documenting this intent.

I think this applies to `maxNoN` and `nsd` too, so I would prefer `size_t` over `int` for both of them.

Underflow is a possible concern, but given the semantics of these variables it should be fine without guards most of the time, and can be guarded with `static_assert`s in those cases where it isn't (e.g. before declaring something of size `nsd - 2`, one could place a `static_assert(nsd >= 2, ...)` just to be safe).

The guard is used to occasionally enforce a stricter constraint than just non-negativity (e.g. it must be greater than 2 in previous example), and this is better than using `int`s, which amounts to no compile-time constraint at all (or would require `static_assert` constraints in far more places).

Code/Source/solver/consts.h
```diff
@@ -18,6 +18,8 @@ namespace consts {
 
 const int maxNSD = 3;
 
+const int maxNoN = 27; // Max node count in nn_elem_props.h
```
michelebucelli: 1. I imagine you have discussed this internally already, but: why not place this in `nn_elem_props.h`, since that's where the information that this number is derived from is also stored?
2. Also, how does this relate to the FE basis refactoring recently implemented by @zasexton? Do you envision there will be a place in the new infrastructure to save/query this information?
3. It might be a good idea to declare this (and maybe other variables in this file) as `constexpr` instead of `const`.

   It shouldn't make a difference most of the time, since I expect the compiler will be able to figure it out by itself, but it makes it more explicit to the reader that these are supposed to be used in compile-time expressions (e.g. template as arguments).

Code/Source/solver/mat_models.cpp
```diff
@@ -816,6 +810,16 @@ void compute_pk2cc(const ComMod &com_mod, const CepMod &cep_mod,
   cc_to_voigt_eigen<nsd>(CC, Dm);
 }
 
+// The element routines know their dimension at compile time and call the
+// template directly, so instantiate the dimensions the solver supports.
```
michelebucelli: Minor: I suggest dropping the comment, because it reads a bit abstract and AI-ish without carrying much meaning. If we want to keep it, I would suggest rephrasing to something more concrete, e.g:
```cpp
// Explicitly instantiate compute_pk2cc for 2D and 3D.
```

Code/Source/solver/mat_models.cpp
```diff
@@ -852,20 +851,13 @@ void compute_pk2cc(const ComMod& com_mod, const CepMod& cep_mod, const dmnType&
...
-        auto F_3D = mat_fun::convert_to_eigen_matrix<Eigen::Matrix3d>(F);
-
-        // Copy fiber directions to Eigen matrix
-        Eigen::Matrix<double, 3, Eigen::Dynamic> fl_3D(3, nfd);
-        for (int i = 0; i < nfd; i++) {
-            fl_3D(0, i) = fl(0, i);
-            fl_3D(1, i) = fl(1, i);
-            fl_3D(2, i) = fl(2, i);
-        }
+        auto F_3D = mat_fun::convert_to_eigen_matrix<Matrix<3>>(F);
```
michelebucelli: I suggest making this `const` (the compiler probably figures it out on its own, but it might be better to be explicit).

Code/Source/solver/mat_fun.h
```diff
@@ -22,12 +23,68 @@
...
+    inline Eigen::Map<const Eigen::MatrixXd>
+    eigen_view(const Array<double>& A) {
+        return {A.data(), A.nrows(), A.ncols()};
+    }
+
+    /// @brief Writable Eigen view of a whole Array, sharing its storage.
+    inline Eigen::Map<Eigen::MatrixXd>
+    eigen_view_mut(Array<double>& A) {
+        return {A.data(), A.nrows(), A.ncols()};
+    }
```
michelebucelli: Minor: I suggest the more explicit name `eigen_view_mutable`.

Code/Source/solver/mat_fun.h
```diff
@@ -22,12 +23,68 @@
 //
 namespace mat_fun {
     // Define templated type aliases for Eigen matrices and tensors for convenience
-    template<size_t nsd>
+    template<int nsd>
     using Matrix = Eigen::Matrix<double, nsd, nsd>;
```
michelebucelli: I'm not sure myself about this suggestion, but I think there's an argument to be made for this to be called `Tensor` instead of matrix (and the current `Tensor` alias below be renamed to `Tensor4`).

This type is used to represent square matrices with as many elements as there are spatial dimensions (so the name should maybe be more restrictive than just `Matrix` to reflect this). In the constitutive modelling code this is used to represent objects that, mathematically, are tensors rather than general square matrices. I think changing the name would make this more apparent.

Code/Source/solver/mat_fun.h
```diff
@@ -22,12 +23,68 @@
...
 
-    template<size_t nsd>
+    template<int nsd>
     using Tensor = Eigen::TensorFixedSize<double, Eigen::Sizes<nsd, nsd, nsd, nsd>>;
```
michelebucelli: I suggest to give Doxygen documentation to these aliases (especially with reference to their intended use and whether they are fixed- or dynamic-sized).

Code/Source/solver/mat_fun.h
```diff
@@ -208,14 +236,29 @@ namespace mat_fun {
     Tensor4<double> ten_ddot_3424(const Tensor4<double>& A, const Tensor4<double>& B, const int nd);
 
     /**
-     * @brief Contracts two 4th order tensors A and B over two dimensions, 
-     * 
+     * @brief Contracts two 4th order tensors A and B over two dimensions.
```
michelebucelli: I think it would be helpful to report the expression on the contraction (e.g. in Einstein notation) in the documentation, to make sure this is clear.

Code/Source/solver/mat_models.h
michelebucelli: Are there any call sites left where this is needed? If so, would it be possible to have them use `Matrix` too, and drop this overload altogether? Or is this needed to dynamically dispatch at runtime to the correct dimension (since `com_mod.nsd` is a run-time variable)?

The less overloads there are, the easier it is for the reader to figure out which function is being called (especially for functions with this many arguments).

Code/Source/solver/mat_models.h
```diff
@@ -67,20 +85,41 @@ void g_vol_pen(const ComMod& com_mod, const dmnType& lDmn, const double p,
...
+    /// entry of the block coupling nodes a and b.
+    double du(const int ij, const int a, const int b) const { return Kvis_u_(ij, a, b); }
+
+    /// @brief Tangent w.r.t. velocity.
+    double dv(const int ij, const int a, const int b) const { return Kvis_v_(ij, a, b); }
+
+  private:
+    Matrix<nsd> Svis_;
+    Array3<double> Kvis_u_, Kvis_v_;
+};
```
michelebucelli: If I understand this right, this class is meant to be a helper that assembly routines use to walk over the mesh elements and evaluate a certain quantity (viscous stress) at their integration points. Is this correct?

If it is, I think this has some analogy with the `ActiveStress::Evaluator` class I have implemented in #650, and I expect a similar pattern to pop up in a number of other places.

I think it might be a good idea to abstract this away and establish a common infrastructure for "element walkers"/"integration point evaluators", to avoid repeated work, ensure consistency, and facilitating adding new instances of this pattern.

I propose we discuss this after this PR and #650 are merged.

Code/Source/solver/sv_struct.cpp
```diff
@@ -441,20 +427,27 @@ void struct_2d(ComMod &com_mod, CepMod &cep_mod, const int eNoN, const int nFn,
...
-  // Viscous 2nd Piola-Kirchhoff stress and tangent contributions
-  mat_models::compute_visc_stress_and_tangent(dmn, eNoN, Nx, vx, F, Svis, Kvis_u, Kvis_v,
-                                              recompute_visc);
+  // Viscous 2nd Piola-Kirchhoff stress and tangent contributions.
+  // Reuse from the previous Gauss point when shape function gradients
+  // are constant within an element (e.g. linear triangles, tetrahedra).
+  static mat_models::ViscousResponse<2> visc;
+  if (recompute_visc) {
+    visc.update(dmn, eNoN, Nx, vx, F);
+  }
```
michelebucelli: I think this works under the assumption that repeated calls to this method happen in a particular order, which is currently true, but that this method itself has no control over. Should this change in the future, it might give rise to bugs that are very hard to find (and even to notice, probably).

A possible better design could be one where the methods `struct_2d` and `struct_3d` are not responsible for evaluating/updating the viscous stress, but receive a `ViscousResponse` object in input, and this object is managed by the caller. This way, the same function that loops over the elements is the one that decides whether to recompute viscosity or not.

Another possible design would be one where this is delegated to `ViscousResponse` itself, i.e. its `update` method decides whether it really has to update or it can silently reuse the already computed quantities. This however would require passing more information into the `ViscousResponse` class, and I'm not sure we want to do that. It's something we can discuss as part of the proposed evaluator abstraction that I'm mentioning in some other comment.

Similar considerations apply to other places where this class is used (also in `ustruct`).

(Sorry for going back and forth on this issue, I am struggling a bit to find a clean and safe solution that still allows reuse of the viscous tensors).

michelebucelli (commented):
Thanks @dseyler! I did another review pass.

I think using `Eigen` the code looks much cleaner and more readable!

I still have some doubts about the current implementation of viscous stress caching across Gauss points, sorry if I keep returning to that (and for occasionally contradicting my own previous suggestions).

> Chasing Eigen conversions further upstream gets complicated, as it would require editing nn::gnn which has a wide impact on many modules. I also think it would have diminishing returns, as the functions edited in this branch are called per-Gauss point while the rest of the assembly routine is per-element or less frequent.

I think long term it might be a good idea to use `Eigen` for every small, dense linear algebra task (parallel and/or sparse linear algebra is a different question entirely). I can see your point about diminishing returns in terms of efficiency, but I think the returns would not be diminishing in terms of code hygiene and maintenance burden. I agree that such a large refactor would be out of scope for this PR, so I'm fine with what you currently did (and also I appreciate how this PR sets a good example of the improvements that can be obtained).

I think we might make a bit of a collective effort to slowly migrate to `Eigen` the parts of code that we touch when we get the opportunity (and maybe remember to gently suggest to do this in otherwise unrelated PR reviews 😅).

Code/Source/solver/mat_models.h
dseyler: I mostly implemented this to hide the whole `Kvis_u/v` resizing logic from `struct3d/2d`, but I agree it would be nice to have some common infrastructure for this kind of thing. Ideally, I feel like the different contributions to the tangent and residual should be able to be abstracted away so that the whole assembly routine can be separated into viscous, constitutive, prestress, etc. contributions instead of a long routine of disordered arithmetic.

Right now construct_d_solid loops over each element and each Gauss point and then runs struct3d to assemble lK and lR per gauss point. I feel like a lot of this code would be simpler if construct_d_solid (or a future general assembly function) just looped over the elements, and then physics-specific assembly kernels looped over Gauss points. This would make it much easier to separate quantities that need to be computed once per element (viscosity, shape function derivatives for linear elements) and those that need to be computed per Gauss point and would remove the need for the recompute_visc flag

Code/Source/solver/sv_struct.cpp
dseyler: Didn't read this before my response to your comment above on the ViscousResponse class.

But if I understand right, we both agree a lot of this complexity comes from the fact that the boundary between constructd_solid and struct_2/3d is at the level of the Gauss points rather than the element.

If the boundary were in the element loop, the `update` method could decide what to do depending on the element type, but currently it doesn't have that info since the element loop lies one layer up in constructd_solid. If visc.update were called once per element, the method would not need to have a recompute_visc flag passed in

dseyler:
> I still have some doubts about the current implementation of viscous stress caching across Gauss points, sorry if I keep returning to that (and for occasionally contradicting my own previous suggestions).

Thanks for reviewing all of this for a second time! I agree that the viscous stress caching is a bit cumbersome and am not married to it. The ultimate goal of that feature was to save people time by making our simulations faster, but if it costs us more time to review and rewrite than it saves, then it might defeat the purpose :sweat_smile:

Code/Source/solver/sv_struct.cpp
michelebucelli: This is a good way of framing the issue. However, even if `visc.update` were called once per element, it would need to decide whether to compute a different viscous tensor for every integration point or not, which would depend on information that the `ViscousResponse` currently doesn't have.

Because of this, I still think the two problems (whether `ViscousResponse` works per element or per Gauss node, and how to avoid repeated computations) are not exactly the same.

I think for the sake of this PR the first solution I propose (evaluating viscosity in the caller of `struct_2d/3d`) would be a reasonable solution, but we may want to revisit this if we abstract the integration point evaluator pattern away (which I think we should, but after this).

dseyler: Sounds good! I'll go with your first solution

Code/Source/solver/mat_models.h
dseyler: It's used in `post.cpp` and the unit tests. The array overload could be removed and just dispatched to the templated version in `post.cpp` (with a few lines converting arrays to Eigen before calling compute_pk2cc<nsd>)

Code/Source/solver/sv_struct.cpp
dseyler: Since ViscousResponse is templated on runtime `nsd`, if it's moved into the caller, then either 1 (easiest): both need to be declared in `construct_d_solid`, `construct_u_solid` and `construct_d_fsi` or 2: those functions could also be templated on `nsd` as well.

Easiest: 
   ```
  /// Declare both. Use only one depending on whether struct3d or struct 2d is called
  mat_models::ViscousResponse<2> visc2;
  mat_models::ViscousResponse<3> visc3;
```

Ideally, this would be cleaner. Do you have thoughts?

Comment (dseyler) [reply Code/Source/solver/sv_struct.cpp]: Looked into this a bit more:

Computing the ViscousResponse before `struct_3d/2d` and passing it in as an argument would require hoisting the computation of its arguments `F` and `vx` out of `struct3d/2d` as well.

I think this is a good plan to do eventually, since `F` and `vx` are also uniform over linear elements (they only depend on shape function derivatives), so it's a little odd that viscosity is computed once per element but `F` and `vx` are computed for each Gauss point even though `F` and `vx` are arguments of `visc.update`.

If this were implemented, then the `recompute_visc` check would be entirely contained in `construct_dsolid()` and read:
```
for e < lM.nEL {
for g < lM.nG {
if nsd == 2:
  if (g == 0 || !lM.lShpF) {
      nn::gnn(Nx)
      Nxm = eigen_view<2>(Nx)
      vx  = ...
      F = ...
      visc.update()
  struct2d(...,visc)

if nsd == 3
.
.
.
```

I think at that point it would make sense to convert `gnn` to Eigen and template it on `nsd` so the `vx` and `F` computation is a little more legible. `gnn` was also one of the biggest consumers of runtime from my earlier profiling work, but I didn't look into it because it touches so many other modules. Also, just looking at the function, it is a LOT of unrolled matrix multiplications.

How does this sound for a plan:

1.  Either pass recompute_visc as a parameter into visc.update() (Your suggestion 2) or scrap the recompute_visc idea for now and come back to it in a later PR. I don't feel strongly either way.

2. At a later date, convert `gnn` to Eigen and move `visc.update` out of `struct3d.` We could create similar methods for `compute_pk2cc`'s passive and active contributions to S, i.e. passive.update() and active.update() as well, using the general Evaluator class you suggested.

Issue #638, opened 2026-09-04
svMultiPhysics/tests/cases/fluid/dye_AD does not output Concentration


ktbolt:
### Description

The `svMultiPhysics/tests/cases/fluid/dye_AD` test simulates a fluid flow with dye transport. The dye concentration should be output to a VTU results file as a`Concentration` data array.

However, running the simulation does not output `Concentration` to a VTU results file.

### Reproduction

Run the `svMultiPhysics/tests/cases/fluid/dye_AD` test. 

### Expected behavior

The dye concentration should be output to a VTU file as a `Concentration` data array.

### Additional context

This test does not check if concentration, the primary result of the simulation, is correct. 

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct and Contributing Guidelines

ktbolt:
The `svMultiPhysics/tests/cases/fluid/dye_AD/solver.xml` file `Output` parameters for the `scalarTransport` equation are
```
   <Output type="Volume_integral" >
     <Temperature> true </Temperature>
   </Output>

   <Output type="Alias" >
     <Temperature> Concentration </Temperature>
   </Output>
```

The problem is that `<Output type="Alias" >` does nothing unless `Temperature` is also output
```
   <Output type="Spatial" >
     <Temperature> true </Temperature>
   </Output>
```
Adding this to `svMultiPhysics/tests/cases/fluid/dye_AD/solver.xml` causes `Concentration` to be written to the VTU results file.

ktbolt:
Merged.

PR #642, opened 2026-09-09
Fix writing out and testing Concentration.

ktbolt:
This fixes writing Concentration data and updates the CI test to check those results. https://github.com/SimVascular/svMultiPhysics/issues/638.

### DIFF
--- a/tests/cases/fluid/dye_AD/result_001.vtu
+++ b/tests/cases/fluid/dye_AD/result_001.vtu
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:83912ef0d0dfb186de6ff453a8b60eb4e34c3d0c88ac7c535a5a67addda20a25
-size 81322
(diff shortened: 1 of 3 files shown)


kko27 (approved):
Approved! I did have two comments. 

1. Is there documentation for setting up simulations that involve scalar transport? I did not realize that there was a nuance of creating an Alias output type (to rename temperature as concentration).
2. Where does the reference solution for this test case come from?

zasexton (approved):
looks good to merge

zasexton:
I think the CI failures result from the field `Concentration` not being defined in the reference solution file, so test_dye_AD fails with ValueError: Field Concentration not in reference result for 1, 3, and 4 processes on both Ubuntu and macOS. These may need to be regenerated.

ktbolt:
@zasexton Thanks for reminding me about this! 

I did add a new reference solution file for test_dye_AD but it seems that it did not get merged correctly.

Issue #643, opened 2026-09-09
Error in lumen_inlet.flow data


ktbolt:
### Description

A `lumen_inlet.flow` file is used as data for an unsteady flow boundary condition for several CI fluid tests:
casson, newtonian, carreau_yasuda, and dye_AD

However, this file appears to have a wrong flow value at line 10
```
0.250000,-6.283185
```

<img width="625" height="462" alt="Image" src="https://github.com/user-attachments/assets/6e9bc028-2fc5-43bd-b194-b8c65012cbbd" />

### Reproduction

None

### Expected behavior

Flow values should not have a jump.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct and Contributing Guidelines

ktbolt:
This error is was never discovered because the CI tests only run for a single time step. However, if a user ran a test for more time steps they would see confusing results (blue) 

<img width="965" height="564" alt="Image" src="https://github.com/user-attachments/assets/4470e09f-4b6f-42f4-9dc9-61432f77d628" />

This is how I discovered the error.

The `/tests/cases/fluid/pipe_RCR_3d/lumen_inlet.flow` file is correct; line 10 is
```
0.250000    -62.83185
```
so  the incorrect flow value has a decimal shifted one place; wonder how that happened.

PR #644, opened 2026-09-09
Fix error in lumen_inlet.flow files.

ktbolt:
This fixes an error at line 10
```
0.250000,-6.283185
```

in several `tests/cases/fluid` lumen_inlet.flow files used in CI tests. https://github.com/SimVascular/svMultiPhysics/issues/643

CI test results files don't need to be updated because the test simulations are run for a single time step with dt=0.001.


javijv4 (approved):
Looks good to me!

zasexton (approved):
Reviewed this and also think that it looks good to merge.

PR #645, opened 2026-09-14
FE Quadrature Phase 02: bounded Gaussian line generators

zasexton:
## Current situation

[Phase 01 (#593)](https://github.com/SimVascular/svMultiPhysics/pull/593) introduced the concrete `QuadratureRule` value type, but the module does not yet provide generators for Gaussian line rules. This PR implements Phase 02 of
[#580](https://github.com/SimVascular/svMultiPhysics/issues/580) by adding Gauss-Legendre and Gauss-Lobatto-Legendre generators on `[-1, 1]`.

Both functions accept signed `int requested_exactness` and return a complete `QuadratureRule` by value with at least that exactness. Each family owns its conversion to the smallest sufficient point count; future selectors do not need to duplicate these formulas or infer exactness limits from point limits. These line rules provide factors for later element generators. Existing solver quadrature tables and consumers are unchanged.

Newton refinement computes the Legendre polynomial roots or stationary points needed to support each requested size. The implementation uses a three-term recurrence and analytic derivatives, computes one half of the samples, and
mirrors them into increasing order. Refinement has a defensive 100-iteration bound. Correction and measure tolerances are `64 * epsilon` and `32 * 128 * epsilon`, respectively, where `epsilon` is `std::numeric_limits<double>::epsilon()`; both are qualified over every supported point count. Invalid exactness requests raise `InvalidArgumentException` before
point-count conversion or sample allocation; unsuccessful refinement or numerical validation raises `ConvergenceException`. Failed weights are never rescaled to force the reference measure.

## Release Notes

- Add `make_gauss_legendre_rule(int requested_exactness)` for requests `0..255`, using integer division `n = requested_exactness / 2 + 1` to select 1 through 128 interior points with actual exactness `2n - 1`.
- Add `make_gauss_lobatto_rule(int requested_exactness)` for requests `0..253`, using `n = requested_exactness / 2 + 2` to select 2 through 128 points, including both endpoints exactly, with actual exactness `2n - 3`.
- Accept degree zero and preserve actual returned exactness: requests 4 and 5 both produce degree-5 rules, using three Gauss points or four Lobatto points.
- Expose `constexpr noexcept` capacity queries `max_gauss_legendre_exactness()` returning 255 and `max_gauss_lobatto_exactness()` returning 253. The internal 128-point ceiling is a project support bound that limits construction cost and downstream product-rule growth, not a mathematical or convergence limit.
- Return ordered, symmetric samples with positive weights, exact odd-rule centers, and the correct reference measure within the documented tolerance.

## Documentation

The two public headers document supported request ranges, integer-division conversion formulas, at-least-requested semantics, degree-zero behavior, actual metadata, ordering, endpoints, capacity rationale, and exception semantics in
the existing `FE_Quadrature` Doxygen group. Implementation and test comments explain the numerical tolerances and qualification strategy.

The PR contains only the two generator headers, their implementations, and `test_QuadratureGenerators.cpp`. Current upstream's `svmp_fe_quadrature` target and unit-test discovery include these files without CMake edits.

## Testing

The eight focused tests cover:

- Canonical low-order fixtures, degree-zero minimum counts, and returned exactness greater than an even request.
- Every supported request: all 256 Gauss-Legendre degrees and all 254 Lobatto degrees, selecting 128 and 127 distinct rules respectively. Both even and odd requests check point counts, actual metadata, ordering, containment, symmetry, positive weights, exact assigned coordinates, and reference measure.
- All advertised analytic monomial moments for both request parities: 65,536 checks using compensated long-double accumulation, plus first-unadvertised-degree checks for small rules.
- Supported upper boundaries, first unsupported degrees, negative requests, integer extremes, and exception diagnostics.
- Representative Lobatto node comparisons with the independent Basis implementation, with the cross-module dependency confined to test code.
- `svmp_fe_quadrature`, `run_all_unit_tests`, and `svmultiphysics` built successfully with the revised API.
- All eight generator tests passed across four suites.
- All 17 Quadrature tests passed across seven suites.
- All 260 unit tests passed across 35 suites.


tests/unitTests/FE/Quadrature/test_QuadratureGenerators.cpp
```diff
@@ -0,0 +1,477 @@
...
+    } catch (const std::exception& exception) {
+        FAIL() << "Received a different exception type: " << exception.what();
+    } catch (...) {
+        FAIL() << "Received an unknown exception type";
+    }
+}
+
+} // namespace
+
+TEST(QuadratureGeneratorTestSupport, ExercisesSharedLineRuleChecks)
```
michelebucelli, line 286: I suggest, for this and the following ones, to add some documentation explaining what is being checked. I suggest using the `@test` Doxygen command, which will place the test documentation under a test list page.

Code/Source/solver/FE/Quadrature/GaussQuadrature.h
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * @brief Generate a Gauss-Legendre rule on @f$[-1,1]@f$ with at least the
+ *        requested exactness.
+ *
+ * @details For @f$d@f$ = @p requested_exactness, integer division gives the
+ * minimum point count @f$n=\lfloor d/2\rfloor+1@f$. The returned metadata reports
+ * the actual polynomial exactness @f$2n-1@f$, which exceeds even requests by one.
+ * Degree zero produces one point with exactness one. Points are the roots of
+ * @f$P_n@f$ in strictly increasing order inside @f$(-1,1)@f$; neither endpoint
```
michelebucelli, line 41: I think that the meaning of @f$P_n@f$ is not defined. I suggest reporting the name of the polynomial family (e.g. "Points are the roots of Legendre polynomials $P_n$") and linking back to a numerical analysis textbook and/or the Legendre polynomials' Wikipedia page for reference.

Code/Source/solver/FE/Quadrature/GaussQuadrature.cpp
```diff
@@ -0,0 +1,230 @@
...
+} // namespace
+
+QuadratureRule make_gauss_legendre_rule(int requested_exactness)
+{
+    if (requested_exactness < 0 || requested_exactness > max_gauss_legendre_exactness()) {
+        std::ostringstream message;
+        message << "Gauss-Legendre generator: requested_exactness must be in [0, "
+                << max_gauss_legendre_exactness() << ']';
+        svmp::raise<InvalidArgumentException>(message.str());
+    }
```
michelebucelli, line 186: I think using `std::to_string` makes for cleaner code than `ostringstream`:
```suggestion
    svmp::check<InvalidArgumentException>(
      requested_exactness >= 0,
      "requested_exactness cannot be negative.");
      
    svmp::check<InvalidArgumentException>(
      requested_exactness <= max_gauss_legendre_exactness(),
      "requested exactness cannot be greater than " + 
        std::to_string(max_gauss_legendre_exactness() + ".");
```

Code/Source/solver/FE/Quadrature/GaussQuadrature.cpp
```diff
@@ -0,0 +1,230 @@
...
+
+    const int num_points = requested_exactness / 2 + 1;
+    std::vector<QuadPoint> points(
+        static_cast<std::size_t>(num_points), QuadPoint::Zero());
```
michelebucelli, line 190: Why not define `num_points` as a `size_t` and skip the `static_cast`?

```suggestion
    const std::size_t num_points = requested_exactness / 2 + 1;
    std::vector<QuadPoint> points(num_points, QuadPoint::Zero());
```

Code/Source/solver/FE/Quadrature/GaussQuadrature.cpp
```diff
@@ -0,0 +1,230 @@
...
+// sweeps qualify this scale.
+constexpr double kNewtonCorrectionTolerance =
+    64.0 * std::numeric_limits<double>::epsilon();
+// Provide conservative O(n epsilon) accumulation headroom, qualified by those
+// sweeps.
+constexpr double kRuleValidationTolerance =
+    32.0 * static_cast<double>(kMaximumPoints) *
+    std::numeric_limits<double>::epsilon();
+
+std::pair<double, double> evaluate_legendre_with_derivative(
```
michelebucelli, line 43: I understand that all these functions are internal, and so their documentation will not be user-facing. However, I still think there should be some developer-facing documentation to explain what they are doing and how. Otherwise, the code on its own is somewhat obscure, unless one knows very well how this rule is constructed.

The same comment applies to other functions within this anonymous namespace.

Code/Source/solver/FE/Quadrature/GaussQuadrature.cpp
```diff
@@ -0,0 +1,230 @@
...
+    double derivative = 1.0;
+    for (int recurrence_degree = 2; recurrence_degree <= degree; ++recurrence_degree) {
+        const double degree_value = static_cast<double>(recurrence_degree);
+        const double recurrence_factor = static_cast<double>(2 * recurrence_degree - 1);
+        const double next_value =
+            (recurrence_factor * coordinate * value -
+             static_cast<double>(recurrence_degree - 1) * previous_value) / degree_value;
+        const double next_derivative =
+            (recurrence_factor * (value + coordinate * derivative) -
+             static_cast<double>(recurrence_degree - 1) * previous_derivative) / degree_value;
```
michelebucelli, line 59: Why do you need these `static_casts`? The case will be performed implicitly whenever you multiply or add an integer to a `double`. I think they can all be safely removed, improving readability.

Code/Source/solver/FE/Quadrature/GaussQuadrature.cpp
```diff
@@ -0,0 +1,230 @@
...
+    int root_index,
+    int iteration,
+    double diagnostic_value,
+    std::string_view detail)
+{
+    std::ostringstream message;
+    message << "Gauss-Legendre generator: " << detail
+            << ", num_points=" << num_points
+            << ", root_index=" << root_index
+            << ", diagnostic_value=" << diagnostic_value;
```
michelebucelli, line 81: Minor: same as in my other comment, I think using `std::to_string` makes for somewhat lighter code (and spares you from including `sstream`).

Code/Source/solver/FE/Quadrature/GaussQuadrature.cpp
```diff
@@ -0,0 +1,230 @@
...
+
+    const double residual = std::isfinite(diagnostic_value)
+                                ? std::abs(diagnostic_value)
+                                : 0.0;
```
michelebucelli, line 85: If the residual is not finite (e.g. it's `NaN` or `Inf`), wouldn't it be better to report it as such, rather than reporting it as zero? This might mask the issue underlying the exception to some extent (`NaN` and `Inf` would likely be symptoms of different issues, so distinguishing them might help diagnose the problem).

Code/Source/solver/FE/Quadrature/GaussQuadrature.cpp
```diff
@@ -0,0 +1,230 @@
...
+    svmp::raise<ConvergenceException>(message.str(), iteration, residual);
+}
+
+void require_generation(bool condition, int num_points, int root_index, int iteration,
+    double diagnostic_value, std::string_view detail)
+{
+    if (!condition) {
+        raise_generation_failure(num_points, root_index, iteration, diagnostic_value, detail);
+    }
+}
```
michelebucelli, line 95: I think the indirection introduced by this function masks its meaning a bit where it is used below. I think it would be cleaner to remove this and just call `if (!condition) { raise_generation_failure(...); }` wherever necessary.

Code/Source/solver/FE/Quadrature/GaussQuadrature.h
```diff
@@ -0,0 +1,59 @@
...
+ * @brief Return the largest supported Gauss-Legendre exactness request.
+ * @details Degree 255 corresponds to the internal 128-point project support
+ * bound, which limits generator work and downstream product-rule growth; it
+ * is not a mathematical or convergence limit.
+ * @return The inclusive requested-exactness limit, 255.
+ */
+[[nodiscard]] constexpr int max_gauss_legendre_exactness() noexcept
+{
+    return 255;
+}
```
michelebucelli, line 31: Is this bound strictly necessary?

I assume no one will ever request such a large formula, but the code can reasonably be expected to work even fopr high degree of exactness.

On the other hand, I don't think that this maximum size is currently used in any way (e.g. I can't see any point where the compiler would optimize knowing this maximum size), so as far as I can tell this adds a little amount of complexity/noise without any practical advantage. Unless I am missing something, I would recommend to remove this and the associated assertion.

Code/Source/solver/FE/Quadrature/GaussLobattoQuadrature.h
michelebucelli, line 1: My comments on this file and on the associated source file essentially mirror those I left for `GaussQuadrature.{h,cpp}`.

michelebucelli (commented):
Thanks @zasexton! I left some comments below.

Code/Source/solver/FE/Quadrature/GaussQuadrature.h
zasexton: I think the 128-point limit helps to define a supported and exhaustively tested range. Removing it
might require more numerical re-qualification and overflow-safe arithmetic for larger requests (of course the use of such large requests is doubtful). I have these limits as practical policy for arbitrarily large yet finite rule-making. At least this was my thinking.

tests/unitTests/FE/Quadrature/test_QuadratureGenerators.cpp
zasexton: Addressed in [ce871ec3](https://github.com/SimVascular/svMultiPhysics/commit/ce871ec318dfa014767f83837dde0c858195953a). Added `@test` descriptions for all eight generator tests and included the test file in Doxygen.

Code/Source/solver/FE/Quadrature/GaussQuadrature.h
zasexton: Addressed in [a6ca9bf5](https://github.com/SimVascular/svMultiPhysics/commit/a6ca9bf56587cf37cea6f36c8efd5757a26c2056). Defined `P_n` as the Legendre polynomial of degree `n` and added a reference to the NIST DLMF with website link in documentation to proper standards.

Code/Source/solver/FE/Quadrature/GaussQuadrature.cpp
zasexton: Addressed in [b751f2d3](https://github.com/SimVascular/svMultiPhysics/commit/b751f2d3420ca0aa252232dbcb67b2f7f8e569b0). Removed the redundant integer-to-double casts

Code/Source/solver/FE/Quadrature/GaussQuadrature.cpp
zasexton: Addressed in [a177daf3](https://github.com/SimVascular/svMultiPhysics/commit/a177daf3b072e611746b341edf39197757225ad8). The convergence exception now preserves `NaN` and infinity in its stored residual instead of replacing them with zero

Code/Source/solver/FE/Quadrature/GaussQuadrature.cpp
zasexton: Addressed in [eef3a4a5](https://github.com/SimVascular/svMultiPhysics/commit/eef3a4a500a2b1864327ea524b434c6c19d4a2f1). Replaced `require_generation` with explicit `if` checks calling `raise_generation_failure`

Code/Source/solver/FE/Quadrature/GaussLobattoQuadrature.h
zasexton: Addressed in [eaa55855](https://github.com/SimVascular/svMultiPhysics/commit/eaa558559f2828e0d214915692c2ce4c1f4230f6) for mirrored concerns in Lobatto rule

Code/Source/solver/FE/Quadrature/GaussQuadrature.cpp
zasexton: I think `sstream` was just a preference of mine since some small values like `std::to_string(1e-14)` produce `0.000000` in C++20 for double values. But I will take another look

PR #647, opened 2026-09-15
Fix FSILS RCS diagonal cancellation and general block scaling

### LINKED ISSUE #646: FSILS RCS preconditioner alters small matrix diagonals and degrades convergence
### Description

The FSILS row and column scaling (RCS) preconditioner changes small, unconstrained matrix diagonal entries while applying Dirichlet boundary conditions. This introduces rounding error into the operator before row and column scaling begins.

This is a shared linear-solver issue. The same failure has been reproduced with the existing `heatS` equation and the Darcy equation reviewed in [PR #620](https://github.com/SimVascular/svMultiPhysics/pull/620). For an exactly representable linear solution on a small triangular mesh, both RCS runs have a maximum nodal error of approximately **0.383** after five outer iterations. Changing only the preconditioner to `fsils` gives zero error at the recorded output precision.

The problematic expression appears in every scalar and block-size branch of `precond_rcs`. Other physics using this FSILS preconditioner may therefore be affected when their assembled matrix diagonals are small.

#### Solver-level reproduction

Use a 2D unit square with a structured 5-by-3 node grid and 16 linear triangles:

- Prescribe `u = 1` on the left edge and `u = 0` on the right edge.
- Leave the top and bottom edges at their natural zero-flux boundary condition.
- Use a zero initial field and no source. The exact steady solution is `u(x, y) = 1 - x`, which this mesh can represent exactly.
- For `heatS`, set `Density = 0` and `Conductivity = 5e-16`. The zero density deliberately removes the transient term for this steady diffusion test.
- For the equivalent Darcy check, set fluid density to `0.5`, permeability to `1e-15`, viscosity to `1`, and compressibility to `0`.
- Use one time step of size `0.1`, spectral radius `0`, and at most five outer iterations with tolerance `1e-10`.
- Select the FSILS linear algebra backend with CG, relative tolerance `1e-12`, absolute tolerance `1e-30`, and at most 30 linear iterations. Compare `rcs` and `fsils` preconditioners.

Measured maximum nodal error, `max_i |u_i - (1 - x_i)|`:

| Physics | Preconditioner | Maximum nodal error |
| --- | --- | ---: |
| `heatS` | `rcs` | `3.8263694109545121e-01` |
| `heatS` | `fsils` | `0.0000000000000000e+00` |
| `darcy` | `rcs` | `3.8263694109545121e-01` |
| `darcy` | `fsils` | `0.0000000000000000e+00` |

The RCS runs reach the five-iteration outer limit without meeting the outer convergence tolerance, even though the inner linear solves report very small residual ratios. The resulting corrections produce slow outer convergence and inaccurate output at the iteration limit.

Tightening the linear absolute tolerance to `1e-30` does not eliminate the problem, as the table shows. In a separate Darcy run using 30 outer iterations and the default linear absolute tolerance, the maximum nodal error was still `3.9171435852137615e-02`.

### Expected behavior

Dirichlet handling should leave unconstrained matrix diagonals unchanged and set constrained diagonals to one before applying the intended row and column scaling. Preconditioning should preserve the equivalent linear system.

Uniformly reducing the diffusion coefficient should preserve the exact solution `u = 1 - x`. The test should converge to the requested accuracy without the large coefficient-dependent error shown above.

#### Shared code path and numerical mechanism

The relevant code is [`precond_rcs` in `Code/Source/linear_solver/precond.cpp`](https://github.com/SimVascular/svMultiPhysics/blob/00f5c9204bb5a9991e4c8d18eddd3b1492f3b59a/Code/Source/linear_solver/precond.cpp#L271-L348). After masking constrained rows and columns, it performs the following update for a scalar diagonal:

```cpp
Val(0,d) = Wr(0,Ac) * (Val(0,d) - 1.0) + 1.0;
```

The same expression is repeated for block sizes 2, 3, 4, and the general block-size case.

For an unconstrained degree of freedom, `Wr = 1`, so this evaluates `(a - 1.0) + 1.0` for a diagonal value `a`. Although algebraically equal to `a`, the subtraction rounds at the scale of one and can lose the small diagonal value. For example, ordinary double-precision evaluation gives:

| Original diagonal `a` | After `(a - 1.0) + 1.0` |
| ---: | ---: |
| `1e-11` | approximately `1.000000082740371e-11` |
| `1e-16` | approximately `1.1102230246251565e-16` |
| `1e-17` | `0` |

This operation modifies free diagonals as well as constrained ones. Subsequent scaling cannot recover the lost information.

[`fsils_solve` dispatches to this shared routine](https://github.com/SimVascular/svMultiPhysics/blob/00f5c9204bb5a9991e4c8d18eddd3b1492f3b59a/Code/Source/linear_solver/solve.cpp#L96-L104), independently of the originating physics assembly. Existing CEP and Stokes test configurations also select FSILS RCS. 

#### Version and environment

- Tested commit: [`28f48ece9f79c29df7ae02bb1417e5f7ace69f38`](https://github.com/SimVascular/svMultiPhysics/commit/28f48ece9f79c29df7ae02bb1417e5f7ace69f38).
- `precond.cpp` is identical to the PR's base commit, [`00f5c9204bb5a9991e4c8d18eddd3b1492f3b59a`](https://github.com/SimVascular/svMultiPhysics/commit/00f5c9204bb5a9991e4c8d18eddd3b1492f3b59a). 
- Ubuntu 22.04.5 LTS; GCC 13.1.0; Release build (`-O3 -DNDEBUG`) with array index checking enabled.
- Open MPI 4.1.2; one MPI rank per solver run.
- Solver linked against VTK 9.1; reproducer used Python VTK 9.3.1 and ASCII mesh input files.

#### Suggested resolution and regression coverage

Preserve unconstrained diagonals directly and set only constrained diagonals to one, with consistent behavior in every block-size branch.

The standard FSILS diagonal preconditioner (`<Preconditioner>fsils</Preconditioner>`) is a verified workaround for the two scalar examples in this report.

### Code of Conduct

zasexton:
## Current situation

Fixes #646.

The FSILS row and column scaling (RCS) preconditioner evaluates `(a - 1.0) + 1.0` for unconstrained matrix diagonals while applying Dirichlet conditions. In double precision, this perturbs small coefficients and can erase them before scaling begins. The resulting operator no longer represents the assembled system, which can degrade convergence and solution accuracy across physics using this shared preconditioner.

For block sizes greater than four, column scaling also skips the last row of each block, and the row-maximum calculation omits the first entry. The latter can produce infinite scaling weights when a constrained diagonal is the only nonzero entry in its row.

## Release Notes

- Preserve unconstrained RCS diagonals and set only constrained diagonals to one, using the normalized boundary mask for every block size.
- Correct the general block path to scale every block row and include every entry when computing row maxima.
- Add RCS matrix regression coverage for small coefficients, boundary constraints, and block sizes one through five.

## Documentation

The numerical cancellation guard is documented inline in `Code/Source/linear_solver/precond.cpp`. The regression test documents its sparse matrix stencil and scale-normalized error checks.

## Testing

- **327 unit tests passed**, including **75 RCS matrix cases**. These cover block sizes 1–5, coefficient scales `1`, `1e-11`, `1e-15`, `1e-17`, and `1e-20`, and excluded, partially constrained, and fully constrained boundary faces.
- The matrix tests verify finite scaling weights, the reconstructed operator and right-hand side, the recovered solution, and the residual against the original system. Error checks are normalized by coefficient scale so lost small entries cannot pass through an absolute tolerance.
- Before their respective fixes, **48 small-coefficient cases** reproduced diagonal corruption, including complete erasure, and **15 five-component cases** reproduced the general block defects. All 75 cases pass with the fixes.
- **15 existing physics regressions passed** on 1, 3, and 4 MPI ranks: heat diffusion with CG, BICG, and GMRES; CEP `spiral_BO_2d`; and the Stokes `P1P1/N004` manufactured solution.
- **Three additional Stokes comparisons passed**, using temporary configurations that explicitly select RCS on 1, 3, and 4 MPI ranks.


zasexton:
adding @kko27 and @michelebucelli to review this PR since RCS preconditioning is occasionally used with CEP physics. Mostly want to know if we should add the unit test for regression assessments or instead remove that file and just have the bug fix only in this PR.

tests/unitTests/linear_solver_tests/test_precond_rcs.cpp
kko27, line 1: Seeing that this is the first unit test we have for the linear solver, maybe this could be a separate issue with a dedicated plan in creating a usable template.

kko27 (approved):
Thanks for fixing this!

tests/unitTests/linear_solver_tests/test_precond_rcs.cpp
michelebucelli: I do not think this needs to be in a separate PR. I do think that it would be nice to have a bit of documentation of what the test is doing so that it can be more easily adopted as a template/example to build other tests on.

I would recommend to expand the test's Doxygen documentation for this (Doxygen has a `@test` command for this purpose, I believe).

michelebucelli (commented):
@zasexton I have no comments besides the one about detailing the test rationale a bit more.

tests/unitTests/linear_solver_tests/test_precond_rcs.cpp
zasexton: added doxygen documentation for the test at https://github.com/SimVascular/svMultiPhysics/pull/647/commits/2d59b06e58a0625dfc8450727854f9139b28b298

michelebucelli (approved):

Issue #639, opened 2026-09-04
Fiber stretch rate always evaluated to zero for active stress coupling


michelebucelli:
### Problem

For electromechanics simulations, the fiber stretch rate is currently computed in `Integrator::predictor` by evaluating the fiber stretch with `solutions.current.get_displacement()` and with the displacement `solutions.old.get_displacement()`, and then taking the difference divided by the time step.

However, at the moment when this is evaluated, the current and old solution store the same thing. The `predictor` call comes at the beginning of the `while (true)` loop in the `iterate_solution` function from `main.cpp`, and at at the end of the same loop there's a `Do = Dn` statement to prepare for the next time step.

As a consequence of this, the two fiber stretches will be equal, and the stretch rate will be evaluated to zero.

### Possible solutions

Due to other problems taking priority, I haven't thought too deeply about how this can be solved cleanly. Some possible fixes:
1. change the order in which things are evaluated, to make sure that fiber stretch is computed before `Do = Dn`;
2. store a separate old displacement array;
3. store a fiber stretch array somewhere, and reuse that in the computation of stretch rate (also suggested in #604).

I don't particularly like any of these on their own, and I think it would be better to take this chance to rethink and clarify a bit the way explicitly coupled models (such as the ionic model or the active contraction model) interact with the PIC scheme.

I'll get back to this once I have an electromechanics simulation setup that works properly without this feedback.

### Additional context

1. The fiber stretch rate still shows up different from zero in the VTU output, because `write_vtus` is called in `iterate_solution` before `Do = Dn`, that is before the old solution is made equal to the current.

Issue #178, opened 2024-01-31
Object-oriented solid material models


mrp089:
### Use Case

This affects all solid material models that are currently implemented and future ones.

### Problem

We currently have a massive `switch/case` statement for our solid materials, inherited from `svFSI`. Many of these material models overlap heavily. Furthermore, we cannot combine material models to form a new one, making it hard to implement new (more complex) material models (like G&R).

### Solution

Create a class structure for all existing material models.

### Alternatives considered

`FEBio` has a comprehensive [collection](https://github.com/febiosoftware/FEBio/tree/develop/FEBioMech) of material models. This can help us figure out our material models and some of their inheritance structures.

`Abaqus` has interfaces for user material subroutines. There is one for [general solid materials](https://classes.engineering.wustl.edu/2009/spring/mase5513/abaqus/docs/v6.6/books/sub/default.htm?startat=ch01s01asb31.html) and an invariant-based one for [hyperelastic materials.](https://classes.engineering.wustl.edu/2009/spring/mase5513/abaqus/docs/v6.6/books/sub/default.htm?startat=ch01s01asb28.html). This can help to define interfaces between abstract (e.g., hyperelastic) and derived classes (e.g., Neo-Hooke, Mooney-Rivlin).

### Additional context

This is a good opportunity to take "inventory" of what materials are currently implemented. We can use our (currently under-utilized) `Doxygen` [documentation](https://simvascular.github.io/svFSIplus/annotated.html) when implementing new classes to document the governing equations. This also helps visualize the inheritance structure (see here a [0D example](https://simvascular.github.io/svZeroDSolver/class_block.html)).

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct and Contributing Guidelines

lpapamanolis:
I have started looking into this: I think the hierarchical structure implemented in `Abaqus` is very close to how we would like to restructure our code. The rough class hierarchy would look something like this:
1. **Material**: user-defined label, user should be able to combine different pre-defined material properties/models from the levels below
2. **(Material) Properties**: Material properties grouped in broad categories (categories in abaqus: elastic mechanical properties, inelastic mechanical properties, thermal properties, acoustic properties, hydrostatic fluid properties, equations of state, mass diffusion properties, electrical properties). Here we could add "growth and remodelling properties".
[...]

PR #650, opened 2026-09-21
Fix instability of force-length coupling in electromechanics, and uniform formulation of active stress tensor

### LINKED ISSUE #635: Uniform application of active stress across material models
## Current situation

Different passive material models apply the active stress in different ways:
1. some models do not support active stress (not sure if this is checked at runtime);
2. some models apply tension only along fibers, others also along sheets and cross-fibers; this is checked at runtime in `compute_p2kcc` by whitelisting the models that have all three directions, and throwing an exception if tension along sheets or cross-fibers is positive;
3. most models that support active stress add it to the stress tensor `S_bar`, before calling `bar_to_iso`;  the resulting active stress tensor is different from the one that is usually found in literature (see e.g. [Arostica et al. (2025), eq. (2)](https://doi.org/10.1016/j.cma.2024.117485), or [Fedele et al. (2023), eq. (15)](https://doi.org/10.1016/j.cma.2023.115983)). The one exception to this is `stIso_HO_ma`, which doesn't use `bar_to_iso` at all, and so adds the active contribution to the stress tensor `S` directly;
4. in some models, addition of the active stress is mixed with the passive stress computation, instead of being added at the end; this is mostly a stylistic issue, but I think it would be good for readability to clearly demarcate active and passive contributions.

<details>

<summary> Table of active stress formulation by model </summary>

| Model         | Supports active stress | Supports three directions | AS before `bar_to_iso`            |
|---------------|------------------------|---------------------------|-----------------------------------|
| `stIso_lin`   | no                     | -                         | -                                 |
| `stIso_StVK`  | no                     | -                         | -                                 |
| `stIso_mStVK` | no                     | -                         | -                                 |
| `stIso_nHook  | **yes**                | no                        | yes                               |
| `stIso_MR`    | **yes**                | no                        | yes                               |
| `stIso_HGO`   | **yes**                | no                        | yes                               |
| `stIso_Gucci` | **yes**                | **yes**                   | yes                               |
| `stIso_HO`    | **yes**                | **yes**                   | yes                               |
| `stIso_HO_ma` | **yes**                | **yes**                   | **no** (there is no `bar_to_iso`) |
| `stArtificialNeuralNet` | **yes**      | no                        | **no** (there is no `bar_to_iso`) |

</details>

I believe that there is no reason for points 1 and 2, although they are mostly harmless (unless one wants to use one of the unsupported combinations).

On the other hand, I believe point 3 to be incorrect. I think that active stress should be applied to `S`, instead of `S_bar`, as it does not come from an elastic energy, and as such the derivation of the relation between `S` and `S_bar` does not apply to it.

## Proposed solution

I suggest hoisting the addition of the active stress to after the `switch` statement on the constitutive model. The structure would look something like this:
```cpp
switch (stM.isoType) {
  // ...compute the passive stress from the constitutive model,
  // and add it to the tensor S.
}

// Active stress contribution (maybe, but not necessarily, guarded
// by a check that active stress is present, to save unnecessary
// computations, and maybe rewritten for efficiency, to avoid
// allocating temporary Arrays).
S += Tfa * Hff + Tsa * Hss + Tna * Hnn;
```
This would simultaneously address all four points discussed above.

## Additional context

1. Fixing points 1, 2 and 4 should have no impact on the solution; however, changing point 3 is expected to change the result for those tests that currently use active tension, at least for some of the constitutive models. I believe that the proposed change is more correct, and so the new solutions would also be more correct.
2. The proposed change is related but orthogonal to what is discussed [here](https://github.com/SimVascular/svMultiPhysics/issues/594) about the different formulations/normalizations for active tension. While technically independent of that discussion, making the application of active stress uniform across models gives us a better place to place the active stress formulation switch.
3. Another related but orthogonal issue is #178. While the proposed change would not provide object-oriented refactoring of material models, better decoupling of active and passive material contributions would probably facilitate that in the future. 

------

What are your thoughts @dseyler @javijv4 @aabrown100-git @kko27? Do you know if any of the issues I raise were deliberately implemented as they currently are, and why?

### LINKED ISSUE #641: Temporal instability in electromechanics force-fiber-stretch feedback
## Problem description

Coupled electromechanics simulations, in some settings, give rise to unphysical oscillations in the structural solution, that build up over time eventually leading up to the crash of the solver.

I attach some screenshots of this behavior below, for the slab test case and a ventricular simulation, both using `ActiveStressRegazzoni`. Both were taken at the last computed timestep, after which the linear solver for `struct` failed to converge.

ImageImage

The slab test case was slightly modified with respect to the test suite, to use active stress parameters from [Fedele et al. (2023)](https://doi.org/10.1016/j.cma.2023.115983), and I have introduced a small perturbation to the fiber orientation in one point to break the symmetry (otherwise, the test would use an axis-aligned domain, with an axis-aligned structured grid). I plan to test this again on an unstructured tetrahedral grid without artificial perturbations to the fibers.

I think that @kko27 has observed a similar behavior in testing the implementation of the Land-Niederer model (#606).

## Probable explanation

For realistic active stress models such as `ActiveStressRegazzoni`, the active tension $T_{act}$ depends on the fiber stretch $\lambda$, both directly through the expression of the active tension itself, $T_{act} = T_{act}(\mathbf{s}, \lambda)$, and indirectly through the dependence of the ODE system on $\lambda$ (so that the state vector $\mathbf{s}$ ends up depending on $\lambda$ too).

This dependence is treated explicitly. More precisely:
1. in `Integrator::predictor`, the active stress model is advanced by calling `ActiveStress::advance_time_step` passing the fiber stretch from previous time step;
2. `Integrator::initiator` and `Integrator::corrector` proceed to solve the `struct` equation keeping the active tension fixed.

I believe that the oscillations and blowup are due to the instability (or maybe conditional stability) of this explicit coupling.

## Evidence supporting the explanation

1. In the slab test, reducing $\Delta t$ delays or removes the instability: with $\Delta t = 1 \text{ms}$, the solver blows up at $t = 193 \text{ms}$; with $\Delta t = 0.5 \text{ms}$, it blows up at $t = 230 \text{ms}$; with $\Delta t = 0.25 \text{ms}$, blows up at around $t = 340 \text{ms}$.
2. The oscillations in velocity have a period of 2 timesteps, regardless of timestep size, which is a hallmark of instability due to explicit timestepping.
3. Removing the feedback removes the issue: if I hardcode $\lambda = 1$ in the active stress model code, oscillations disappear.
4. Experimental (AI-generated) code implementing implicit coupling (the active stress model is updated within the nonlinear iterations in which `struct` is also solved) also removes the instabilities (at the price of much increased number of iterations and computational cost). I trust this code only up to a point, but I'll verify it, push it and update this description.
## Possible solutions

1. Implicit partitioned coupling (that is, point 3 from above). It seems to increase computational cost substantially, because convergence requires many more (2-3x) iterations. I experimented adding Aitken acceleration but it didn't help much.
2. Semi-implicit discretization. The dependence of $T_{act}$ on $\lambda$ could be made implicit only in direct part (i.e. in the way $\lambda$ appears in the expression of $T_{act}$), but not in the indirect part (i.e. in the dependence of $\mathbf{s}$ on $\lambda$ through the ODE timestepping). The tangent for this term would be manageable, I think. I have not verified whether this solves the problem or not.

## Open questions

1. The explicit coupling is not uncommon in electromechanics models (see e.g. [Fedele et al. (2023)](https://doi.org/10.1016/j.cma.2023.115983)). It is not obvious to me why it should not work here. Some candidate (currently unverified) explanations:
    1. other solvers may use more dissipative timestepping methods, and numerical dissipation compensates for instability;
    2. other models may have further sources of damping (e.g. more viscous boundary conditions)

michelebucelli:
Fixes #635; fixes #641.

## Current situation

1. Different passive material models apply the active stress in different ways, although in principle active and passive models could be fully independent (see discussion on #635);
2. Explicit coupling between active stress and structural mechanics through the fiber stretch $\lambda$ may give rise to temporal instabilities (see discussion on #641).

## Release Notes

The PR addresses both issues above. Active tension, in particular, is now evaluated implicitly, and appropriate terms are added to the `(u)struct` matrices to account for the coupling. A detailed list of changes follows.

1. **Modified `compute_pk2cc` from `mat_models.cpp`** so that all material models use the same formulation for the active stress tensor.
2. **Added the purely virtual function `ActiveStress::compute_active_tension_derivative_local`** to compute the partial derivative of the active tension $T_\text{act}$ with respect to fiber stretch $\lambda$. The function was implemented in concrete models, with all models returning `0.0` except for `ActiveStressRegazzoni`.
3. **Added the helper function `ActiveStress::compute_tension`**, to return active tensions along principal directions and their derivatives with respect to $\lambda$, bundled in an object of the new type `ActiveStress::ActiveTension`, introduced to be easily passed to downstream functions.
4. The previous code computed active tension at integration points by
    1. projecting $\lambda$ from integration points to mesh nodes with a lumped $L^2$ projection;
    2. computing $T_\text{act}$ at mesh nodes;
    3. interpolating $T_\text{act}$ at integration points.
    
   In this framework, computing the tangent contribution of active stress would have required computing the derivative of $T_\text{act}$ wrt displacement, which would have meant differentiating through the $L^2$ projection. This would be a nonlocal, and thus nontrivial, operation.
   
   The new code **evaluates $\lambda$ and $T_\text{act}$ directly at integration points**. The nodal evaluation of $T_\text{act}$ is kept for the sake of output.
5. **Added helper class `ActiveStress::evaluator`** to compute active tension and its derivative at integration points.
6. **Modified `compute_pk2cc` from `mat_models.cpp`** to also account for the derivative of active tension when computing the tensor `Dm`. This also required modifying `cc_to_voigt_eigen` to relax its symmetry assumption.
7. By default, the active stress model state is still advanced explicitly (i.e. implicit coupling is only used to evaluate the active tension function). Optionally, it can be made implicit by setting the XML parameter `Active_stress/Implicit_state_coupling` to true. This coupling does not contribute to the tangent, and is thus a fixed-point coupling between the two models. I implemented this experimentally, but opted to keep it in view of exploring its effect on force-velocity coupling stability (see #639).
8. **Refactoring of `Integrator`** to accomodate for the implicit coupling.
9. Updated all downstream consumers of active stress to accomodate for changed interfaces.

## Documentation

All modified/new classes or functions were given Doxygen documentation.

## Testing

A few tests (`electromechanics/slab/Regazzoni`, `struct/tensile_adventitia_Guccione_active`, `ustruct/LV_Guccione_active`) failed due to the change to the active stress formulation. I have regenerated their reference solution, as per the discussion in #635.

Besides that, all automatic tests are passing.

## Additional context

1. In principle, the uniform formulation of active stress and the instabilities are independent changes; however, fixing the instabilities required evaluating the tangent of active stress, for which having a uniform formulation was helpful. This is why the two changes are merged into a single PR.
5. A unified formulation for active stress will also be beneficial when we will introduce the option to switch between different active stress tensor definitions, as discussed [here](https://github.com/SimVascular/svMultiPhysics/issues/594#issuecomment-5121460863).

### DIFF
--- a/Code/Source/solver/ActiveStress.cpp
+++ b/Code/Source/solver/ActiveStress.cpp
@@ -16,6 +16,7 @@ void ActiveStress::read_parameters(const ActiveStressParameters &params) {
 
   implicit_coupling_ = params.get_implicit_coupling();
   relaxation_coefficient = params.get_relaxation_coefficient();
+  aitken_relaxation_enabled_ = params.get_aitken_relaxation_enabled();
 
   svmp::check<svmp::ParseException>(
       relaxation_coefficient > 0.0 && relaxation_coefficient <= 1.0,
@@ -31,6 +32,14 @@ void ActiveStress::read_parameters(const ActiveStressParameters &params) {
       "is disabled, but got " +
           std::to_string(relaxation_coefficient) + ".");
 
+  // Aitken's method estimates the relaxation coefficient from the residuals of
+  // two consecutive fixed-point iterations, which only exist when the coupling
+  // is implicit.
+  svmp::check<svmp::ParseException>(
+      implicit_coupling_ || !aitken_relaxation_enabled_,
+      "Active stress Aitken relaxation requires Implicit_coupling to be "
+      "enabled.");
+
   read_model_specific_parameters(
       params.get_parameters(params.get_model_name()));
 }
@@ -43,6 +52,7 @@ void ActiveStress::distribute_parameters(const CmMod &cm_mod,
 
   cm.bcast(cm_mod, &implicit_coupling_);
   cm.bcast(cm_mod, &relaxation_coefficient);
+  cm.bcast(cm_mod, &aitken_relaxation_enabled_);
 
   distribute_model_specific_parameters(cm_mod, cm);
 }
@@ -63,9 +73,21 @@ void ActiveStress::init(const unsigned int tnNo) {
   states_at_time_step_start = states;
 
   active_tension.resize(tnNo);
+
+  if (aitken_relaxation_enabled_) {
+    aitken_relaxation.resize(tnNo);
+    previous_residual.resize(tnNo);
+  }
 }
 
-void ActiveStress::time_advance() { states_at_time_step_start = states; }
+void ActiveStress::time_advance() {
+  states_at_time_step_start = states;
+
+  if (aitken_relaxation_enabled_)
+    aitken_relaxation = relaxation_coefficient;
+
+  previous_residual_available = false;
+}
 
 void ActiveStress::update(const double t, const double dt,
                           const Vector<double> &calcium,
@@ -73,16 +95,34 @@ void ActiveStress::update(const double t, const double dt,
                           const Vector<double> &fiber_stretch_rate) {
   time = t;
 
-  const double omega = relaxation_coefficient;
-
   for (int i = 0; i < active_tension.size(); ++i) {
     Vector<double> state_loc = states_at_time_step_start.col(i);
     advance_time_step_local(t, dt, calcium[i], fiber_stretch[i],
                             fiber_stretch_rate[i], state_loc);
     states.set_col(i, state_loc);
 
-    const double tension =
-        compute_active_tension_local(state_loc, fiber_stretch[i]);
-    active_tension[i] = omega * tension + (1.0 - omega) * active_tension[i];
+    // Residual of the fixed-point iteration on the active tension.
+    const double residual =
+        compute_active_tension_local(state_loc, fiber_stretch[i]) -
+        active_tension[i];
+
+    double omega = relaxation_coefficient;
+
+    if (aitken_relaxation_enabled_) {
+      // Node-wise Aitken estimate. The formula divides by the difference
+      // between the two residuals, so the coefficient of the previous
+      // iteration is kept where they coincide exactly, which is the case at
+      // every node whose active tension has stopped changing.
+      if (previous_residual_available && residual != previous_residual[i])
+        aitken_relaxation[i] *=
+            -previous_residual[i] / (residual - previous_residual[i]);
+
+      omega = aitken_relaxation[i];
+      previous_residual[i] = residual;
+    }
+
+    active_tension[i] += omega * residual;
   }
+
+  previous_residual_available = true;
 }
--- a/Code/Source/solver/ActiveStress.cpp
+++ b/Code/Source/solver/ActiveStress.cpp
@@ -14,7 +14,7 @@ void ActiveStress::read_parameters(const ActiveStressParameters &params) {
   eta_s = params.get_eta_s();
   eta_n = params.get_eta_n();
 
-  implicit_coupling_ = params.get_implicit_coupling();
+  implicit_state_coupling_ = params.get_implicit_state_coupling();
 
   read_model_specific_parameters(
       params.get_parameters(params.get_model_name()));
@@ -26,7 +26,7 @@ void ActiveStress::distribute_parameters(const CmMod &cm_mod,
   cm.bcast(cm_mod, &eta_s);
   cm.bcast(cm_mod, &eta_n);
 
-  cm.bcast(cm_mod, &implicit_coupling_);
+  cm.bcast(cm_mod, &implicit_state_coupling_);
 
   distribute_model_specific_parameters(cm_mod, cm);
 }
--- a/Code/Source/solver/ActiveStress.cpp
+++ b/Code/Source/solver/ActiveStress.cpp
@@ -17,6 +17,8 @@ void ActiveStress::read_parameters(const ActiveStressParameters &params) {
   implicit_coupling_ = params.get_implicit_coupling();
   relaxation_coefficient = params.get_relaxation_coefficient();
   aitken_relaxation_enabled_ = params.get_aitken_relaxation_enabled();
+  global_aitken_relaxation_enabled_ =
+      params.get_global_aitken_relaxation_enabled();
 
   svmp::check<svmp::ParseException>(
       relaxation_coefficient > 0.0 && relaxation_coefficient <= 1.0,
@@ -40,6 +42,11 @@ void ActiveStress::read_parameters(const ActiveStressParameters &params) {
       "Active stress Aitken relaxation requires Implicit_coupling to be "
       "enabled.");
 
+  svmp::check<svmp::ParseException>(
+      aitken_relaxation_enabled_ || !global_aitken_relaxation_enabled_,
+      "Active stress Global_Aitken_relaxation requires Aitken_relaxation to be "
+      "enabled.");
+
   read_model_specific_parameters(
       params.get_parameters(params.get_model_name()));
 }
@@ -53,11 +60,15 @@ void ActiveStress::distribute_parameters(const CmMod &cm_mod,
   cm.bcast(cm_mod, &implicit_coupling_);
   cm.bcast(cm_mod, &relaxation_coefficient);
   cm.bcast(cm_mod, &aitken_relaxation_enabled_);
+  cm.bcast(cm_mod, &global_aitken_relaxation_enabled_);
 
   distribute_model_specific_parameters(cm_mod, cm);
 }
 
-void ActiveStress::init(const unsigned int tnNo) {
+void ActiveStress::init(const unsigned int tnNo,
+                        const Vector<double> &owned_nodes_) {
+  owned_nodes = owned_nodes_;
+
   states.resize(n_states, tnNo);
 
   if (n_states > 0) {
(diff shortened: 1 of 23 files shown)


Code/Source/solver/Integrator.cpp
```diff
@@ -464,79 +650,7 @@ void Integrator::predictor()
...
-        for (int a = 0; a < mesh.nNo; ++a)
-          fiber_stretch_rate[mesh.gN[a]] = tmp[a];
-      }
-    } else {
-      // No domain solves for displacement, or no model reads stretch rate:
-      // Set the fiber stretch rate to 0, corresponding to no movement.
-      fiber_stretch_rate = 0.0;
-    }
-  }
+  compute_fiber_stretch(fiber_stretch, fiber_stretch_rate);
```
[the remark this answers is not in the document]
michelebucelli: This is already discussed in issue #639, and will be addressed in a separate PR.

Code/Source/solver/ActiveStress.cpp
```diff
@@ -3,17 +3,71 @@
...
+  const unsigned int n_states = active_stress.n_states;
+
+  if (state_.nrows() != n_states || state_.ncols() != nodes.size())
+    state_.resize(n_states, nodes.size());
+
+  // Friend access to active_stress.states, so that gathering the state of an
+  // element does not need to go through an accessor.
+  for (int a = 0; a < nodes.size(); ++a)
+    for (unsigned int j = 0; j < n_states; ++j)
+      state_(j, a) = active_stress.states(j, nodes(a));
```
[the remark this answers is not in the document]
michelebucelli: This is true, but it is also probably true that `ActiveStress` is not fully compatible with restarting, even before this PR.

This requires better testing, and I think it's best to leave it for a future issue/PR. Similar concerns apply to `IonicModel`, especially after the fixes that @samibismar is doing in #549.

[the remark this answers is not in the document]
kko27: @michelebucelli, I agree with deferring this to a separate PR. I believe that simulations with genBC also cannot restart, and this might be a broader issue where any ODE coupling in the solver hasn't thought about re-initialization/restarts.

Code/Source/solver/ActiveStressRegazzoni.h
```diff
@@ -45,8 +45,12 @@
...
  * customized time-stepping scheme to handle the stiffness of the model.
  *
- * @todo Force-strain-rate feedback requires a stabilization strategy for robust
- * use in coupled electromechanics. This will be addressed in a follow-up PR.
+ * @note Both the direct dependence of @f$\Tact@f$ on the fiber stretch and the
+ * force-strain-rate feedback make the active tension a function of the
+ * mechanics solution. Treating those dependences explicitly can be unstable in
+ * time. The direct one is always resolved within the nonlinear iterations of
+ * the mechanics problem; enabling @c Implicit_state_coupling resolves the
+ * indirect one there too (see @ref ActiveStress).
```
michelebucelli: This comment is slightly incorrect and needs fixing. The previous comment on force-strain-rate feedback being currently unstable/unsupported should be restored.

Code/Source/solver/Integrator.cpp
```diff
@@ -371,6 +384,180 @@ void Integrator::update_residual_arrays(eqType& eq) {
...
+
+      n_domains++;
+    }
+
+    if (n_domains > 0) {
+      cep_mod.cem.Ya_f[Ac] = Ta_f / n_domains;
+      cep_mod.cem.Ya_s[Ac] = Ta_s / n_domains;
+      cep_mod.cem.Ya_n[Ac] = Ta_n / n_domains;
+    }
+  }
```
michelebucelli: Since this is only used for output, it should not be here, but rather in the output code. This is not only cleaner, but also avoids recomputing this at every time step (or multiple times per step, in case of implicit coupling).

I'm currently trying to figure out how to best realize this.

Code/Source/solver/Integrator.cpp
```diff
@@ -371,6 +384,180 @@ void Integrator::update_residual_arrays(eqType& eq) {
...
+
+  return false;
+}
+
+//------------------------
+// update_active_stress
+//------------------------
+void Integrator::update_active_stress(eqType& eq, const Vector<double>& fiber_stretch,
+    const Vector<double>& fiber_stretch_rate, const bool within_nonlinear_iterations) {
```
michelebucelli: The flag `within_nonlinear_iteration` can be simplified away by splitting the time advancing and the update steps into separate functions. This makes the code a bit cleaner, and makes more logical sense.

Code/Source/solver/mat_models.cpp
```diff
@@ -101,56 +101,35 @@ void cc_to_voigt(const int nsd, const Tensor4<double>& CC, Array<double>& Dm)
...
-      }
+  // Index pairs of the tensor corresponding to each index in Voigt notation.
+  constexpr int n_voigt = 3 * (nsd - 1);
+  constexpr int voigt_row[6] = {0, 1, 2, 0, 1, 2};
+  constexpr int voigt_col[6] = {0, 1, 2, 1, 2, 0};
+
+  // In 2D the only shear index is (0,1), which sits in the fourth entry of the
+  // maps above rather than in the third.
+  constexpr int voigt_2d_row[3] = {0, 1, 0};
+  constexpr int voigt_2d_col[3] = {0, 1, 1};
```
michelebucelli: These can probably be unified through either some metaprogramming or template specialization.

Code/Source/solver/mat_models.cpp
```diff
@@ -786,7 +716,11 @@ void compute_pk2cc(const ComMod &com_mod, const CepMod &cep_mod,
       std::array<Tensor<nsd>,9> ddInv;
       Matrix<nsd> N1;
 
-      // Compute and store invariants and derivatives wrt C in array of matrices/tensors
+      // Compute and store invariants and derivatives wrt C in array of
+      // matrices/tensors
+      // @todo[michelebucelli] Tfa is unused in this call, and it should
+      //   probably be removed from the function signature. Active stress is
+      //   added below in any case.
       CANNModel.computeInvariantsAndDerivatives<nsd>(C, fl, nfd, J2d, J4d, Ci, Idm, Tfa, N1, psi, Inv, dInv, ddInv);
```
michelebucelli, line 731: Is there anyone more familiar with this material model that can confirm removing `Tfa` from the arguments here is safe? The code would be equivalent, I'm not sure if there's some design reason or planned development that would justify keeping it.

Code/Source/solver/mat_models.cpp
```diff
@@ -810,7 +741,75 @@ void compute_pk2cc(const ComMod &com_mod, const CepMod &cep_mod,
...
+  svmp::check<svmp::InternalErrorException>(
+      nfd >= 1,
+      "At least one fiber direction must be defined for active stress.");
+
+  // Derivative of the active stress with respect to the fiber stretch, at fixed
+  // state of the active stress model. Accumulated along with the active stress
+  // itself, and used below to build its tangent.
+  const bool has_tangent = !utils::is_zero(active_tension.d_fibers) ||
+                           !utils::is_zero(active_tension.d_sheets) ||
+                           !utils::is_zero(active_tension.d_sheet_normals);
```
michelebucelli: This declaration should be moved closer to where its value is actually used.

Code/Source/solver/mat_models.cpp
alisonmarsden: @msbazzi and @divya-adil should weigh in here.   they are actively working on the CANN model implementation for more than one paper - so yes it is important and we need to keep it.   They can explain the structure.

Code/Source/solver/Integrator.cpp
michelebucelli: Done: now neither `ActiveStress` nor `Integrator` update any nodal active tension vector, but those are only computed on demand when writing output.

The vectors `cep_mod.cem.Ya_{f,s,n}` have also been removed, since they were no longer used. They were being read from/written into by restart, but restart for `ActiveStress` is not working anyways, because the state vectors are not restarted, so this doesn't change the existing behavior.

Restart will be addressed in a follow-up issue/PR.

Code/Source/solver/mat_models.cpp
msbazzi: @michelebucelli, you are correct that `Tfa` is not currently used in `computeInvariantsAndDerivatives`, so in the present CANN implementation, removing it from this particular call would not change the results.

The reason I would like to preserve the active-tension capability is because of where we are taking the CANN model next. We have been developing the CANN framework as a more flexible constitutive representation for vascular tissue, particularly for vein mechanics, rather than treating it only as another passive material model. The next step is to use the CANN-based constitutive description within a growth and remodeling framework.

My current plan is to build on the Fortran-based constrained-mixture/FSG implementation we discussed in #2, while making the changes needed to make it clearer and more general within `svMultiPhysics`. In that setting, the passive constitutive response and active smooth-muscle tension will ultimately need to coexist as part of the evolving wall mechanics.

That said, I’m also open to suggestions on the cleanest way to handle this. If `Tfa` does not need to be passed into `computeInvariantsAndDerivatives` yet, we can refactor it while keeping the active-tension pathway for the planned G&R implementation.

michelebucelli: @msbazzi thanks for explaining!

I think that, for the purpose of this PR, the question to address is whether the CANN-based stress tensor evaluation pipeline returns (now or in the planned future) the passive stress tensor only (option A) or the sum of passive and active stress tensor (option B).

Under option A, keeping `Tfa` as an input argument would mean that the passive stress can depend on the active tension, which is maybe counterintuitive but plausible in principle, and the code of this PR can be left as it is. `Tfa` can maybe be removed in the interest of code hygiene, but it is somewhat of a minor issue.

On the other hand, under option B the code of this PR is incorrect, because the lines after the switch add active stress regardless of the passive constitutive model, and the CANN-based model would end up having the active stress added to it twice (perhaps with two different formulations even).

I am not familiar with the CANN-based model, so there may be something I'm missing, but I think option A would be the cleanest, both conceptually and in the code, because it neatly follows the additive decomposition of stress which is behind the active stress formalism, and keeps passive and active constitutive models separate, which is a good step towards #178. However, if option B is necessary, then I'd have to rework this PR a little bit to accomodate for that.

I also note that the CANN model currently evaluates the stress as the derivative of an energy function (through `CANNModel.evaluate`), but active stress in general cannot be expressed in that way (I think), which makes me lean even more towards option A.

msbazzi: @michelebucelli, currently (and I would say likely for the near future), CANN-based models are used only to describe passive stress, while active stress will come from a different formulation, given that I agree with you that A is the best and cleanest option.

michelebucelli: I see, thanks! I think the current code for this PR is mostly fine then.

I propose we remove the argument `Tfa` from this call, since it is currently unused and it might be a bit counterintuitive (as in, it might suggest option B to the reader, like it did for me), and we can bring it back if necessary in the future.

Code/Source/solver/CepMod.h
kko27: @michelebucelli do you know why we have this `cemModelType` class? It looks like legacy code for stretch activated currents...

Code/Source/solver/post.cpp
```diff
@@ -802,51 +810,111 @@ void fib_stretch(const ComMod& com_mod, const int iEq, const mshType& lM,
   }
 }
 
+/// @brief Compute active tension along fibers, sheets and sheet normals at
+/// every mesh node.
+//
+void active_tension(const ComMod &com_mod, const int iEq, const mshType &lM,
```
kko27: @michelebucelli. Will the eventual goal be to make `active_tension`, `fib_stretch`, and `fib_stretch_rate` purely post-processing functions (just for writing to output files)? I see `active_tension` is already just for post-processing, and `fib_stretch_rate` is a bug if it is used in the integrator (as discussed previously).

Code/Source/solver/Integrator.cpp
```diff
@@ -110,6 +110,18 @@ bool Integrator::step(bool save_results) {
     // Compute body forces
     set_body_forces();
 
+    // Implicit state coupling of the active stress: re-advance the state of the
```
kko27: @michelebucelli Just wanted to check if this is intended behavior. So if one of my domains has `implicit_state_coupling` set to true, then all of my domains get treated implicitly

Code/Source/solver/Integrator.cpp
```diff
@@ -371,6 +383,134 @@ void Integrator::update_residual_arrays(eqType& eq) {
...
+
+    else if (eq.phys == Equation_CEP) {
+      need_fiber_stretch = true;
+    }
+  }
+
+  // If we need to compute fiber stretch, we iterate through all meshes, compute
+  // the stretch for each mesh, and then copy the mesh-local resulting vector
+  // into the global vector.
+  if (have_active_stress || need_fiber_stretch) {
```
kko27: We may at some point want to figure out how much computation time is spent from calling `fib_stretch` and `fib_stretch_rate`, especially for larger meshes. Maybe for implicit coupling, we don't always have to call this at every iteration

tests/cases/electromechanics/slab/result_Regazzoni_001.vtu
kko27: @michelebucelli Does this mean that by default, all active stress models will treat the direct dependence on active tension implicitly? (i.e., there is no need to specify a xml parameter for implicit coupling, unless I want the states to implicitly depend on the fiber stretch).

kko27 (approved):
@michelebucelli Thanks for making these changes! This definitely simplifies many of the function signatures. I mostly have clarification questions and comments related to future PRs.

